PR #13, opened 2024-02-29
admin: Add --defaults options

uristdwarf:
Previous iteration of the configuration dumping command did not allow dumping default values of services, since default values are set by the caller to the cfg_get* functions, thus there being no single truth to these option values or even names.

This commit adds maps of the currently known configuration options, and the dump-config command uses these to figure out what are the defaults for each service.

This is done with best-effort, because this list may become outdated as time goes on and new options are added. Until we have a single source of truth for these options, this will have to do.

### DIFF
--- a/src/admin/dump_config_command.cc
+++ b/src/admin/dump_config_command.cc
@@ -20,11 +20,14 @@
 #include "admin/dump_config_command.h"
 
 #include <unistd.h>
+#include <yaml-cpp/yaml.h>
 #include <iostream>
 
 #include "admin/registered_admin_connection.h"
 #include "protocol/cltoma.h"
 #include "protocol/matocl.h"
+
+std::string add_defaults(std::string &config);
 
 std::string DumpConfigurationCommand::name() const { return "dump-config"; }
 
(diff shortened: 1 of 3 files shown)


src/admin/dump_config_command.cc
```diff
@@ -48,5 +61,170 @@ void DumpConfigurationCommand::run(const Options &options) const {
...
 	std::string configs;
 	matocl::adminDumpConfiguration::deserialize(adminResponse, configs);
+	if (options.isSet(defaultsMode)) {
+		configs = add_defaults(configs);
+	}
 	std::cout << configs << "\n";
 }
+
+// Default configuration options below...
+// This is informational only, and not used anywhere else in the codebase.
```
aNeutrino, line 71: Good Enough first version.

aNeutrino (approved):
LGTM

lgsilva3087 (approved):
I understand this is a quick fix.
Please take into consideration that it introduces another place to update every time a default value changes.
Please, create the task to implement my proposal:
```
- class Option;
- class Configuration; //(singleton, map of Options)
- only access function cfg_get_* inside the previous classes.
a function for each module to initialize the Options with default values.
```

PR #14, opened 2024-02-29
admin: add --help option

uristdwarf:


src/admin/main.cc
lgsilva3087, line 1: Even if the --help is now accepted. The help output is strange:
image
--help should be a valid command and the exception message should not be printed.

lgsilva3087 (changes requested):
Please see my comment.

src/admin/main.cc
uristdwarf: I was working on trying to fix up the option handling and got this error too. I thought just keeping a simple one line change would fix it. Guess not.

I'll try fixing it, but I might just decide to refactor the command-line option handling.

uristdwarf: From my git stash, I tried fixing it but I decided to keep the change simple now.
image

lgsilva3087: I understand, please create the tasks to fix it later.

lgsilva3087 (approved):
Ready.

PR #20, opened 2024-03-07
all: Add option to generate compile_commands.json

lgsilva3087:
Clangd and vscode rely on this file to understand the build context of each file in the project. This commit adds the option to easily generate it. The option is enabled by default.

To properly work, the file should be in the root of the repository. At this point, a manual soft link should be generated, pointing to the actual file (cmake_build_directory/compile_commands.json).


Antuan-AE (approved):
LGTM

ralcolea (approved):
Nice job!!!

PR #23, opened 2024-03-11
all: Increase the nofile security limits

lgsilva3087:
The previous limit (10000) in the number of open files (nofile) for user saunafs was too small for two main reasons:
- In the past we had to change this limit on customer side. Some workloads are requesting a big number of open file descriptors at least in the chunkserver side.
- The new SPLIT format for the chunks can double the usage.

This commit increases the limit and makes it base 2.


ralcolea (approved):
Nice job!!! :+1:

Antuan-AE (approved):
LGTM

PR #25, opened 2024-03-12
all: Refactor initializations based on runtabs

lgsilva3087:
The initialization behavior, shared between master, metalogger and chunkserver, was written in plain old C. This commit updates the related parts in the code to be more readable in C++. The refactor scope was:

- Remove duplicated code from the init.h files (there are 3 of them).
- Remove dead code (unused macros and includes).
- Replace C-like callbacks by std::function.
- Replace char * by std::string.
- Replace the C style arrays by std::vector.
- Refactor the 'initialize' function to be more readable.
- Document the main ideas.

### DIFF
--- a/src/common/run_tab.h
+++ b/src/common/run_tab.h
@@ -1,9 +1,5 @@
 /*
-   Copyright 2005-2010 Jakub Kruszona-Zawadzki, Gemius SA
-   Copyright 2013-2014 EditShare
-   Copyright 2013-2015 Skytechnology sp. z o.o.
    Copyright 2023      Leil Storage OÜ
-
 
    SaunaFS is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


src/common/run_tab.h
```diff
@@ -0,0 +1,29 @@
...
+#include <string>
+
+/// Simple function prototype for callbacks
+using RunTabFuntion = std::function<int()>;
+
+/// Structure to pack the function and a descriptive name together
+struct RunTab {
+	RunTabFuntion function;  ///< Actual function to call
+	std::string name;        ///< Descriptive name (useful to log errors)
+};
```
Antuan-AE, line 29: Please add an empty line at the end of the file.

src/master/init.h
```diff
@@ -36,35 +36,30 @@
...
+    // has to be second
+    RunTab{metadataserver::personality_init, "personality"},
+    RunTab{rnd_init, "random generator"},
+    // has to be before 'fs_init' and 'matoclserv_networkinit'
+    RunTab{dcm_init, "data cache manager"},
+    // has to be before 'fs_init'
+    RunTab{matoclserv_sessionsinit, "load stored sessions"},
+    RunTab{exports_init, "exports manager"},
+    RunTab{topology_init, "net topology module"},
+	// the lambda is used to select the correct fs_init overload
```
Antuan-AE, line 56: This comment has an incorrect indentation. Please fix it.

Antuan-AE (approved):
Please review my suggestions. LGTM.

src/common/run_tab.h
lgsilva3087: Done.

src/master/init.h
lgsilva3087: Done.

src/main/main.cc
```diff
@@ -136,37 +136,41 @@ static void signal_pipe_serv(const std::vector<pollfd> &pdesc) {
...
+
 	eventloop_updatetime();
-	return ok;
+
+	return isOk;
 }
 
-int initialize_early(void) {
-	return initialize(EarlyRunTab);
+bool initialize_early() {
```
rolysr, line 164: Although it's not the priority in this PR, maybe, it might be beneficial to use camel case for this type of functions and variables to update them to the desired format. Perhaps, we could dedicate a separate PR to this.

rolysr (approved):
I think it is ready, anyway please see my single comment related to this PR.

dmga44 (approved):

src/main/main.cc
dmga44: I think @lgsilva3087 chose to keep the same style of the file, though it is not the one we are applying everywhere. Looks good either way to me.

ralcolea (approved):
Very nice job @lgsilva3087.
These changes improve indeed the technical debt in our system :+1:

PR #26, opened 2024-03-13
admin: Deprecate saunafs-probe command

lgsilva3087:
The command was introduced in version 2.5.0 (2014) and later renamed from 'probe' to 'admin' (2015). A soft link was kept for backward compatibility. Even if both work, there were some inconsistencies, like:

- The package name was admin, but all the main classes still had probe in the name.
- Some of the tests had admin in the filename while other had probe.
- They look like different commands but the other is just a soft link.

This commit deprecates the saunafs-admin command and removes all the references from the tool, tests, packages and documentation.


uristdwarf:
I've noticed two major problems we currently have with semantic versioning.

1. We haven't clearly defined what our public API is. What exactly do we consider as breaking changes? This needs to be updated for the documentation.
2. Is saunafs-probe part of the public API? My gut feeling says yes because it's part of the documentation.

In short, even though no one should be using this tool anymore, we can't remove it entirely just yet, but we need to make absolutely clear that this is deprecated and you should use saunafs-admin instead (thus if anyone uses saunafs-probet it's their fault).

removal != deprecation in this case.

debian/saunafs-adm.install
```diff
@@ -1,2 +1 @@
 usr/bin/saunafs-admin
-usr/bin/saunafs-probe
```
uristdwarf, line 2: Potentially breaking change

rpm/saunafs.spec
```diff
@@ -402,8 +402,6 @@ rm -rf $RPM_BUILD_ROOT
 %doc NEWS README.md UPGRADE
 %attr(755,root,root) %{_bindir}/saunafs-admin
 %{_mandir}/man8/saunafs-admin.8*
-%{_bindir}/saunafs-probe
-%{_mandir}/man8/saunafs-probe.8*
```
uristdwarf, line 406: Potentially breaking change

src/admin/CMakeLists.txt
```diff
@@ -8,9 +8,3 @@ link_unittest(saunafs-admin-lib sfscommon)
 add_executable(saunafs-admin ${SAUNAFS_ADMIN_MAIN})
 target_link_libraries(saunafs-admin saunafs-admin-lib)
 install(TARGETS saunafs-admin RUNTIME DESTINATION ${BIN_SUBDIR})
-
-add_custom_target(saunafs-probe ALL
-    COMMAND ln -sf saunafs-admin saunafs-probe
-    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-    DEPENDS saunafs-admin)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/saunafs-probe DESTINATION ${BIN_SUBDIR})
```
uristdwarf, line 16: Potentially breaking change

doc/CMakeLists.txt
```diff
@@ -8,7 +8,6 @@ set(MANPAGES
     saunafs.7
     saunafs-admin.8
     saunafs-cgiserver.8
-    saunafs-probe.8          # not a source
```
uristdwarf, line 11: Instead of removing the man file entirely, we should perhaps make it's own man file and clearly explain that it is deprecated and not supported.

uristdwarf (commented):
Everything else is fine, but we need to really think twice removing something like this in a minor version.

Antuan-AE (approved):
Please review my comments before merging. LGTM.

debian/saunafs-adm.install
lgsilva3087: Reverted.

rpm/saunafs.spec
lgsilva3087: Reverted.

src/admin/CMakeLists.txt
lgsilva3087: Reverted.

doc/CMakeLists.txt
lgsilva3087: Reverted.

tests/test_suites/ShortSystemTests/test_admin_info.sh
```diff
@@ -15,4 +15,4 @@ done
 rm dir_3/file
 rm dir_xor2/file
 expect_equals "$SAUNAFS_VERSION 2 0 0 9 5 4 4 7 7" \
```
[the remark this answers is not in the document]
lgsilva3087: Fixed.

uristdwarf (approved):
I've added a section in the saunafs-admin man to use that instead of saunafs-probe

PR #27, opened 2024-03-15
tests: Update check on probe info command

Antuan-AE:
After Metadata parallel loading (# 15), the metadata server is returning an extra field for symlinks amount inodes to the info command of saunafs-probe tool. This commit fix a functional test checking this tool.


lgsilva3087 (approved):
Ready.

uristdwarf (approved):

PR #28, opened 2024-03-18
chunkserver: Fix global variables visibility

lgsilva3087:
The static variable gPerformSync was not shared correctly between the chunkserver and the chunkserver-common at runtime. Causing different values from each side.

This commit makes it inline, which also fixes the
ShortSystemTest.test_fsync_eio_in_chunkserver.

The variable gScansInProgress was also changed to be inline to avoid the same issue.


rolysr (approved):

uristdwarf (approved):
Works on my machine :P

PR #31, opened 2024-03-20
cgi: Update sfs.css

ss-leil:
CSS improvements for the monitoring.

### DIFF
--- a/src/cgi/sfs.css
+++ b/src/cgi/sfs.css
@@ -25,9 +25,9 @@ body {
 	bottom:0;
 	right:0;
 	overflow:auto;
-	background:#010101;
+	background:#EFF2F6;
 	padding:15px 2px 15px 2px;
-	color:#fff;
+	color:#000;
 	text-align:center;
 }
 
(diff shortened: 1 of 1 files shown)


dmga44:
Maybe not relevant @ss-leil , but could you use our commit message format? Your commit message should probably look like this: "cgi: Update file sfs.css web interface".

I think this change has nothing to do with CI and we could merge it even before CI works again.

uristdwarf:
It's good, but as @dmga44 mentioned, please format your PR title to something like `cgi: Update sfs.css`. While someone can do it later when merging, it's good practice.

> I think this change has nothing to do with CI (only changes colors) and we could merge it even before CI works again.

Repository settings do not allow merging when CI doesn't pass and while I could force it through, it's best to wait until it works again unless it's really urgent.

uristdwarf (approved):

lgsilva3087 (changes requested):
The PR is pointing to main, it should be dev.

lgsilva3087:
I changed the target to dev.

lgsilva3087 (approved):

PR #34, opened 2024-03-21
fuse, mount: Add sfsignoreflush option

dmga44:
The sfsignoreflush option targets answering inmediately to
the flush or fsync syscall without properly checking its responses. The
core of implementation of both syscall is write_data_flush function,
this is why is the mocked one. This fools single-file copiers into
believing the files are already copied, and the system can handle those
copies in a multithreaded fashion.

The tests performed in Windows show that to achieve the desired behavior
similar to Linux's ignoreflush, the read syscall should be also mocked
in this special mode. This is also the reason behind the renaming of
this option.

On both platforms, it has been proven the value of this option in the copy of small files.

### DIFF
--- a/doc/sfsmount.1.adoc
+++ b/doc/sfsmount.1.adoc
@@ -204,6 +204,11 @@ Equivalent to *--nostdopts* (*-n*) option for use in fstab.
 *-o nonempty*::
 Equivalent to *--nonempty* option for use in fstab.
 
+*-o sfsignoreflush*::
+DANGEROUS. Ignore flush/fsync usual behavior by replying SUCCESS to it 
+inmediately. Targets fast creation of small files, but may cause data loss 
+during crashes when allegedly flushed data is still being processed.
+
 General mount options (see *mount*(8) manual):
 
 *-o rw* | *-o ro*::
--- a/doc/sfsmount.1.adoc
+++ b/doc/sfsmount.1.adoc
@@ -205,8 +205,8 @@ Equivalent to *--nostdopts* (*-n*) option for use in fstab.
 Equivalent to *--nonempty* option for use in fstab.
 
 *-o sfsignoreflush*::
-DANGEROUS. Ignore flush usual behavior by replying SUCCESS to it 
-inmediately. Targets fast creation of small files, but may cause data loss 
+Advanced: use with caution. Ignore flush usual behavior by replying SUCCESS to it 
+immediately. Targets fast creation of small files, but may cause data loss 
 during crashes when allegedly flushed data is still being processed.
 
 General mount options (see *mount*(8) manual):
--- a/doc/sfsmount.1.adoc
+++ b/doc/sfsmount.1.adoc
@@ -205,7 +205,7 @@ Equivalent to *--nostdopts* (*-n*) option for use in fstab.
 Equivalent to *--nonempty* option for use in fstab.
 
 *-o sfsignoreflush*::
-DANGEROUS. Ignore flush/fsync usual behavior by replying SUCCESS to it 
+DANGEROUS. Ignore flush usual behavior by replying SUCCESS to it 
 inmediately. Targets fast creation of small files, but may cause data loss 
 during crashes when allegedly flushed data is still being processed.
 
--- a/src/mount/fuse/mount_config.cc
+++ b/src/mount/fuse/mount_config.cc
@@ -91,6 +91,7 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("bandwidthoveruse=%lf", bandwidthoveruse, 1),
 	SFS_OPT("sfsdirentrycachesize=%u", direntrycachesize, 0),
 	SFS_OPT("nostdmountoptions", nostdmountoptions, 1),
+	SFS_OPT("DANGEROUS_sfsignoreflush", ignoreflush, 1),
 
 	SFS_OPT("enablefilelocks=%u", filelocks, 0),
 	SFS_OPT("nonempty", nonemptymount, 1),
@@ -221,6 +222,8 @@ void usage(const char *progname) {
 "    -o enablefilelocks=0|1      enables/disables global file locking "
 				"(disabled by default)\n"
 "    -o nonempty                 allow mounts over non-empty file/dir\n"
+"    -o DANGEROUS_sfsignoreflush  ignore flush/fsync usual behavior by "
+				"replying SUCCESS to it inmediately\n"
 "\n",
 		SaunaClient::FsInitParams::kDefaultUseRwLock,
 		SaunaClient::FsInitParams::kDefaultMkdirCopySgid,
--- a/src/mount/fuse/mount_config.cc
+++ b/src/mount/fuse/mount_config.cc
@@ -222,9 +222,9 @@ void usage(const char *progname) {
 "    -o enablefilelocks=0|1      enables/disables global file locking "
 				"(disabled by default)\n"
 "    -o nonempty                 allow mounts over non-empty file/dir\n"
-"    -o sfsignoreflush           DANGEROUS. Ignore flush usual behavior by "
-				"replying SUCCESS to it inmediately. Targets fast creation of "
-				"small files, but may cause data loss during crashes.\n"
+"    -o sfsignoreflush           Advanced: use with caution. Ignore flush usual "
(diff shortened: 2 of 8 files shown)


uristdwarf:
Seems fine (@lgsilva3087 should have a look before merging, I'm not entirely familiar with the Windows client), but since the CI doesn't work right now, we need to wait a bit.

In the meantime, you might want to copy edit the PR message a bit, since it's going to be squashed and included as the commit message. Someone else who's going to merge it can do it later though if you don't have time (since I can't merge it next week due to being on vacation).

src/mount/fuse/mount_config.cc
```diff
@@ -90,6 +90,7 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("bandwidthoveruse=%lf", bandwidthoveruse, 1),
 	SFS_OPT("sfsdirentrycachesize=%u", direntrycachesize, 0),
 	SFS_OPT("nostdmountoptions", nostdmountoptions, 1),
+	SFS_OPT("DANGEROUS_sfsignoreflush", ignoreflush, 1),
```
Antuan-AE, line 93: Consider renaming the variable to match naming conventions. If the DANGEROUS tag for this option is to be highlighted, consider doing it in the documentation (help command, man pages, etc).

src/mount/fuse/mount_config.cc
```diff
@@ -224,6 +225,8 @@ printf(
 "    -o enablefilelocks=0|1      enables/disables global file locking "
 				"(disabled by default)\n"
 "    -o nonempty                 allow mounts over non-empty file/dir\n"
+"    -o DANGEROUS_sfsignoreflush  ignore flush/fsync usual behavior by "
```
Antuan-AE, line 228: See my previous suggestion about tag DANGEROUS for naming the variable.

src/mount/sauna_client.h
```diff
@@ -79,8 +79,10 @@ struct FsInitParams {
 	static constexpr float    kDefaultBandwidthOveruse = 1.0;
 	static constexpr unsigned kDefaultChunkserverWriteTo = 5000;
 #ifdef _WIN32
+	static constexpr bool     kDefaultSpecialCopyMode = false;
 	static constexpr unsigned kDefaultWriteCacheSize = 50;
 #else
+	static constexpr bool     kDefaultIgnoreFlush = false;
```
Antuan-AE, line 85: Consider using the same variable name for Unix and Windows platforms as they are intended for the same purpose and have identical initialization values.

Antuan-AE (dismissed):
Please see my suggestions. LGTM

src/mount/sauna_client.cc
```diff
@@ -2133,6 +2134,15 @@ ReadCache::Result read(Context &ctx,
 		safs::log_debug("read from inode {} up to {} bytes from position {}",
 		                ino, size, off);
 	}
+#ifdef _WIN32
+	if (ignore_read) {
+		ReadCache::Entry *entry = new ReadCache::Entry(off, size);
```
dmga44, line 2139: This memory is not being freed (issue notice by @aNeutrino ).

src/mount/sauna_client.h
dmga44: Well, they are used for the equivalent options: an option which tries to fool file-by-file copiers into believing file are fast copied to make them send the request to write another file, and we could process those writes in parallel.

But the option worked differently for both OSs, for the Linux only had to mock the flush syscall, but under Windows had to mock also the read. So, I'll leave it this way because I don't want to mix both.

src/mount/fuse/mount_config.cc
dmga44: You point out to something here, which I have just checked. It seems we don't have updated our man pages, at least the readworkers and maxreadaheadrequests are not present there. I must add them also.

I'll rename it to sfsignoreflush, but as a first word in the help and man page comment I'll the word DANGEROUS. I think it is very important that people know they may lose data when using this option.

src/mount/writedata.cc
```diff
@@ -213,6 +213,7 @@ static std::condition_variable fcbcond;
 static uint32_t fcbwaiting = 0;
 static int64_t freecacheblocks;
 static inodedata **idhash;
+static bool ignoreflush = 0;
```
rolysr, line 216: Consider to set this to false as it is done in some other static bool variables around the repo.

rolysr (commented):
Can you consider if it would be necessary to add a test for this option?

uristdwarf (approved):

doc/sfsmount.1.adoc
```diff
@@ -204,6 +204,11 @@ Equivalent to *--nostdopts* (*-n*) option for use in fstab.
 *-o nonempty*::
 Equivalent to *--nonempty* option for use in fstab.
 
+*-o sfsignoreflush*::
+DANGEROUS. Ignore flush usual behavior by replying SUCCESS to it
```
lgsilva3087, line 208: It is strange to see DANGEROUS here. Probably we can change it to:
Advanced: use with caution.

src/mount/fuse/mount_config.cc
```diff
@@ -221,6 +222,9 @@ void usage(const char *progname) {
 "    -o enablefilelocks=0|1      enables/disables global file locking "
 				"(disabled by default)\n"
 "    -o nonempty                 allow mounts over non-empty file/dir\n"
+"    -o sfsignoreflush           DANGEROUS. Ignore flush usual behavior by "
+				"replying SUCCESS to it inmediately. Targets fast creation of "
+				"small files, but may cause data loss during crashes.\n"
```
lgsilva3087, line 227: Same here.

src/mount/fuse/mount_config.cc
```diff
@@ -91,6 +91,7 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("bandwidthoveruse=%lf", bandwidthoveruse, 1),
 	SFS_OPT("sfsdirentrycachesize=%u", direntrycachesize, 0),
 	SFS_OPT("nostdmountoptions", nostdmountoptions, 1),
+	SFS_OPT("sfsignoreflush", ignoreflush, 1),
```
lgsilva3087, line 94: Is it enabled by default or disabled?

lgsilva3087 (changes requested):
It is strange to have similar options with different names in Linux and Windows (ignore_flush and special_write_mode + ignore_read). I also don't like to ignore the reads, which makes the mount point write-only in Windows.
From your comments, looks like the Windows copier is triggering the read, if it is the case, we should not ignore the reads and just recommend another copier for small files.

src/mount/fuse/mount_config.cc
dmga44: It is disabled by default.

dmga44:
> just recommend another copier for small files.

Well, I've already performed extra tests and checked that Ultracopier, FastCopy and our custom sfscp also perform the extra read operation. Looks like it is even tied to the ```std::filesystem::copy``` which we are using in the sfscp tool. I was expecting some positive result from some of them, but it seems that the copier should be based on much lower level operations, like directs open, write, close.

src/mount/fuse/mount_config.cc
aNeutrino: yes lets remove DANGEROUS here....

doc/sfsmount.1.adoc
aNeutrino: maybe we can reuse the same text without DANGEROUS... from below:

Ignore flush usual behavior by replying SUCCESS to it immediately. Targets fast creation of small files, but may cause data loss during crashes."

src/mount/fuse/mount_config.cc
aNeutrino: @dmga44 typo in word: ~inmediately~ - should be: immediately

doc/sfsmount.1.adoc
dmga44: Done.

src/mount/fuse/mount_config.cc
dmga44: Done, both.

lgsilva3087 (approved):
LGTM.

lgsilva3087:
> LGTM.

Please, squash before merge.

PR #36, opened 2024-03-21
mount: Fix UB when clients reconnect

uristdwarf:
In commit 488ee3e, fs_connect contained a procedure to send client configuration, however this used fs_send_custom, which used fs_threc_flush that locks the FD. While this is fine when initially connecting, it causes problems when fs_receive_thread tries to create a new session when reconnecting, because it also locks the FD, causing undefined behaviour. This commit fixes this by bringing the function out of fs_connect, and unlocking the FD before calling the configuration function, and then locking it again. This fixes a lot of tests (see below for a list).

Also fixes an error in SFSCommunication.h introduced in the commit, which was noticed while fixing this issue.

Tests fixed:

* ShortSystemTests.test_metadata_recovery
* ShortSystemTests.test_multiple_master_promotions
* ShortSystemTests.test_shadow_metadata_generate
* ShortSystemTests.test_shadow_promotion_during_dumping
* ShortSystemTests.test_shadow_reloading_metadata
* ShortSystemTests.test_shadow_sessions
* Potentially may fix tests in LongSystemTests as well.

### DIFF
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -758,7 +758,6 @@ int fs_connect(bool verbose) {
 		mintrashtime = 0;
 		maxtrashtime = 0;
 	}
-	fs_register_config();
 	free(regbuff);
 	lastwrite=time(NULL);
 	if (!verbose) {
@@ -1170,6 +1169,9 @@ void* fs_receive_thread(void *) {
 			if (sessionlost) {      // if previous session is lost then try to register as a new session
 				if (fs_connect(false)==0) {
 					sessionlost=0;
+					fdLock.unlock();
+					fs_register_config();
+					fdLock.lock();
 				}
 			} else {        // if other problem occurred then try to resolve hostname and portname then try to reconnect using the same session id
 				if (fs_resolve(false, gInitParams.bind_host, gInitParams.host, gInitParams.port) == 0) {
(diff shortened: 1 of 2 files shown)


src/protocol/SFSCommunication.h
```diff
@@ -1239,7 +1239,7 @@ enum class SugidClearMode {
 /// msgid:32 status:8
 
 //0x01E6
-#define SAU_CLTOMA_REGISTER_CONFIG (1000U + 485U)
+#define SAU_CLTOMA_REGISTER_CONFIG (1000U + 486U)
```
dmga44, line 1242: This also looks nice. (Makes equal constant and comment, though it is probably not necessary for the fix you are doing).

dmga44 (approved):

src/mount/mastercomm.cc
```diff
@@ -1295,11 +1297,15 @@ int fs_init_master_connection(SaunaClient::FsInitParams &params
 		return 1;
 	}
 #ifdef _WIN32
-	sessionFlags = &session_flags; 
+	sessionFlags = &session_flags;
```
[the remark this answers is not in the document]
uristdwarf: This was actually accidental, my vim config automatically removes trailing whitespace when I save a file.

src/protocol/SFSCommunication.h
uristdwarf: The bug was related to this commit https://github.com/leil-io/saunafs/commit/488ee3e0228efd6ae39e0865ca6d48574606aeff. At first I thought the problem was here, but I was wrong. I still kept the fix in though, since it's still kinda related.

uristdwarf: I've updated the PR description

lgsilva3087 (approved):

PR #38, opened 2024-03-22
mount: Extend DirEntryCache to save requests to master

dmga44:
Client was not caching any coming data from master after succeeding in the lookup and getattr syscalls. This behavior may cause to issue extra requests to master for no good reason as that data can be cached.

To get it done, inserting single entry without full info (parent inode, name, index and next index) to the DirEntryCache must be implemented and therefore, untie its inner lookup and index sets.

This change is specially important in reducing the amount of lookup requests from Windows client to master server, with  huge impact in the small files read/write issue and performance of many other operations.

### DIFF
--- a/src/mount/direntry_cache.h
+++ b/src/mount/direntry_cache.h
@@ -21,6 +21,7 @@
 #include "common/platform.h"
 
 #include <atomic>
+#include <limits>
 
 #include "common/attributes.h"
 #include "common/shared_mutex.h"
@@ -31,9 +32,9 @@
 #include <boost/intrusive/list.hpp>
 #include <boost/intrusive/set.hpp>
 
-#define INVALID_INDEX 0xffff
-#define INVALID_PARENT 0xffffffff
-#define EMPTY_NAME ""
+constexpr uint64_t kInvalidIndex = std::numeric_limits<uint64_t>::max();
+constexpr uint32_t kInvalidParent = std::numeric_limits<uint32_t>::max();
+constexpr char kEmptyName[] = "";
 
 /*! \brief Cache for directory entries
  *
@@ -356,7 +357,7 @@ public:
 		if (inode_it != inode_multiset_.end()) {
 			erase(std::addressof(*inode_it));
 		}
-		addEntry(ctx, parent_inode, inode, INVALID_INDEX, INVALID_INDEX, name, attr, timestamp);
+		addEntry(ctx, parent_inode, inode, kInvalidIndex, kInvalidIndex, name, attr, timestamp);
 	}
 
 	/*! \brief Add directory entry information to cache.
@@ -377,7 +378,7 @@ public:
 		if (inode_it != inode_multiset_.end()) {
 			erase(std::addressof(*inode_it));
 		}
-		addEntry(ctx, INVALID_PARENT, inode, INVALID_INDEX, INVALID_INDEX, EMPTY_NAME, attr, timestamp);
+		addEntry(ctx, kInvalidParent, inode, kInvalidIndex, kInvalidIndex, kEmptyName, attr, timestamp);
 	}
 
 	/*! \brief Add data to cache from container.
@@ -574,11 +575,11 @@ public:
 
 protected:
 	void erase(DirEntry *entry) {
-		if (entry->parent_inode != INVALID_PARENT && !entry->name.empty()) {
+		if (entry->parent_inode != kInvalidParent && !entry->name.empty()) {
 			lookup_set_.erase(lookup_set_.iterator_to(*entry));
 		}
-		if (entry->parent_inode != INVALID_PARENT &&
-		    entry->index != INVALID_INDEX) {
+		if (entry->parent_inode != kInvalidParent &&
+		    entry->index != kInvalidIndex) {
 			index_set_.erase(index_set_.iterator_to(*entry));
 		}
 		inode_multiset_.erase(inode_multiset_.iterator_to(*entry));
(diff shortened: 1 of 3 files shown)


dmga44:
Is it ok now, @Antuan-AE ?

src/mount/direntry_cache.h
```diff
@@ -544,23 +607,38 @@ class DirEntryCache {
...
-		lookup_set_.insert(*entry);
-		auto result = index_set_.insert(*entry);
+	void addEntry(const SaunaClient::Context &ctx, uint32_t parent_inode,
+	              uint32_t inode, uint64_t index, uint64_t next_index,
+	              std::string name, Attributes attr, uint64_t timestamp) {
+		DirEntry *entry = new DirEntry(ctx, parent_inode, inode, index,
+		                               next_index, name, attr, timestamp);
+		if (parent_inode != INVALID_PARENT && !name.empty()) {
+			lookup_set_.insert(*entry);
+		}
```
Antuan-AE, line 620: Consider checking entry was properly allocated before inserting it to avoid possible crash of the system.

src/mount/sauna_client.cc
```diff
@@ -836,6 +854,7 @@ AttrReply getattr(Context &ctx, Inode ino) {
 	Attributes attr;
 	char attrstr[256];
 	int status;
+	bool fromcache;
```
Antuan-AE, line 857: Consider using camelCase for variable name.
eg: 
```
bool fromCache;
```

src/mount/sauna_client.cc
```diff
@@ -858,10 +877,12 @@ AttrReply getattr(Context &ctx, Inode ino) {
 		}
 		stats_inc(OP_DIRCACHE_GETATTR);
 		status = SAUNAFS_STATUS_OK;
+		fromcache = 1;
```
Antuan-AE, line 880: Consider using true/false in the assignment for the bool variable.

src/mount/direntry_cache.h
```diff
@@ -31,6 +31,10 @@
 #include <boost/intrusive/list.hpp>
 #include <boost/intrusive/set.hpp>
 
+#define INVALID_INDEX 0xffff
+#define INVALID_PARENT 0xffffffff
+#define EMPTY_NAME ""
+
```
lgsilva3087, line 37: Consider something like this:
```cpp
#include <limits>

constexpr uint32_t kInvalidIndex = std::numeric_limits<uint32_t>::max();
constexpr uint64_t kInvalidParent = std::numeric_limits<uint64_t>::max();
constexpr char kEmptyName[] = "";
```
This is a more modern and type-safe approach in C++. Plus, they are still compile-time constants like the #define macros.

src/mount/sauna_client.cc
```diff
@@ -878,9 +899,27 @@ AttrReply getattr(Context &ctx, Inode ino) {
...
+		std::unique_lock<shared_mutex> write_guard(gDirEntryCache.rwlock());
+		gDirEntryCache.updateTime();
+
+		gDirEntryCache.insert(ctx, ino, attr, data_acquire_time);
+		if (gDirEntryCache.size() > gDirEntryCacheMaxSize) {
+			gDirEntryCache.removeOldest(gDirEntryCache.size() -
+			                            gDirEntryCacheMaxSize);
+		}
+
+		write_guard.unlock();
```
lgsilva3087, line 916: This lock should be automatically released here. Please consider removing the explicit unlock.

src/mount/sauna_client.cc
```diff
@@ -878,9 +899,27 @@ AttrReply getattr(Context &ctx, Inode ino) {
...
+	}
+
 	makeattrstr(attrstr,256,&o_stbuf);
-	oplog_printf(ctx, "getattr (%lu): OK (%.1f,%s)",
+	oplog_printf(ctx, "getattr (%lu)%s: OK (%.1f,%s)",
 			(unsigned long int)ino,
+			fromcache?" (using open dir cache)":"",
```
lgsilva3087, line 922: Please add the spaces:
```cpp
fromcache ? " (using open dir cache)" : "",
```

lgsilva3087 (commented):
Nice change, please see my comments.

src/mount/direntry_cache_unittest.cc
```diff
@@ -143,20 +195,36 @@ TEST(DirEntryCache, Repetitions) {
...
+	    current_time);
+	// Lookup set finds the entries and should overwrite them
+	cache.insertSequence(
+	    SaunaClient::Context(0, 0, 0, 0), 9,
+	    std::vector<DirectoryEntry>{{1, 2, 8, "a1", dummy_attributes},
+	                                {2, 3, 9, "a2", dummy_attributes},
+	                                {3, 4, 10, "a3", dummy_attributes},
+	                                {4, 5, 11, "a4", dummy_attributes},
+	                                {5, 6, 12, "a5", dummy_attributes},
+									{6, 7, 13, "a6", dummy_attributes}},
```
dmga44, line 215: Must fix alignment here.

src/mount/sauna_client.cc
dmga44: Done.

src/mount/sauna_client.cc
dmga44: Done.

src/mount/direntry_cache.h
dmga44: Applied.

src/mount/sauna_client.cc
dmga44: Done.

src/mount/direntry_cache_unittest.cc
dmga44: Done.

src/mount/direntry_cache.h
```diff
@@ -320,9 +325,62 @@ class DirEntryCache {
...
+			return;
+		}
+		removeExpired(1, timestamp);
+		auto inode_it = find(ctx, inode);
+		if (inode_it != inode_multiset_.end()) {
+			erase(std::addressof(*inode_it));
+		}
+		addEntry(ctx, kInvalidParent, inode, kInvalidIndex, kInvalidIndex, kEmptyName, attr, timestamp);
+	}
+
```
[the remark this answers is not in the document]
lgsilva3087: We agree to do it in a separate change.

lgsilva3087 (approved):

PR #39, opened 2024-03-23
mount (win): Fix .saunafs_tweaks behavior

rolysr:
Attempting to update the value of a variable from the `.saunafs_tweaks` file under some Windows machines has been unsuccessful due to a read operation being executed after the corresponding write operation. This resulted in the changes from the write operation not being reflected because the `getAllValues` function, which resets the `.saunafs_tweaks` values to their defaults and did not check if a write operation had been executed.

This issue was resolved by adding a check in the read operation function related to `.saunafs_tweaks`. If the file had been written to, a procedure is initiated that constructs the expected string format and applies the correct update operation on the `.saunafs_tweaks` data, according to the format of the input from PowerShell commands.

### DIFF
--- a/src/mount/special_release.cc
+++ b/src/mount/special_release.cc
@@ -73,6 +73,12 @@ namespace InodeTweaks {
 static void release(FileInfo *fi) {
 	MagicFile *file = reinterpret_cast<MagicFile*>(fi->fh);
 	if (file->wasWritten) {
+#ifdef _WIN32
+		if (isWStringFromWindows(file->value)) {
+			file->value = convertWStringFromWindowsToString(file->value);
+		}
+#endif
+
 		auto separatorPos = file->value.find('=');
 		if (separatorPos == file->value.npos) {
 			safs_pretty_syslog(LOG_INFO, "TWEAKS_FILE: Wrong value '%s'",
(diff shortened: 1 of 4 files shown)


src/mount/special_release.cc
```diff
@@ -73,28 +73,26 @@ namespace InodeTweaks {
...
+		    file->value[0] == -1) {  // Powershell command
+			// Given the special format of the Powershell input, this patch
+			// must be applied
+			std::string value;
+			for (int i = 2; i < (int)file->value.size(); i += 2) {
+				value.push_back(file->value[i]);
+			}
+			file->value = value;
+		}
+#endif
```
lgsilva3087, line 87: Looks like the behavior of a wide string (std::wstring). Please check.

Antuan-AE (approved):
Consider adding a functional test on Windows to check the new implementation. LGTM.

src/mount/special_release.cc
```diff
@@ -103,6 +94,7 @@ static void release(FileInfo *fi) {
 			                   name.c_str(), value.c_str());
 		}
 	}
+
```
dmga44, line 97: Maybe you should remove this line.

rolysr: Done

uristdwarf (approved):

PR #40, opened 2024-03-25
tests: Fix test new mount vs old saunafs

lgsilva3087:
It was failing for the following reasons:
- The previous version uses SAU as EOF marker for the metadata file insted of the new SFS.
- The new admin tool was also expecting an extra item in the 'admin info' command.
- The sfs* tools are deprecated (sfsgetgoal does not exists).

This applies the following fixes:
- Also check for old signature in the metadata EOF marker.
- Use saunafs_old_admin_master instead of saunafs_admin_master.
- Use, for instance, 'saunafsXX saunafs setgoal' instead of 'saunafsXX sfssetgoal'.

### DIFF
--- a/src/common/metadata.cc
+++ b/src/common/metadata.cc
@@ -90,12 +90,17 @@ uint64_t metadataGetVersion(const std::string& file) {
 		throw MetadataCheckException("Can't read the metadata file");
 	}
 
-	if (memcmp(chkbuff, SFSSIGNATURE "M 2.9", strlen(SFSSIGNATURE "M 2.9")) == 0) {
+	std::string signature = std::string(chkbuff, 8);
+	std::string sfsSignature = std::string(SFSSIGNATURE "M 2.9");
+	std::string sauSignature = std::string(SAUSIGNATURE "M 2.9");
+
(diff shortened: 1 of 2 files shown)


src/common/metadata.cc
```diff
@@ -90,12 +90,17 @@ uint64_t metadataGetVersion(const std::string& file) {
 		throw MetadataCheckException("Can't read the metadata file");
 	}
 
-	if (memcmp(chkbuff, SFSSIGNATURE "M 2.9", strlen(SFSSIGNATURE "M 2.9")) == 0) {
+	std::string signature = std::string(chkbuff, 8);
+	std::string sfsSignature = std::string(SFSSIGNATURE "M 2.9");
+	std::string sauSignature = std::string(SAUSIGNATURE "M 2.9");
+
+	if (signature == sfsSignature || signature == sauSignature) {
```
Antuan-AE, line 97: I had already fixed this in the dump metadata [branch](https://github.com/leil-io/saunafs/commit/b84b824637354f3f0330f80691c5a36d2aeb6417) but is not finish yet. LGTM.

Antuan-AE (approved):

PR #41, opened 2024-03-26
tests: Fix test_saunafs_upgrade_ec_basic

lgsilva3087:
The test is now using the new tools and we 'trust' the gpg key of our own repo.


Antuan-AE (approved):
LGTM

PR #42, opened 2024-03-30
all: Fix build under ENABLE_TRACES

lgsilva3087:
Some of the calls to the traces macros were not updated in previous commits. This change fixes the building by updating these calls with correct parameters and extends the TracePrinter to optionally output to our usual syslog.

To use the traces, just enable the option ENABLE_TRACES in the cmake call. By default it prints to stdout, so, to print to the syslog, the option ENABLE_SYSLOG_FOR_TRACES must be enabled too (useful when running the tests).

To disable the traces for specific modules, add the following line in the module cmake file:

remove_definitions(-DENABLE_SYSLOG_FOR_TRACES)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,7 @@ option(ENABLE_POLONAISE         "Enable polonaise"                              
 option(ENABLE_CLIENT_LIB        "Enable dynamic client library (libsaunafs-client)"     OFF)
 # end of SaunaFS options
 option(ENABLE_TRACES            "Enable traces"                                          OFF)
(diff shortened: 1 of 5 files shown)


dmga44 (commented):
Looks good to me. There is still the issue of it not passing CI, but I guess it will be fixed soon.

I have one small add up, maybe you could consider this idea. Set the ENABLE_TRACES compiler to a numerical value: 0,1 or 2. 0 does not uses traces, 1 uses oplog and 2 syslog. This way we could even have other types of traces in the future without adding more options.

Antuan-AE (approved):
LGTM

rolysr (approved):
It is ok for me

PR #43, opened 2024-04-01
mount: Add client option feature

Antuan-AE:
The DirectIO flag on the client was set only thought the hidden tweak file.
A mount option is added to set DirectIO mode at client initialization stage.
A functional test is added to test the new feature.
A duplicated message from usage help command is removed at client.
The version (-v|-V) flag is set case insensitive on every module.

### DIFF
--- a/tests/test_suites/SanityChecks/test_mount_directio.sh
+++ b/tests/test_suites/SanityChecks/test_mount_directio.sh
@@ -1,0 +1,21 @@
+#
+# To run this test you need to add the following lines to /etc/sudoers.d/saunafstest:
+#
+# saunafstest ALL = NOPASSWD: /usr/bin/tee
+#
+#
+MOUNTS=2 \
+  USE_RAMDISK=YES \
+  MOUNT_0_EXTRA_CONFIG="sfsdirectio=0" \
+  MOUNT_1_EXTRA_CONFIG="sfsdirectio=1" \
+  setup_local_empty_saunafs info
+
+expect_equals "false" "$(cat ${info[mount0]}/.saunafs_tweaks | grep -i directio | awk '{print $2}')"
+expect_equals "true" "$(cat ${info[mount1]}/.saunafs_tweaks | grep -i directio | awk '{print $2}')"
+
+# Modify the DirectIO option on the fly
+echo "DirectIO=true" | sudo tee ${info[mount0]}/.saunafs_tweaks
+echo "DirectIO=false" | sudo tee ${info[mount1]}/.saunafs_tweaks
+
+expect_equals "true" "$(cat ${info[mount0]}/.saunafs_tweaks | grep -i directio | awk '{print $2}')"
+expect_equals "false" "$(cat ${info[mount1]}/.saunafs_tweaks | grep -i directio | awk '{print $2}')"
--- a/src/common/read_plan_executor.h
+++ b/src/common/read_plan_executor.h
@@ -80,13 +80,13 @@ public:
 		return networking_failures_;
 	}
 
-	/// Counter for the .saunafds_tweaks file.
+	/// Counter for the .saunafs_tweaks file.
 	static std::atomic<uint64_t> executions_total_;
 
-	/// Counter for the .saunafds_tweaks file.
+	/// Counter for the .saunafs_tweaks file.
 	static std::atomic<uint64_t> executions_with_additional_operations_;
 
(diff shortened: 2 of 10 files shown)


tests/test_suites/SanityChecks/test_mount_directio.sh
uristdwarf, line 1: Is this test really necessary? It just checks if saunafs_tweaks contains the configuration value, not actually testing any functionality right now.

uristdwarf (changes requested):
I don't quite understand the change, is this a WIP for directIO, since this variable isn't used anywhere? Is it used by the Windows client?

Also, the first two commits aren't really related to this PR, they should have been done as separate PR's. But these aren't really deal breakers.

rolysr (changes requested):
Please check last line of this [commit message](https://github.com/leil-io/saunafs/pull/43/commits/a259aeaf8962f6e79adacb1d796b09de3ca7d8b2) with 75 chars lenght and try to reduce it to 72 as the 50-70 convention we are using for commit titles and descriptions.

uristdwarf:
> I don't quite understand the change, is this a WIP for directIO, since this variable isn't used anywhere? Is it used by the Windows clients

I checked that directIO is a feature in FUSE, so I think we are just enabling it as an option? Please correct me if I'm wrong, but if I am not, maybe make it a bit more clear in the commit message :)

I still think we don't need a test for this though, unless there's some actual functionality we are testing.

tests/test_suites/SanityChecks/test_mount_directio.sh
```diff
@@ -0,0 +1,21 @@
+#
+# To run this test you need to add the following lines to /etc/sudoers.d/saunafstest:
+#
+# saunafstest ALL = NOPASSWD: /usr/bin/tee /tmp/SaunaFS-autotests/mnt/sfs*
```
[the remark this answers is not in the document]
Antuan-AE: setup_machine was updated to allow client saunafstest user write access to hidden .saunafs_tweaks file.

tests/test_suites/SanityChecks/test_mount_directio.sh
Antuan-AE: The main idea for this test, is first checking the new mount option ```sfsdirectio``` can be set for each client mount-point independently. At the same time the test is validating the option was properly set after client was mounted by querying the hidden tweak file and the previous implementation to modify the  option is still available. What this option is doing to improve performance is beyond the scope of this simple sanity test and could be address maybe at short system test.

Antuan-AE:
> I don't quite understand the change, is this a WIP for directIO, since this variable isn't used anywhere? Is it used by the Windows client?
> 
> Also, the first two commits aren't really related to this PR, they should have been done as separate PR's. But these aren't really deal breakers.

This is a feature for linux client to use the maximum block size at fuse layer for reading files thus improving performance in some specific scenarios. This option was already available at the client with a hard coded default value that could be on the fly modified by writing to the hidden tweaks file. This commit intents to also make this available as a mount option. First two commit although not related were small enough to be included as they are only fixing/improving client help menu and will be squashed in a single commit after PR to merge.

Antuan-AE:
> Please check last line of this [commit message](https://github.com/leil-io/saunafs/pull/43/commits/a259aeaf8962f6e79adacb1d796b09de3ca7d8b2) with 75 chars lenght and try to reduce it to 72 as the 50-70 convention we are using for commit titles and descriptions.

Fixed.

rolysr (approved):

tests/test_suites/SanityChecks/test_mount_directio.sh
uristdwarf: If this is the only one that has both a saunafs_tweaks and a mount option, then I guess it's okay for now. But in the future this could be justified for every such option, so maybe in the future this test could expanded to include other options as well (otherwise, we might end up with many similar tests)

uristdwarf:
> This is a feature for linux client to use the maximum block size at fuse layer for reading files thus improving performance in some specific scenarios. This option was already available at the client with a hard coded default value that could be on the fly modified by writing to the hidden tweaks file. This commit intents to also make this available as a mount option. First two commit although not related were small enough to be included as they are only fixing/improving client help menu and will be squashed in a single commit after PR to merge.

Thanks for the detailed explanation, makes much more sense now. I've no further complaints.

PR #45, opened 2024-04-04
fix(tests): Add missing dependancy for setup_machine.sh

uristdwarf:


uristdwarf:
> This commit is also including these dependencies: https://github.com/leil-io/saunafs/commit/a791f722463c624a310629986517bdd7c52c0c39

In that case I'll leave it for you to handle :-)

PR #48, opened 2024-04-18
chore(.gitignore): Add .dockerignore and update .gitignore

uristdwarf:
.dockerignore allows docker images building from the source context to ignore files, including build files and .git, which can balloon the context and significantly slow down image build times. The file is very similar to .gitignore, with minor changes specific for docker

On that note, .gitignore has been updated to not include files that are made by setup_machine.sh, and its aesthetics has been tidied up.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,12 +1,6 @@
 *~
 .DS_Store
-
-# External libraries #
-######################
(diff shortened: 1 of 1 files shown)


Antuan-AE (approved):
LGTM

PR #52, opened 2024-04-18
tests: Fix metadata ACL tests

ictus4u:
The test was failing by having wrong argument to the setfacl command.


Antuan-AE (approved):
Great Catch. LGTM

uristdwarf:
Why does this test pass on my PC :thinking:

uristdwarf:
Ok, that makes sense, since I'm using Ubuntu 22.04. And since it works from my PC, I guess it's good!

uristdwarf (approved):

PR #54, opened 2024-04-20
mount: Reduce readings during ec write

rolysr:
Tests performed show that during EC writes the client connects and asks
to the chunkservers for the missing data parts of the current stripe,
which are the data parts the client are not writing. This is required if
the file is bigger than the last byte it is writing, but it should not
ask for data on offsets beyond the file size. It is specially important
for the writings of small files, and removing those extra requests
should reduce overhead.

Current solution manages to calculate chunkSizeInBlocks_ for each
ChunkWriter at the moment of initialization from the field maxfleng of
the current inode's inodedata. Thus, the maxfleng field should be
updated the best way possible. The chunkSizeInBlocks_ is then used to
check if client ws about to was to chunkservers for not existing data.

### DIFF
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -775,6 +775,8 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
 	LOG_AVG_TILL_END_OF_SCOPE0("hddRead");
 	TRACETHIS3(chunkId, offset, size);
 
+	safs_silent_syslog(LOG_DEBUG, "chunkserver.hddRead");
+
 	uint32_t offsetWithinBlock = offset % SFSBLOCKSIZE;
 
 	if ((size == 0) || ((offsetWithinBlock + size) > SFSBLOCKSIZE)) {
--- a/src/mount/chunk_locator.h
+++ b/src/mount/chunk_locator.h
@@ -97,6 +97,10 @@ public:
 	virtual void locateAndLockChunk(uint32_t inode, uint32_t index);
 	virtual void unlockChunk();
 
+	uint64_t fileLength() const { 
+		return locationInfo_.fileLength; 
+	}
+
 	uint32_t chunkIndex() {
 		return index_;
 	}
--- a/src/mount/chunk_locator.h
+++ b/src/mount/chunk_locator.h
@@ -97,7 +97,7 @@ public:
 	virtual void locateAndLockChunk(uint32_t inode, uint32_t index);
 	virtual void unlockChunk();
 
-	uint64_t fileLength() const { 
+	inline uint64_t fileLength() const { 
 		return locationInfo_.fileLength; 
 	}
 
--- a/src/mount/chunk_writer.cc
+++ b/src/mount/chunk_writer.cc
@@ -126,7 +126,7 @@ ChunkWriter::~ChunkWriter() {
 	}
 }
 
-void ChunkWriter::init(WriteChunkLocator* locator, uint32_t chunkserverTimeout_ms, uint32_t chunkSize) {
+void ChunkWriter::init(WriteChunkLocator* locator, uint32_t chunkserverTimeout_ms) {
 	LOG_AVG_TILL_END_OF_SCOPE0("ChunkWriter::init");
 	sassert(pendingOperations_.empty());
 	sassert(executors_.empty());
@@ -134,8 +134,6 @@ void ChunkWriter::init(WriteChunkLocator* locator, uint32_t chunkserverTimeout_m
 	Timeout connectTimeout{std::chrono::milliseconds(chunkserverTimeout_ms)};
 	combinedStripeSize_ = 0;
 	locator_ = locator;
-	chunkSizeInBlocks_ =
-	    (chunkSize == 0 ? 0 : (chunkSize - 1) / SFSBLOCKSIZE + 1);
 
 	for (const ChunkTypeWithAddress& location : locator_->locationInfo().locations) {
 		// If we have an executor writing the same chunkType, use it
@@ -330,6 +328,11 @@ std::list<WriteCacheBlock> ChunkWriter::releaseJournal() {
 	return std::move(journal_);
 }
 
+void ChunkWriter::setChunkSizeInBlocks(uint32_t chunkSize) {
+	chunkSizeInBlocks_ =
+	    (chunkSize == 0 ? 0 : (chunkSize - 1) / SFSBLOCKSIZE + 1);
+}
+
 void ChunkWriter::addOperation(WriteCacheBlock&& block) {
 	sassert(block.type != WriteCacheBlock::kParityBlock);
 	sassert(acceptsNewOperations_);
@@ -441,6 +444,7 @@ void ChunkWriter::fillOperation(Operation &operation, int first_block, int first
 /*!
  * Fills given operation with a range of blocks beyond file size required to be read.
  * \param operation operation to be filled
+ * \param first_block first block of the operation
  * \param first_index index of the first block required to be read
  * \param size number of required blocks
  * \param stripe_element map of blocks in a stripe
@@ -453,12 +457,14 @@ void ChunkWriter::fillNotExisting(Operation &operation, int first_block, int fir
 	}
 	int block_from = operation.journalPositions.front()->from;
 	int block_to = operation.journalPositions.front()->to;
+	int beyond_start = first_block + first_index;
 
 	std::vector<WriteCacheBlock> blocks;
 	blocks.reserve(size);
 	// Fills not existing blocks with zeroes
-	for (int index = first_block; index < first_block + size; ++index) {
-		WriteCacheBlock block(locator_->chunkIndex(), index, WriteCacheBlock::kReadBlock);
+	for (int index = beyond_start; index < beyond_start + size; ++index) {
+		WriteCacheBlock block(locator_->chunkIndex(), index,
+		                      WriteCacheBlock::kReadBlock);
 		memset(block.data(), 0, SFSBLOCKSIZE);
 		block.from = block_from;
 		block.to = block_to;
--- a/src/mount/chunk_writer.cc
+++ b/src/mount/chunk_writer.cc
@@ -126,7 +126,7 @@ ChunkWriter::~ChunkWriter() {
 	}
 }
 
-void ChunkWriter::init(WriteChunkLocator* locator, uint32_t chunkserverTimeout_ms) {
+void ChunkWriter::init(WriteChunkLocator* locator, uint32_t chunkserverTimeout_ms, uint32_t chunkSize) {
 	LOG_AVG_TILL_END_OF_SCOPE0("ChunkWriter::init");
 	sassert(pendingOperations_.empty());
 	sassert(executors_.empty());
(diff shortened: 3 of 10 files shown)


aNeutrino (changes requested):
Missing 2..3 automated test suits which would test new functionality.

uristdwarf (commented):
I'm not very familiar in this area, so I won't approve or request changes specifically for approval, but I do have a few comments.

1. If it's possible, try to make a unit test for `fillNotExisting`. I say if possible because I understand there may be certain situations it's not because of external dependencies. But on first glance I would try, since we really lack unit tests and this is a good candidate.
2. This will probably come in a later release (4.2 most likely), so this probably won't be merged before Wednesday.

uristdwarf:
Short System tests `test_auto_recovery_incversion`, `test_auto_recovery_nextchunkid` and `test_multiple_master_promotions` are failing on my system, compared to dev

src/mount/sauna_client.cc
```diff
@@ -2359,7 +2359,18 @@ BytesWritten write(Context &ctx, Inode ino, const char *buf, size_t size, off_t
...
 		fileinfo->data = write_data_new(ino);
 	}
-	err = write_data(fileinfo->data,off,size,(const uint8_t*)buf);
+
+	Attributes attr;
+	attr.fill(0);
+	if (usedircache) { 
+		gDirEntryCache.lookup(ctx, ino, attr); 
+	}
+	stat *stbuf = new stat;
```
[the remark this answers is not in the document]
uristdwarf: Ah dangit, didn't see it was fixed in the next commit in the series. Sorry

src/chunkserver/hddspacemgr.cc
```diff
@@ -775,6 +775,8 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
 	LOG_AVG_TILL_END_OF_SCOPE0("hddRead");
 	TRACETHIS3(chunkId, offset, size);
 
+	safs_silent_syslog(LOG_DEBUG, "chunkserver.hddRead");
```
dmga44, line 778: Is this line required by the test?

dmga44:
> @rolysr do we have test which covers this case? Eg. 2 separate mount points Create file and Set gooal EC(4,2)
> 
> write 64KiB...128KiB byte on byte 0 on 1st mount point (part 1 and/or 2) In parallel (same time) 2nd mount point should write 1byte/64KiB to position 192KiB (part 4 only) or position 128KiB (part 3 only)
> 
> Above is example of 6 test cases Which must check that one part is(3 or 4) never read and final file include all data from parts 1,2 and (3 or 4).
> 
> I do not see any new tests in your PR...it seems wrong.
> 
> I would like to se 2 or 3 test suits as proposed above.

I would like to comment a bit here.

To my understanding, you are asking for a test which after creating a file and setting its goal, performs in parallel the following two operations:
- write a number of bytes from 64KiB to 128KiB to the file starting from offset 0 (meaning writes on blocks 1 and/or 2)
- write a number of bytes from 1 to 64KiB (it doesn't really matter which) to position 192KiB (meaning writes on the block 4) or equivalent for block 3

So, I think we have two scenarios:
- the mountpoint 1 is the first one to lock the chunk and performs its writes (block 1 and/or 2) without reading anything extra from the chunkservers, and calculates the parity parts. The mountpoint 2 then performs its writes (block 4), but to calculate the parity parts, it needs to read from the previosly written blocks from the first mountpoint.
- the mountpoint 2 is the first one to lock the chunk and performs its writes (block 4) and to calculate the parity parts it needs to read the first 3 blocks, which should be full of zeroes. The mountpoint 1 then performs its writes (blocks 1 and/or 2), and to calculate the parity parts, it needs to read from the previosly existing blocks from the stripe (blocks 3 (full of zeroes) and 4 (some content)).

Anyway we'll see some readings to calculate parities.

src/chunkserver/hddspacemgr.cc
rolysr: Yes, that line is used for checking the extra read operations

tests/test_suites/ShortSystemTests/test_ec_truncate_atomicity.sh
```diff
@@ -0,0 +1,36 @@
...
+done
+# In the meanwhile, shorten file from 400K to 200K, chopping 1 kB in each of 200 steps
+# Sometimes share mountpoint with dd (mount2), sometimes not (mount3, mount4)
+for i in {399..200}; do
+	truncate -s ${i}K "${info[mount$((2 + i % 3))]}/$file"
+done
+wait # Wait for all dd processes to finish
+
+# Now file should be equal to the source file. Let's validate it!
+MESSAGE="Testing ec(4,2) config" assert_success file-validate "$file"
```
lgsilva3087, line 36: This tests is similar to test_xor_truncate_atomicity. In the original test, the first stripe was removed in this point to check if the system can read also from the parities. Please, consider doing the same.

tests/test_suites/ShortSystemTests/test_reads_during_small_ec_files_write_alternated.sh
```diff
@@ -0,0 +1,39 @@
+timeout_set 120 seconds
+
+CHUNKSERVERS=6 \
+	MOUNTS=2 \
+	USE_ZONED_DISKS=YES \
```
lgsilva3087, line 5: Same here. This repo should not contain specific SMR tests.

tests/test_suites/ShortSystemTests/test_reads_during_small_ec_files_write_alternated.sh
```diff
@@ -0,0 +1,39 @@
+timeout_set 120 seconds
```
lgsilva3087, line 1: How much time is consuming this test? 120 seconds is a long time.

lgsilva3087 (changes requested):
Please see my comments in the tests.
This change was originated because there was a lot of reading during creation of small files in EC. Please consider adding a test that creates 100 small EC files (33 bytes) and check that there are no unneeded reads.

src/mount/chunk_locator.h
```diff
@@ -97,6 +97,10 @@ class WriteChunkLocator {
 	virtual void locateAndLockChunk(uint32_t inode, uint32_t index);
 	virtual void unlockChunk();
 
+	uint64_t fileLength() const {
```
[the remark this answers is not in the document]
rolysr: Done

src/mount/chunk_writer.cc
```diff
@@ -328,6 +328,11 @@ std::list<WriteCacheBlock> ChunkWriter::releaseJournal() {
 	return std::move(journal_);
 }
 
+void ChunkWriter::setChunkSizeInBlocks(uint32_t chunkSize) {
```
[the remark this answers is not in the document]
rolysr: Done

src/mount/chunk_writer.cc
```diff
@@ -436,6 +441,46 @@ void ChunkWriter::fillOperation(Operation &operation, int first_block, int first
...
 
+/*!
+ * Fills given operation with a range of blocks beyond file size required to be read.
+ * \param operation operation to be filled
+ * \param first_block first block of the operation
+ * \param first_index index of the first block required to be read
+ * \param size number of required blocks
+ * \param stripe_element map of blocks in a stripe
+ */
+void ChunkWriter::fillNotExisting(Operation &operation, int first_block, int first_index, int size,
```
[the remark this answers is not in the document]
rolysr: Done

src/mount/chunk_writer.cc
```diff
@@ -436,6 +441,46 @@ void ChunkWriter::fillOperation(Operation &operation, int first_block, int first
...
+		std::vector<uint8_t *> &stripe_element) {
+	assert(size >= 0);
```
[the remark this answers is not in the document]
rolysr: There is no code calling that function with 0 blocks.

src/mount/chunk_writer.h
```diff
@@ -127,6 +127,8 @@ class ChunkWriter {
 
 	bool acceptsNewOperations() { return acceptsNewOperations_; }
 
+	void setChunkSizeInBlocks(uint32_t chunkSize);
```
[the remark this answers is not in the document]
rolysr: Done

tests/test_suites/ShortSystemTests/test_ec_truncate_atomicity.sh
rolysr: Done

tests/test_suites/ShortSystemTests/test_reads_during_small_ec_files_write.sh
```diff
@@ -0,0 +1,85 @@
+timeout_set 120 seconds
+
+CHUNKSERVERS=6 \
+	MOUNTS=2 \
+	USE_ZONED_DISKS=YES \
```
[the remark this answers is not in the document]
rolysr: Done

tests/test_suites/ShortSystemTests/test_reads_during_small_ec_files_write_alternated.sh
rolysr: Done

tests/test_suites/WindowsClientTests
lgsilva3087, line 1: This soft link should not be here. Please verify.

lgsilva3087 (changes requested):
Looks good. Please verify the soft link.

dmga44 (commented):
LGTM, though I'm coauthor.

src/mount/chunk_writer.cc
```diff
@@ -436,6 +436,48 @@ void ChunkWriter::fillOperation(Operation &operation, int first_block, int first
...
 
+/*!
+ * Fills given operation with a range of blocks beyond file size required to be read.
+ * \param operation operation to be filled
+ * \param first_block first block of the operation
+ * \param first_index index of the first block required to be read
+ * \param size number of required blocks
+ * \param stripe_element map of blocks in a stripe
+ */
+void ChunkWriter::fillNotExisting(Operation &operation, int first_block,
```
Antuan-AE, line 447: Consider declaring first_block parameter as unsigned integer.

src/mount/chunk_writer.h
```diff
@@ -175,6 +180,7 @@ class ChunkWriter {
 	bool acceptsNewOperations_;
 	int combinedStripeSize_;
 	int dataChainFd_;
+	int chunkSizeInBlocks_;
```
Antuan-AE, line 183: Consider declaring this variable as unsigned interger as negative values have not practical meaning for this variable.

src/mount/sauna_client.cc
```diff
@@ -2359,7 +2359,17 @@ BytesWritten write(Context &ctx, Inode ino, const char *buf, size_t size, off_t
...
+	Attributes attr;
+	attr.fill(0);
+	if (usedircache) {
+		gDirEntryCache.lookup(ctx, ino, attr); 
+	}
+	struct stat stbuf;
+	attr_to_stat(ino, attr, &stbuf);
+	size_t currentSize = stbuf.st_size;
+
+	err = write_data(fileinfo->data,off,size,(const uint8_t*)buf, currentSize);
```
Antuan-AE, line 2372: Consider adding a space between comma separated parameters to comply with code style.

Antuan-AE (approved):
LGTM

src/mount/chunk_writer.cc
lgsilva3087: We will leave it for now to match with already existing code.
image

src/mount/chunk_writer.h
lgsilva3087: Same here.

src/mount/sauna_client.cc
lgsilva3087: Done.

lgsilva3087:
@aNeutrino your requested changes seem to be addressed with the newly added tests. Please re-check.

aNeutrino (approved):
LGTM

PR #55, opened 2024-04-23
fix(tests): Various fixes for the testing framework 

uristdwarf:
This is a series of 3 commits:

tidy(tests): Remove unnecessary ticks

This messes up some editor syntax highlighting

fix(tests): Fix binary output being outputted

The variable used in the conditional is different from the actual
filename used. This conditional would always succeed since awk would
return 0 if the conditional in awk would not match, which it didn't
since stdout was empty (only stderr was populated with the error
message, but it is disregarded).

fix(tests): Coredumps not being generated

At least 3 years ago, `sudo` had a change of behaviour regarding
coredumps. This fixes that by setting the default for cores as the
system.

### DIFF
--- a/tests/tools/test.sh
+++ b/tests/tools/test.sh
@@ -86,8 +86,8 @@ debug_command() {
 	fi
 	if (( depth <= DEBUG_LEVEL )); then
 		local indent=">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
-		echo "${indent:0:$depth} $(basename "${BASH_SOURCE[1]}"):${BASH_LINENO[0]}:`
-				`${FUNCNAME[1]}(): ${BASH_COMMAND}" >&2 | true
+		echo "${indent:0:$depth} $(basename "${BASH_SOURCE[1]}"):${BASH_LINENO[0]}:
+				${FUNCNAME[1]}(): ${BASH_COMMAND}" >&2 | true
 	fi
 }
 
--- a/tests/tools/test.sh
+++ b/tests/tools/test.sh
@@ -86,8 +86,12 @@ debug_command() {
 	fi
 	if (( depth <= DEBUG_LEVEL )); then
 		local indent=">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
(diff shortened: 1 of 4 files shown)


uristdwarf:
This PR is dependent on #48

tests/tools/test.sh
```diff
@@ -86,8 +86,8 @@ debug_command() {
 	fi
 	if (( depth <= DEBUG_LEVEL )); then
 		local indent=">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
-		echo "${indent:0:$depth} $(basename "${BASH_SOURCE[1]}"):${BASH_LINENO[0]}:`
-				`${FUNCNAME[1]}(): ${BASH_COMMAND}" >&2 | true
```
aNeutrino, line 90: wow what this double grave `` were here for?

aNeutrino (approved):
LGTM

tests/tools/test.sh
uristdwarf: Indeed, this change does seem to affect at least this particular one:

```
+++test.sh:90:debug_command(): echo '>> timeout.sh:71:
				timeout_set(): timeout_set 2 minutes'
>> timeout.sh:71:
				timeout_set(): timeout_set 2 minutes
+++timeout.sh:72:timeout_set(): debug_command
```

I would still like to avoid using `, since they are hard for even other parsers to correctly detect (like from github in the issue above)

uristdwarf: Best solution in my opinion is just to make them variables instead, also improves readability:

```bash
		local indent=">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
		local level=${indent:0:$depth}
		local source=$(basename "${BASH_SOURCE[1]}")
		local line=${BASH_LINENO[0]}
		local function=${FUNCNAME[1]}
		local command=${BASH_COMMAND}
		echo "${level} ${source}:${line}: ${function}(): ${command}" >&2 | true
```

uristdwarf:
> Please, reword the commit message (maybe you intend to squash both commits, but the current message refers other commits that are not there anymore). Also, notice some of these changes are not atomic (in a sense that if you need to revert a commit, it wouldn't produce collateral effects)

I've reworded it a bit so it looks like a full commit. On the last part, I mostly agree but these are 2 rather trivial changes not related to the main codebase (only relating to tests), so it's not worth the effort to split these apart, especially since the last point on the bullet list is dependent on the second.

PR #56, opened 2024-04-24
tests: Fix long tests to build project

Antuan-AE:
- Update the repository path.
- Add disks to setup script for Long tests that requieres 8 disks to run.

### DIFF
--- a/tests/test_suites/LongSystemTests/test_build_while_auto_recovery.sh
+++ b/tests/test_suites/LongSystemTests/test_build_while_auto_recovery.sh
@@ -26,9 +26,9 @@ master_kill_loop() {
 ( master_kill_loop & )
 
 cd "${info[mount0]}"
-assert_success git clone https://github.com/saunafs/saunafs.git
+assert_success git clone https://github.com/leil-io/saunafs.git
 saunafs setgoal -r 2 saunafs
 mkdir saunafs/build
 cd saunafs/build
-assert_success cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install
+assert_success cmake .. -G 'Unix Makefiles' -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install
 assert_success make -j${PARALLEL_JOBS} install
--- a/tests/test_suites/LongSystemTests/test_build_while_changing_master.sh
+++ b/tests/test_suites/LongSystemTests/test_build_while_changing_master.sh
@@ -55,9 +55,9 @@ master_kill_loop() {
 master_kill_loop &
 
 cd "${info[mount0]}"
-assert_success git clone https://github.com/saunafs/saunafs.git
+assert_success git clone https://github.com/leil-io/saunafs.git
 saunafs setgoal -r 2 saunafs
 mkdir saunafs/build
 cd saunafs/build
-assert_success cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install
+assert_success cmake .. -G 'Unix Makefiles' -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=../install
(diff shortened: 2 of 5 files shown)


uristdwarf:
I don't really the understand the purpose of these tests. Are they testing building SaunaFS _while_  SaunaFS is running? I don't understand the point unless you're doing it for performance, but there's got to better ways than this?

uristdwarf:
Actually, after reading the commit history, the point here is that it's supposed to test how the filesystem is functioning while a pretty long process is ongoing on that filesystem (in this case, building Saunafs. Clever.). So this now makes a lot more sense.

Because of that, I feel like my previous concern is addressed, and the fixes look good enough. Also, I think I will backport these fixes into release/4.1.0, just so I can get a final test summary regarding the release.

P.S: @Antuan-AE Did you mark it as a draft because of my comment or something else? I could approve right now.

tests/test_suites/LongSystemTests/test_build_while_auto_recovery.sh
```diff
@@ -26,9 +26,9 @@ master_kill_loop() {
 ( master_kill_loop & )
 
 cd "${info[mount0]}"
-assert_success git clone https://github.com/saunafs/saunafs.git
+assert_success git clone https://github.com/leil-io/saunafs.git
 saunafs setgoal -r 2 saunafs
 mkdir saunafs/build
 cd saunafs/build
-assert_success cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install
```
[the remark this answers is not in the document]
aNeutrino: it is true that `Debug` would be te fastest... but this is long test so maybe we are OK to build all 
It is just _interesting_ test a bit like use compiler to compile itself.
I would leave it with `RelWithDebInfo`

tests/test_suites/LongSystemTests/test_build_while_changing_master.sh
```diff
@@ -55,9 +55,9 @@ master_kill_loop() {
 master_kill_loop &
 
 cd "${info[mount0]}"
-assert_success git clone https://github.com/saunafs/saunafs.git
+assert_success git clone https://github.com/leil-io/saunafs.git
 saunafs setgoal -r 2 saunafs
 mkdir saunafs/build
 cd saunafs/build
-assert_success cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install
+assert_success cmake .. -G 'Unix Makefiles' -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=../install
```
aNeutrino, line 62: is it for making sure that under Windows we have still Unix make-files ? 

... what this `-G 'Unix Makefiles'` is for ?

aNeutrino (approved):
Perhaps commit message could mention more details about this patch.

Antuan-AE:
> Actually, after reading the commit history, the point here is that it's supposed to test how the filesystem is functioning while a pretty long process is ongoing on that filesystem (in this case, building Saunafs. Clever.). So this now makes a lot more sense.
> 
> Because of that, I feel like my previous concern is addressed, and the fixes look good enough. Also, I think I will backport these fixes into release/4.1.0, just so I can get a final test summary regarding the release.
> 
> P.S: @Antuan-AE Did you mark it as a draft because of my comment or something else? I could approve right now.

It was moved to draft because of a segFault that was on dev unNoticed and the issue was solved in another PR. It's now ready for review.

lgsilva3087 (approved):
LGTM.

tests/test_suites/LongSystemTests/test_build_while_changing_master.sh
Antuan-AE: The cmake tool has different types of generator (Unix Makefiles, Ninja, Visual Studio, Eclipse, CodeBlocks, kDevelop3, etc). The -G 'Unix Makefiles' instruct cmake to use the Unix/Linux Makefile generator because we are building the code using make command after that line.

PR #57, opened 2024-04-29
tidy(debian): Cleanup control file

uristdwarf:
* Add a missing build dependancy
* Remove Conflicts/Replaces/Provides. These were leftovers from LizardFS that ensured MooseFS and LizardFS could not be installed at the same time (since they shared executable names). We don't need this anymore since all the names have changed, and the sfs was a little confusing due to the rename, which probably delayed people from noticing this.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,7 +14,6 @@ Build-Depends: asciidoc (>= 8.6.6),
                libboost-program-options-dev,
                python3,
                libyaml-cpp-dev,
-
 Rules-Requires-Root: no
 Standards-Version: 3.9.3
 Homepage: https://saunafs.org/
--- a/debian/control
+++ b/debian/control
@@ -13,7 +13,8 @@ Build-Depends: asciidoc (>= 8.6.6),
                libboost-system-dev,
                libboost-program-options-dev,
                python3,
-	       libyaml-cpp-dev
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -13,6 +13,7 @@ Build-Depends: asciidoc (>= 8.6.6),
                libboost-system-dev,
                libboost-program-options-dev,
                python3,
+	       libyaml-cpp-dev
```
aNeutrino, line 16: something looks wrong with _whitespace_ indentation here.

aNeutrino (approved):
LGTM

debian/control
uristdwarf: Seems my editor has mixed spaces and tabs for some reason

debian/control
```diff
@@ -13,6 +13,7 @@ Build-Depends: asciidoc (>= 8.6.6),
                libboost-system-dev,
                libboost-program-options-dev,
                python3,
+               libyaml-cpp-dev,
 Rules-Requires-Root: no
 Standards-Version: 3.9.3
 Homepage: https://saunafs.org/
```
lgsilva3087, line 19: Redirects to leil's web page.

lgsilva3087 (approved):
LGTM.

PR #58, opened 2024-04-29
fix(tests): Upgrade tests

uristdwarf:
This is a series of 2 commits squashed:

fix(tests): Upgrade tests on freshly installed machines

This commit fixes an issue from earlier regarding updates to setup_machine.sh. Now that the SaunaFS key is acquired from the setup_machine.sh script and it adds the repository, there is no need to add a source to the script when doing upgrade tests. Doing so would conflict with the already setup repository

fix(tests): Long upgrade tests

* Fix no old version checking by using latest client instead.
* Fix recursion bug by not using sfs* tools, instead using saunafs

### DIFF
--- a/tests/tools/saunafsXX.sh
+++ b/tests/tools/saunafsXX.sh
@@ -27,7 +27,6 @@ Dir::Cache "${TEMP_DIR}/apt/var/cache/apt";
 Dir::Etc::Parts "${TEMP_DIR}/apt/apt.conf.d";
 END
 		local destdir="${TEMP_DIR}/apt/var/cache/apt/archives"
-		echo "deb [arch=amd64 trusted=yes] https://repo.saunafs.com/repository/saunafs-ubuntu-22.04/ ${codename} main" >${TEMP_DIR}/apt/saunafs.list
 		env APT_CONFIG="${TEMP_DIR}/apt/apt.conf" apt-get update
 		env APT_CONFIG="${TEMP_DIR}/apt/apt.conf" apt-get -y --allow-downgrades install -d \
 			saunafs-master=${SAUNAFSXX_TAG_APT} \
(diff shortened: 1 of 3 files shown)


tests/tools/saunafsXX.sh
```diff
@@ -141,7 +140,9 @@ assert_saunafsXX_services_count_equals() {
 }
 
 assert_no_saunafsXX_services_active() {
-	assert_saunafsXX_services_count_equals 0 0 0
+	assert_equals 0 "$(saunafs_admin_master info | grep "${SAUNAFSXX_TAG}" | wc -l)"
```
aNeutrino, line 143: @uristdwarf why having it here in 3 lines is better then in one line ?

aNeutrino: is it because there was `saunafs_old_admin_master` inside of `assert_saunafsXX_services_count_equals`

can you explain logic here?

why there are two of them?
`saunafs_old_admin_master` & `saunafs_admin_master`

old is before upgrade...also called XX and other is after upgrade ?

aNeutrino: I am bit lost... something is wrong about this XX in name that it is confusing... maybe we should use 'previous' and 'current 'or 'old' and 'new' instead of 'XX' and '' 


however this is separate refactoring... not related with this PR directly...
So I will accept it.

aNeutrino (approved):

tests/tools/saunafsXX.sh
uristdwarf: Yes, XX is the "old" version, or rather the previous version from the current. Why such an awkward name was chosen is a mystery to me.

> is it because there was saunafs_old_admin_master inside of assert_saunafsXX_services_count_equals
can you explain logic here?
why there are two of them?
saunafs_old_admin_master & saunafs_admin_master
old is before upgrade...also called XX and other is after upgrade ?

The problem occurred when all the old services where stopped and replaced with new ones. The metadata parallel dumping feature modified the info command API with master, which means older admin executables no longer work with newer masters/chunkservers etc. The solution here was to stop using the old version and instead use the newer one instead to check it.

This is actually something of a question of mine, should we support using older versions of executables to communicate with newer services? And which executables/services exactly?

uristdwarf: Actually, I believe the newer version cannot communicated either

lgsilva3087 (approved):
LGTM.

PR #59, opened 2024-04-29
tests: Add basic performance benchmark

lgsilva3087:
This commit adds a very basic write/read speed estimation using dd. The idea is to use it to compare if the performance was affected by changes introduced in code.

### DIFF
--- a/tests/test_suites/Benchmarks/test_basic_performance_estimation.sh
+++ b/tests/test_suites/Benchmarks/test_basic_performance_estimation.sh
@@ -1,0 +1,25 @@
+# This test does a very basic estimation for write and read speed of the system.
+# The idea is to use it to compare the performance after some changes.
+
+CHUNKSERVERS=1 \
+	USE_RAMDISK=YES \
+	MOUNT_EXTRA_CONFIG="mfscachemode=NEVER" \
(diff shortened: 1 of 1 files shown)


uristdwarf (approved):

tests/test_suites/Benchmarks/test_basic_performance_estimation.sh
```diff
@@ -0,0 +1,25 @@
...
+CHUNKSERVERS=1 \
+	USE_RAMDISK=YES \
+	MOUNT_EXTRA_CONFIG="mfscachemode=NEVER" \
+	CHUNKSERVER_EXTRA_CONFIG="GARBAGE_COLLECTION_FREQ_MS = 0|HDD_TEST_FREQ = 100000" \
+	AUTO_SHADOW_MASTER="NO" \
+	setup_local_empty_saunafs info
+
+cd "${info[mount0]}"
+
+sizeFittingRamdisk=1536
```
Antuan-AE, line 13: It's there any particular reason for this value? Is it related to the size on the chunks and having an integer amount of chunks being copied? Other than that i think it is ok for me to have this added value. LGTM.

Antuan-AE (approved):

PR #60, opened 2024-04-29
ci, mount: Fix segFault on libsaunafsmount_shared

Antuan-AE:
A typo at CMAKE_BUILD_TYPE made the cmake to build the code in Debug mode by default. This was making a segFault error pass un-detected because linkage modification introduced at PR#19.


lgsilva3087 (approved):
LGTM

PR #62, opened 2024-05-01
chunkserver: Add option to disable CRC at writing

lgsilva3087:
The chunkserver check the CRC at every write, so the CRC calculation is the function taking most of the CPU when writing. This commit adds a configuration option to disable it. By default, the CRC is always checked, which is the safe way. Disable it under your responsibility.


rolysr (approved):
LGTM.

PR #64, opened 2024-05-03
mount: Fix Ganesha multi_export test

dmga44:
Currently the test_nfs_ganesha_multi_export is failing. The solution was
to keep the write data cache inodedata maxfleng a while longer and
update it also during truncates: test was failing due to master not
having the correct size while the client already dropped their cached
size (the inodedata's maxfleng). Thus, keeping it some extra time should
suffice.

For now, it seems that it fixes the test_nfs_ganesha_multi_export, and passes the SanityChecks, Ganesha test suite and ShortSystemTests.

### DIFF
--- a/src/mount/writedata.cc
+++ b/src/mount/writedata.cc
@@ -367,7 +367,9 @@ void write_job_delayed_end(inodedata* id, int status, int seconds, Glock &lock) 
 		write_cb_release_blocks(id->dataChain.size(), lock);
 		id->dataChain.clear();
 		id->inqueue = false;
-		id->maxfleng = 0; // proper file length is now on the master server, remove our length cache
+		// We don't reset maxfleng (id->maxfleng = 0;) for a while longer, to
+		// have its value ready to some quick cache responses in lookup and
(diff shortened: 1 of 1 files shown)


lgsilva3087 (approved):

src/mount/writedata.cc
```diff
@@ -367,7 +367,7 @@ void write_job_delayed_end(inodedata* id, int status, int seconds, Glock &lock)
 		write_cb_release_blocks(id->dataChain.size(), lock);
 		id->dataChain.clear();
 		id->inqueue = false;
-		id->maxfleng = 0; // proper file length is now on the master server, remove our length cache
+		// We don't reset maxfleng for a while longer
```
Antuan-AE, line 370: Is this comment correct here or should it be placed below inside the if statement?

Antuan-AE (approved):
LGTM

src/mount/writedata.cc
lgsilva3087: Done.

PR #69, opened 2024-05-07
all: Apply fix to build using gcc-13 on release

Antuan-AE:
When using cmake flag RelWithDebInfo using gcc version 13 two issues arise:

- common: the small_vector template raises the Werror=stringop-overflow so compiler is instructed to ignore this warning until a proper fix can be applied.
- uraft: boost::asio std::exchange is not found.

### DIFF
--- a/src/common/small_vector.h
+++ b/src/common/small_vector.h
@@ -22,7 +22,10 @@
 
 #include <algorithm>
 #include <array>
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
 #include <boost/container/small_vector.hpp>
+#pragma GCC diagnostic pop
 #include <cassert>
 #include <cstddef>
 #include <memory>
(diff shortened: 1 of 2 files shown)


ralcolea (approved):
LGTM. Before merging, please check typos in the commit message:
- **arise** instead of **arrise**.
- **issues** instead of **issue**.

Antuan-AE:
> LGTM. Before merging, please check typos in the commit message:
> 
> * **arise** instead of **arrise**.
> * **issues** instead of **issue**.

Fixed.

lgsilva3087 (approved):
LGTM.

uristdwarf:
Probably best to merge these changes to release as well

uristdwarf:
> Probably best to merge these changes to release as well

Actually, nevermind. The small vector is not changed in release, so there's no need (and it conflicts anyway). However I will check if there's still a build error just in case

PR #71, opened 2024-05-08
chore(INSTALL.md): Include instructions for Ubuntu 22.04

uristdwarf:
Follow-up to commit 806603a, this adds instructions for 22.04 and fixes a mistake with the codename still being jammy.


lgsilva3087 (approved):
LGTM.

PR #73, opened 2024-05-08
mount: Fix gid in oplog output

lgsilva3087:
The gid is encoded in some operations to indicate that the secondary groups were also sent. This is done by setting the most significative bit using this constant:

constexpr std::uint32_t kSecondaryGroupsBit = (std::uint32_t)1 << 31;

The oplog was printing the gid directly, without checking, showing 2147483650 (1 << 31) as gid. This commit checks the flag and prints the correct primary group.


aNeutrino (commented):
LGTM

aNeutrino (approved):

PR #75, opened 2024-05-09
chore: Update NEWS and debian/changelog

uristdwarf:
* Add tool for updating both, so you need to only write it once.
* Fix date in previous NEWS entry

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,23 @@
 This file lists noteworthy changes in SaunaFS.
 
- * SaunaFS (4.0.1) (2024-02-28)
+ * SaunaFS (4.1.0) (2024-05-09)
+ - (master) Improve master startup time by allowing parallel metadata loading
+ - (mount) Add `sfsdirectio` mount option
+   Previously you had to edit .saunafs_tweaks file.
+ - (mount) Fix undefined behavior in client
+   Could cause a deadlock when reconnecting after a session was lost.
+ - (mount) Fix wrong GID being shown in .oplog
+ - (mount) Improve DirEntryCache behaviour by caching getattr/lookup operations
+ - (admin) Add `--help option` in saunafs-admin
+ - (admin) Add `--defaults` option for `dump-config` in saunafs-admin
+ - (cgi) Update saunafs-cgi.
+ - (all) Add support for Ubuntu 24.04
+ - (all) Fix compilation issues on GCC 13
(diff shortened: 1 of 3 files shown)


utils/update_news.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "Enter version (e.g., 4.0.2):"
+read version
+echo "Enter urgency (low, medium, high):"
+read urgency
+
+# Current date for changelog
+current_date=$(get_current_date)
+
+# Prepare temporary file for editing
+TMP_FILE="/tmp/changelog_$version"
```
Antuan-AE, line 22: Consider adding brackets around variables like: 
"${version}"

Antuan-AE (approved):
Please check comments regarding variables without brackets at update_news.sh. Aside from that it LGTM.

NEWS
```diff
@@ -1,6 +1,22 @@
...
+   Could cause a deadlock when reconnecting after a session was lost.
+ - (mount) Fix wrong GID being shown in .oplog
+ - (admin) Add `--help option` in saunafs-admin
+ - (admin) Add `--defaults` option for `dump-config` in saunafs-admin
+ - (cgi) Update saunafs-cgi.
+ - (all) Add support for Ubuntu 24.04
+ - (all) Fix compilation issues on GCC 13
+ - (all) Various bug fixes, especially regarding tests
+ - (windows) Support compiling common and client library
+ - (windows) Improve behavior for Windows client
```
lgsilva3087, line 17: Consider adding:
- (mount) Improve DirEntryCache behaviour by caching getattr and lookup operations

lgsilva3087 (approved):
LGTM. Please see my suggestion.

utils/update_news.sh
```diff
@@ -0,0 +1,73 @@
...
+    # Check if line starts with '(' indicating a new item
+    if [[ $line == \(* ]]; then
+        news_entry+=" - $line\n"
+        changelog_entry+="  * $line\n"
+        previous_line_new_item=true
+    else
+        news_entry+="   $line\n"
+        changelog_entry+="   $line\n"
+    fi
+done < "$TMP_FILE"
```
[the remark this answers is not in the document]
uristdwarf: I couldn't get it to work :(

But we can always improve it later :)

I just quickly wrote this script so I or anyone else doesn't have to do it twice in two files.

PR #77, opened 2024-05-13
mount: Fix test_sparse_file

dmga44:
Recent runs of this test under Windows failed apparently due to
different reasons. Sometimes during a setattr syscall and sometimes
during a write syscall. Debug revealed weird values of the maxfleng
field when it was supposed to be zero.

The cause of this issue was that the inodedata (```id```) could be
released in the ```write_data_lcnt_decrease``` function and the later
update of the maxfleng field was pointing to already freed memory.
Solution consisted in checking whether the inodedata was released and
only perform the update of its maxfleng if it was not deleted.

Note that the fix is not bounded only to Windows client code. It was only noticed under Windows but the issue is also present in the Linux clients.

### DIFF
--- a/src/mount/writedata.cc
+++ b/src/mount/writedata.cc
@@ -852,11 +852,20 @@ static void write_data_lcnt_increase(inodedata *id, Glock&) {
 	id->lcnt++;
 }
 
-static void write_data_lcnt_decrease(inodedata *id, Glock& lock) {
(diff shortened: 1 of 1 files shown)


Antuan-AE (approved):

lgsilva3087 (approved):
LGTM.

PR #83, opened 2024-05-15
chunkserver: Add thread names

lgsilva3087:
This change renames the chunkserver threads to better identify them at runtime. It is useful for debugging and profiling purposes.

Note: thread names can be up to 16 characters.


Antuan-AE (approved):
LGTM

ralcolea (approved):
LGTM

PR #84, opened 2024-05-15
chunkserver: Fix race condition in BuffersPool

lgsilva3087:
The check for buffers_.empty() was not protected, causing segmentation fault when multiple NetworkWorkers were configured.

The queue max size (kMaxSize) was also increased by 8, to better support heavy workloads.


ralcolea (approved):
LGTM.

uristdwarf (approved):

PR #88, opened 2024-05-16
Fix mount when master down

rolysr:
When master server disconnected, windows client
was unable to be used and the process terminated.
With this change is its possible to the windows
client to know when master is disconnected and
make the mountpoint still accessible.

### DIFF
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -80,10 +80,9 @@ typedef struct _acquired_file {
 	struct _acquired_file *next;
 } acquired_file;
 
-
 #define DEFAULT_OUTPUT_BUFFSIZE 0x1000
 #define DEFAULT_INPUT_BUFFSIZE 0x10000
-
+#define NO_DATA_RECEIVED_FROM_MASTER NULL
 #define RECEIVE_TIMEOUT 10
 
 static threc *threchead=NULL;
@@ -205,6 +204,9 @@ void master_stats_add(uint8_t id,uint64_t s) {
 static inline void setDisconnect(bool value) {
 	std::unique_lock<std::mutex> fdLock(fdMutex);
 	disconnect = value;
+#ifdef _WIN32
+	gIsDisconnectedFromMaster.store(value);
+#endif
 	if(value) {
 		SaunaClient::masterDisconnectedCallback();
 		safs_pretty_syslog(LOG_WARNING,"master: disconnected");
@@ -374,6 +376,14 @@ static bool fs_threc_send_receive(threc *rec, bool filter, PacketHeader::Type ex
 const uint8_t* fs_sendandreceive(threc *rec, uint32_t expected_cmd, uint32_t *answer_leng) {
 	// this function is only for compatibility with Legacy code
 	sassert(expected_cmd <= PacketHeader::kMaxOldPacketType);
+
+#ifdef _WIN32
+	if (gIsDisconnectedFromMaster.load()) {
+		*answer_leng = 0;
+		return NO_DATA_RECEIVED_FROM_MASTER;
+	}
+#endif
+
 	if (fs_threc_send_receive(rec, true, expected_cmd)) {
 		const uint8_t *answer;
 		answer = rec->inputBuffer.data();
@@ -385,7 +395,7 @@ const uint8_t* fs_sendandreceive(threc *rec, uint32_t expected_cmd, uint32_t *an
 
 		return answer;
 	}
-	return NULL;
+	return NO_DATA_RECEIVED_FROM_MASTER;
 }
 
 bool fs_sausendandreceive(threc *rec, uint32_t expectedCommand, MessageBuffer& messageData) {
@@ -1198,6 +1208,9 @@ void* fs_receive_thread(void *) {
 			continue;
 		} else {
 			// connecection succeeded -- reset timeout the initial value
+#ifdef _WIN32
+			gIsDisconnectedFromMaster.store(false);
+#endif
 			reconnectSleep_ms = initialReconnectSleep_ms;
 		}
 		fdLock.unlock();
--- a/src/mount/mastercomm.h
+++ b/src/mount/mastercomm.h
@@ -36,6 +36,10 @@
 #include "protocol/directory_entry.h"
 #include "protocol/named_inode_entry.h"
 
+#ifdef _WIN32
+inline std::atomic_bool gIsDisconnectedFromMaster;
+#endif
+
 void fs_getmasterlocation(uint8_t loc[14]);
 uint32_t fs_getsrcip(void);
 
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -230,6 +230,8 @@ uint8_t session_flags;
 
 uint8_t get_session_flags() { return session_flags; }
 
+bool isMasterDisconnected() { return gIsDisconnectedFromMaster.load(); }
+
 #define LOCAL_USERS_THRESHOLD 0x30000
 static int32_t mounting_uid = USE_LOCAL_ID;
 static int32_t mounting_gid = USE_LOCAL_ID;
--- a/src/mount/sauna_client.h
+++ b/src/mount/sauna_client.h
@@ -334,6 +334,8 @@ struct RequestException : public std::exception {
 
 uint8_t get_session_flags();
 
+bool isMasterDisconnected();
+
 void update_last_winfsp_context(const unsigned int uid, const unsigned int gid);
 
 void convert_winfsp_context_to_master_context(unsigned int& uid, unsigned int& gid);


src/mount/mastercomm.cc
```diff
@@ -205,6 +205,7 @@ void master_stats_add(uint8_t id,uint64_t s) {
 static inline void setDisconnect(bool value) {
 	std::unique_lock<std::mutex> fdLock(fdMutex);
 	disconnect = value;
+	masterServerIsDisconnected = value;
```
lgsilva3087, line 208: Please consider renaming this variable to isDisconnectedFromMaster or isConnectedToMaster, which is usually recommended for boolean variables.

src/mount/mastercomm.cc
```diff
@@ -374,6 +375,21 @@ static bool fs_threc_send_receive(threc *rec, bool filter, PacketHeader::Type ex
 const uint8_t* fs_sendandreceive(threc *rec, uint32_t expected_cmd, uint32_t *answer_leng) {
 	// this function is only for compatibility with Legacy code
 	sassert(expected_cmd <= PacketHeader::kMaxOldPacketType);
+
+#ifdef _WIN32
+    // Code specific to Windows
+    if (masterServerIsDisconnected) {
+        const uint8_t defaultAnswer[] = {
```
lgsilva3087, line 382: I agree with Baldor, the numbers should be explained in some way.
Also: the variable should be static to avoid initialization every time or even a global constexpr.

src/mount/mastercomm.cc
```diff
@@ -374,6 +375,21 @@ static bool fs_threc_send_receive(threc *rec, bool filter, PacketHeader::Type ex
 const uint8_t* fs_sendandreceive(threc *rec, uint32_t expected_cmd, uint32_t *answer_leng) {
 	// this function is only for compatibility with Legacy code
 	sassert(expected_cmd <= PacketHeader::kMaxOldPacketType);
+
+#ifdef _WIN32
+    // Code specific to Windows
```
lgsilva3087, line 380: From the #ifdef can be inferred that the code is specific for Windows, probably the comment should be related to why it is only required for Windows.

src/mount/mastercomm.h
```diff
@@ -36,6 +36,8 @@
 #include "protocol/directory_entry.h"
 #include "protocol/named_inode_entry.h"
 
+inline bool masterServerIsDisconnected;
```
lgsilva3087, line 39: Is this global variable accessed by only one thread? If it is accesed by multiple threads, then we need to protect it, probbaly using an atomic or mutex. Please check.

src/mount/mastercomm.cc
```diff
@@ -1389,6 +1406,13 @@ void fs_statfs(uint64_t *totalspace,uint64_t *availspace,uint64_t *trashspace,ui
 }
 
 uint8_t fs_access(uint32_t inode,uint32_t uid,uint32_t gid,uint8_t modemask) {
+	#ifdef _WIN32
+	// Code specific to Windows
+	if (masterServerIsDisconnected) {
+		return SAUNAFS_STATUS_OK;
+	}
```
lgsilva3087, line 1413: I don't know if it is correct to return OK if the master is disconnected. Please verify.

lgsilva3087 (commented):
Please take a look at my comments.
I have a doubt about the commit message, at the end it says:
```
and make the mountpoint still accessible
```
My doubt is: How can it be still accessible if it is disconnected from master? Or the idea is that now it does not break?

src/mount/mastercomm.cc
```diff
@@ -374,6 +375,21 @@ static bool fs_threc_send_receive(threc *rec, bool filter, PacketHeader::Type ex
...
 	sassert(expected_cmd <= PacketHeader::kMaxOldPacketType);
+
+#ifdef _WIN32
+    // Code specific to Windows
+    if (masterServerIsDisconnected) {
+        const uint8_t defaultAnswer[] = {
+            0x00, 0x00, 0x00, 0x02, 0xD4, 0xD4, 0x70, 0x00, 0x00, 0x00,
+            0x00, 0x02, 0xD4, 0xD4, 0x70, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x06, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x0E, 0x00, 0x00, 0x00, 0x01
```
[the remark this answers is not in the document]
rolysr: Those numbers represents the expected values for an `struct stat` variable that contains information about an specific inode attributes like permission, last change time, last access time and etc. That data is received from a master server and then parsed to an `struct stat`. For simplicity I put that value con a separate global variable and setted every value to 0, so the permissions and other attributes have default/zero values.

src/mount/mastercomm.cc
rolysr: Done

src/mount/mastercomm.cc
rolysr: Done

src/mount/mastercomm.h
rolysr: Yes, it is accessed by only one thread. Anyway I changed the variable to an atomic bool.

src/mount/mastercomm.cc
rolysr: I checked the behavior and it is the same without that line so I decided to delete it

ralcolea:
@rolysr looks like this feature has compilation issues on Jenkins. Could you check what is wrong?

src/mount/sauna_client.h
```diff
@@ -332,6 +332,11 @@ struct RequestException : public std::exception {
 
 uint8_t get_session_flags();
 
+inline std::atomic<bool> isDisconnectedFromMaster;
```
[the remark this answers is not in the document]
rolysr: Done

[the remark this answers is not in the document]
rolysr: Done

src/mount/sauna_client.h
```diff
@@ -332,6 +332,11 @@ struct RequestException : public std::exception {
...
+
+bool isMasterDisconnected();
+void setIsMasterDisconnected(bool value);
```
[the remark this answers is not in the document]
rolysr: Considered

src/mount/sauna_client.h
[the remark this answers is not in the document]
rolysr: As it will be used on result from https://github.com/leil-io/windows-client/pull/7 where the clients has a new thread that request to that value at the same time it could be change during the client's main process.

src/mount/sauna_client.cc
```diff
@@ -229,6 +229,12 @@ uint8_t session_flags;
 
 uint8_t get_session_flags() { return session_flags; }
 
+bool isMasterDisconnected() { return isDisconnectedFromMaster.load(); }
+
+void setIsMasterDisconnected(bool value) {
+	isDisconnectedFromMaster.store(value);
+}
+
```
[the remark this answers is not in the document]
rolysr: Considered

src/mount/mastercomm.cc
```diff
@@ -374,6 +386,19 @@ static bool fs_threc_send_receive(threc *rec, bool filter, PacketHeader::Type ex
...
+
+#ifdef _WIN32
+	// If the master server disconnects, the system returns a standard
+	// representation for a status object that reflects the requested attributes
+	// for a particular inode. This action aims to mimic the expected outcome
+	// had the master server been connected, thus preventing errors due to
+	// missing values.
+	if (SaunaClient::isMasterDisconnected()) {
+		const uint8_t *defaultAnswerPtr = defaultFsSendAndReceiveResponse;
+		return defaultAnswerPtr;
```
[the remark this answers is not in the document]
rolysr: New implementation simplifies this approach by returning a NO_DATA_RECEIVED_FROM_MASTER values that is equivalent to NULL and represents the fact of no data received from master.

src/mount/mastercomm.cc
```diff
@@ -80,6 +80,15 @@ typedef struct _acquired_file {
...
 } acquired_file;
 
+#ifdef _WIN32
+// These default values enable setting all requested inodes to "no permission,"
+// thereby preventing interaction with them by all users, allowing the client to
+// operate without interruption.
+constexpr uint8_t defaultFsSendAndReceiveResponse[] = {
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
```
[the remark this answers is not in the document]
rolysr: Delete this constant as not needed for new approach returning null in fs_sendandreceive function

src/mount/mastercomm.cc
```diff
@@ -374,6 +386,19 @@ static bool fs_threc_send_receive(threc *rec, bool filter, PacketHeader::Type ex
 const uint8_t* fs_sendandreceive(threc *rec, uint32_t expected_cmd, uint32_t *answer_leng) {
```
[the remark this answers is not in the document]
rolysr: Done

src/mount/mastercomm.cc
```diff
@@ -80,6 +82,15 @@ typedef struct _acquired_file {
...
 } acquired_file;
 
+#ifdef _WIN32
+// These default values enable setting all requested inodes to "no permission,"
+// thereby preventing interaction with them by all users, allowing the client to
+// operate without interruption.
+constexpr uint8_t defaultFsSendAndReceiveResponse[] = {
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
```
uristdwarf, line 92: Where is this variable used?

uristdwarf (commented):
I guess my main question is why it wouldn't work on Windows? Ideally, it would be better to have a solution that is cross-platform, rather than ifdefs all over the place, but it's not a deal breaker.

src/mount/mastercomm.cc
rolysr: After new considerations there was a simplification of the implementation as it is not necessary with the latest approach

rolysr:
> I guess my main question is why it wouldn't work on Windows? Ideally, it would be better to have a solution that is cross-platform, rather than ifdefs all over the place, but it's not a deal breaker.

The goal was to keep this as just implementation for Windows, as the added variable and procedures for detecting master disconnection avoids errors on Windows, this logic is not intended to be used on Linux systems. Anyway is ok to me to consider avoiding ifdef as much as possible to better cross-platform development.

dmga44 (approved):
LGTM.

lgsilva3087 (approved):
LGTM.

uristdwarf:
> The goal was to keep this as just implementation for Windows, as the added variable and procedures for detecting master disconnection avoids errors on Windows, this logic is not intended to be used on Linux systems.

My thought is that if we are going to make the client cross-platform anyway, we may as well go the extra mile and clean up the code so that it could work regardless of platform, otherwise the ifdefs will pile up as we add more hacks to make it work for different platforms. What I'm saying is basically don't be scared to change the code significantly you have the time to make it more cross-platform :). But depending on the scope, it might be a lot of work.

uristdwarf (approved):

Issue #68, opened 2024-05-07
NFS Ganesha, FSAL, file corruption on master fail-over


LukasK32:
Behavior of NFS Ganesha with SaunaFS FSAL on SaunaFS Master fail-over is flaky and sometimes files in-flight are corrupted.

## Software versions
OS: `Ubuntu 22.04LTS`
SaunaFS Version: `4.0.1-20240301-164017-stable-main-82161d4f` (from SaunaFS repositories)
Ganesha Version: `4.3-1` (from lunar repositories)

## Setup
For easy reproduction, steps are performed on a single machine with SaunaFS Master, Chunkserver and NFS Ganesha. Master failover are simulated by restarting the service.

Configuration files:

* `/etc/saunafs/sfsmaster.cfg`

    ```cfg
    PERSONALITY = master
    MASTER_HOST = sfsmaster
    MASTER_PORT = 9419
    MASTER_TIMEOUT = 10
    AUTO_RECOVERY = 1
    REJECT_OLD_CLIENTS = 1
    AVOID_SAME_IP_CHUNKSERVERS = 1
    MATOCS_LISTEN_HOST = *
    MATOCS_LISTEN_PORT = 9420
    MATOCL_LISTEN_HOST = *
    MATOCL_LISTEN_PORT = 9421
    OPERATIONS_DELAY_DISCONNECT = 300
    ```

* `/etc/saunafs/sfsexports.cfg`

    ```cfg
    *                       /       rw,alldirs,maproot=0,ignoregid
    *                       .       rw
    ```

* `/etc/saunafs/sfsgoals.cfg`

    ```cfg
    1 1 : _
    2 2 : _ _
    3 3 : _ _ _
    4 4 : _ _ _ _
    5 5 : _ _ _ _ _
    10 ec21 : $ec(2,1)
    11 ec31 : $ec(3,1)
    12 ec32 : $ec(3,2)
    ```

* `/etc/saunafs/sfschunkserver.cfg`

    ```cfg
    MASTER_HOST = 10.20.32.159
    HDD_LEAVE_SPACE_DEFAULT = 2GB
    ```

* `/etc/saunafs/sfshdd.cfg`

    ```cfg
    /var/lib/saunafs-chunks
    ```

* `/etc/ganesha/ganesha.conf`

    ```conf
    EXPORT
    {
        Export_Id = 77;
        Path = "/";
        Pseudo = "/";
        Access_Type = RW;
        Squash = None;
[...]

PR #91, opened 2024-05-21
test: Add Ganesha test for validating file corruption

ralcolea:
This commit adds an integration test to detect whether there is file corruption during master fail over. This test is based on issue #68 of SaunaFS reported by Sarkans team.

Basically, the test copies a 3Gb file through NFS mount and restarts the master server during the process. After copying the file, checksum of the file is verified twice: first, reading the file through NFS mount and second, reading the file through SaunaFS mount.


Antuan-AE (approved):
LGTM

PR #92, opened 2024-05-21
test: Enable nfs v4.1 in Ganesha multi export test

ralcolea:
At some point, nfs client started failing when connecting using nfs v4.1 and it was disabled in commit dfd282f.

After updating to Ubuntu 24.04, these issues with nfs client v4.1 were solved. This commit revert the changes of commit dfd282f and the test is now passing.


ralcolea:
I noticed the test failed precisely because of the reverted change. I tested in Ubuntu 24.04 and the test is passing. What OS is installed in the node where GaneshaTests is being run? By any change is it Ubuntu 22.04? If it's the case, then we cannot merge this change :)

ralcolea:
The following pictures shows logs for ganesha-enable-nfsv4.1 branch and PR-92 that are created for running Ganesha tests:
logs-branch-PR-92
logs-branch-ganesha-enable-nfsv4 1

Previous logs show that for some reason, the node where the original branch is running supports nfsv4.2. However, the node where the PR-92 code is run, doesn't support neither nfsv4.2 nor nfsv4.1, even if both are running in Ubuntu 24.04 according to the initial part of the log (both uses **registry.ci.leil.io/saunafs-test:dev-ubuntu-24.04-latest**).

I'm not sure how to proceed in this case.

Antuan-AE (approved):
LGTM

tests/test_suites/GaneshaTests/test_nfs_ganesha_multi_export.sh
```diff
@@ -113,7 +113,7 @@ sudo /usr/bin/ganesha.nfsd -f ${info[mount0]}/ganesha.conf
 assert_eventually 'showmount -e localhost'
 
 for x in 1 2 99; do
-	sudo mount localhost:/e$x $TEMP_DIR/mnt/nfs$x
+	sudo mount -o v4.1 localhost:/e$x $TEMP_DIR/mnt/nfs$x
```
aNeutrino, line 116:
```suggestion
	sudo mount -o v4.1 localhost:/e${x} ${TEMP_DIR}/mnt/nfs${x}
```

aNeutrino (approved):
LGTM

ralcolea:
> I'm a bit worried about 22.04, since we still build and ship packages for it and I'm not sure how the CI handles that. Perhaps add an additional check if it's Ubuntu 24.04 on whether to use v4.1 or not.

You're right @uristdwarf. In the worst case, this change could break packages generation in Ubuntu 22.04. I will check if the test passes in Ubuntu 22.04 and then we can decide.

ralcolea:
After checking the failing test with @ictus4u, we noticed different Linux kernel versions in both nodes: specifically kernel 5.15.0-94-generic was used in the node of PR-Merge and failed because it didn't support nfsv4.1. However, in the node that was passing successfully was used kernel version 5.15.0-107-generic. After updating kernel version for all nodes to **5.15.0-107-generic**, everything worked well.

Summarizing, failing support of recent versions of nfs protocol is not related with Ubuntu's version, but with kernel's one. For older kernels like 5.15.0-94-generic, it is not available nfs versions 4.1 and 4.2. So, it's safe to apply the change and continue building packages as usual.

uristdwarf (approved):

tests/test_suites/GaneshaTests/test_nfs_ganesha_multi_export.sh
ralcolea: Applied your suggestion.

PR #93, opened 2024-05-21
misc: Remove BOM from one file

lgsilva3087:


ralcolea (approved):
LGTM. Nice catch :fire: :fire: :fire:

PR #96, opened 2024-05-24
tidy(common): Split off error code

uristdwarf:
This is the first part of splitting off some common features from the common library that are used by some, but not all libraries, both new and future. The reason is to better encapsulate functionality and avoid rebuilding the same libraries again.


src/mount/client/CMakeLists.txt
```diff
@@ -2,15 +2,15 @@ collect_sources(CLIENT)
...
 
 shared_add_library(saunafs-client-cpp client.cc client_error_code.cc)
 shared_target_link_libraries(saunafs-client-cpp mount)
 
-shared_target_link_libraries(saunafs-client ${CMAKE_DL_LIBS})
 
 add_library(saunafs-client_shared SHARED client.cc saunafs_c_api.cc client_error_code.cc)
-set_target_properties(saunafs-client_shared PROPERTIES OUTPUT_NAME "saunafs-client")
 target_link_libraries(saunafs-client_shared ${CMAKE_DL_LIBS} mount_pic)
+set_target_properties(saunafs-client_shared PROPERTIES OUTPUT_NAME "saunafs-client")
```
uristdwarf, line 14: These are not intended, will fix them

uristdwarf:
@ralcolea Could there be problems with Ganesha with these changes?

ralcolea:
@uristdwarf I have not reviewed yet this PR. However, I don't think is correct to apply commits related with Ganesha in your branch, specifically because these are changes not needed in your branch. These changes, should be applied to dev, not in your branch, because these changes do not belong to your change. Actually, I tested locally the idea of your last commit and failed with the same error.

I would like to check with @lgsilva3087, but I think the solution is to create this **sfserr_pic** library which is PIC, exactly like sauna_client does. Otherwise, I'm not sure if there is some other idea, but definitely the solution is to link Ganesha with this new library you created in the right way.

ralcolea:
> These are temporary and will be removed once I get Ganesha passing in CI (which I did it seems).

As you can see in this CI [run](https://jenkins.ci.leil.io/blue/organizations/jenkins/saunafs/detail/PR-96/7/pipeline/), the last changes do not solve the issue and Ganesha still fails. For some reason, the line you added for linking compiles well, but looks like it doesn't link correctly because I get the same error locally after running Ganesha tests.

> Not sure why PIC is needed, for the shared libraries I think we should statically link and make the shared library PIC. But maybe this doesn't work?

I mentioned adding PIC as a possible solution, I could be wrong. That's why I would like to check with @lgsilva3087 to see if he has other idea.

uristdwarf:
> > These are temporary and will be removed once I get Ganesha passing in CI (which I did it seems).
> 
> As you can see in this CI [run](https://jenkins.ci.leil.io/blue/organizations/jenkins/saunafs/detail/PR-96/7/pipeline/), the last changes do not solve the issue and Ganesha still fails. For some reason, the line you added for linking compiles well, but looks like it doesn't link correctly because I get the same error locally after running Ganesha tests.

I thought it showed green for a moment. Guess I was fooled.

I'll put it to draft until I figure out the problem

uristdwarf:
@ralcolea I think you were right. `sfscommon` at least needs PIC libraries for Ganesha and shared libraries

src/common/massert.h
```diff
@@ -25,8 +25,7 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
-
-#include "common/sfserr.h"
+#include <sys/syslog.h>
```
ralcolea, line 28: Unneeded change.

src/common/memory_mapped_file.cc
```diff
@@ -20,6 +20,7 @@
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <sys/syslog.h>
```
ralcolea, line 23: Unneeded change.

src/nfs-ganesha/handle.c
```diff
@@ -1,4 +1,4 @@
-﻿/*
```
ralcolea, line 1: What was the change in the first line?

ralcolea (approved):
LGTM. Good job @uristdwarf. 
Please, check my minor suggestions and remember to squash unneeded commits before merging to dev.

src/common/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ endif()
 
 shared_add_library(sfscommon ${COMMON_SOURCES})
 shared_target_link_libraries(sfscommon ${YAML_CPP_LIBRARIES})
+target_link_libraries(sfscommon sfserr)
+target_link_libraries(sfscommon_pic sfserr_pic)
```
lgsilva3087, line 23: I am not sure if this will work for all scenarios. The PIC targets are not always available, they are if the client-library (dynamic) is selected. I also don't see a direct usage of a _pic target like this in other parts of the project.
Please search in all the cmake files by: 'target_link_libraries(' to confirm it.

src/common/pcqueue.cc
```diff
@@ -28,6 +28,7 @@
 
 #include "common/massert.h"
 #include "devtools/TracePrinter.h"
+#include "sfserr/sfserr.h"
```
lgsilva3087, line 31: This will work because the file is in the same directory than the sfserr folder. But in other places the include was like: "common/sfserr/sfserr.h". Consider changing it to maintain consistency (minor suggestion).

src/common/sfserr/saunafs_error_codes.cc
```diff
@@ -19,8 +19,6 @@
    along with SaunaFS  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "common/platform.h"
-
```
lgsilva3087, line 23: The common/platform.h contains parts related to the error codes, like this:
```cpp
/* Workaround for Debian/kFreeBSD which does not define ENODATA */
#if defined(__FreeBSD_kernel__) || defined(__FreeBSD__)
#ifndef ENODATA
#define ENODATA ENOATTR
#endif
#endif
```
Please check the implications of not including the header.

src/common/sfserr/sfserr.cc
```diff
@@ -18,22 +18,15 @@
    along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "common/platform.h"
```
lgsilva3087, line 21: The same here.

lgsilva3087 (changes requested):
Please check/consider:
- The change breaks compilation of the SMR plugin and potentialy the windows client. So, a related change should be created on those repositories.
- There are conflicts.
- Personal opinion: I prefer the library in src/sfserror and not inside common. I like to release common from not related responsibilities, like the slogger in the other PR. But it is just a personal preference.

uristdwarf:
>  * Personal opinion: I prefer the library in src/sfserror and not inside common. I like to release common from not related responsibilities, like the slogger in the other PR. But it is just a personal preference.

I do kinda agree, but it's more aesthetics/semantics than anything else. Technically it is an independent library, it's just hosted under the common folder.

Mhm, now that I think about it, `sfserr` seems like a bad name, maybe simply just `errors`?

src/common/massert.h
uristdwarf: This is definitely needed, because the macros below require it. I'm guessing it's showing up as unused because of the fact they are per-processor macros.

uristdwarf: The other alternative is to include them where the macros are used, but I think it's easiest this way to already include the requirements with this header file.

src/nfs-ganesha/handle.c
uristdwarf: Probably need to check all the files and remove all these BOM's, but that's a separate PR.

CMakeLists.txt
```diff
@@ -379,6 +379,7 @@ if(NOT MINGW)
   add_subdirectory(src/metalogger)
   add_subdirectory(src/metarestore)
   add_subdirectory(src/tools)
+  add_subdirectory(src/common/sfserr)
```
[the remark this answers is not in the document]
uristdwarf: What about `errors`? I think it's quite clear that the errors are related to SaunaFS. It may be a different matter if it's a shared library (though I think it would be fine if the library was located under some directory labeled SaunaFS, like `/lib/saunafs`) but thankfully it's not.

src/common/CMakeLists.txt
uristdwarf: > But I think it works both with shared libraries enabled and disabled.

Nope, you were right. I missed this because enabling tests enables shared libraries.

uristdwarf: This fixed the issue for both shared and non-shared builds:

`shared_target_link_libraries(sfscommon ${YAML_CPP_LIBRARIES} STATIC sfserr SHARED sfserr_pic)`

However, I expected shared_target_link_libraries to work. I would have expected it to add sfserr and sfserr_pic to the respective libraries sfscommon and sfscommon_pic. I think this might be because `src/common/` is added before `src/common/sfserr/`, thus causing it not to find the library in the shared_target_link_libraries and not adding it in the PIC library.

In any case, I think this is more explicit about it and works regardless of ordering of add_subdirectory.

src/common/sfserr/saunafs_error_codes.cc
uristdwarf: I'll be moderately surprised if it actually builds on FreeBSD even without this change, but I'll check it out.

uristdwarf: I think the build is generally broken on FreeBSD, I couldn't get main to compile at all, so this is the least of our worries. As for the error code itself which is indeed undefined on FreeBSD, I think the easiest solution is just to copy this portion to the library header, so only the libraries that actually need this will have it defined.

`common/platform.h` seems to be full of hacks though, It needs documenting on what is actually needed and why, and removing things which are not.

uristdwarf: e.g to_string is implemented there for platforms that are missing it and was committed 10 years ago, but are there any platforms still missing it and if so, what platforms?

lgsilva3087: The selected code was just an example. The main idea is that the common/platform.h file has a purpose, and that's why it is included first in every file, it also includes config.h. If you are planning to change it, then this is not the commit for it.

uristdwarf: Ok, I will revert it, but I was hoping to discuss it with these changes when we are splitting the common library up. I don't like the fact that platform.h is modifying the std library, including headers and doing a bunch of hacks which may not be needed by every library. At the very least, new libraries should not include it unless you have a good reason to.

uristdwarf:
I've added a PR for zonefs plugin fixes, but I will need to work on Windows a little bit before I can either confirm it's working or not

uristdwarf:
@dmga44 Said he would help deal with the issues on Windows once this is merged, for now the library should hopefully compile on Windows at least

src/errors/saunafs_error_codes.h
```diff
@@ -19,8 +19,7 @@
    along with SaunaFS  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef __SAUNAFS_ERROR_CODES_H
-#define __SAUNAFS_ERROR_CODES_H
+#pragma once
```
aNeutrino, line 22: 👍🏾

aNeutrino (approved):
LGTM

src/errors/sfserr.cc
```diff
@@ -19,20 +19,16 @@
...
-
-#include <array>
 #include <mutex>
-#include <string>
 #include <unordered_map>
 
-#include "common/errno_defs.h"
+#include "saunafs_error_codes.h"
+
+
```
lgsilva3087, line 32: There are 3 empty lines here. Please consider removing 2 of them.

lgsilva3087 (approved):
LGTM.

Not for this change, but please consider the following:

- common contains platform.h.
- common links to the sfserr library.
- sfserr library looks for plarform.h inside common.

I understand it is a header file, but probably we need to extract platform.h from common.

src/common/sfserr/saunafs_error_codes.cc
lgsilva3087: Looks like we need to review the platform.h file and update it.

CMakeLists.txt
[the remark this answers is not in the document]
lgsilva3087: My comment was about completing the word (sfsrror or sfserrors instead of sfserr). I also like only 'errors'. But I don't know how we will use this library in the future. Probably at some point we will link to most of the libraries dynamically. An unique name should not harm. Anyway, this is not a problem, just a personal preference.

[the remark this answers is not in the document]
uristdwarf: Personally, if I were including shared library header files, I would much prefer `#include "saunafs/errors.h"` rather than `#include "saunafs/sfserrors.h"`, since I already know I will be dealing with saunafs errors, so if we keep the shared headers in the directory saunafs, it should cause no problems (But it's also a personal preference). As for the library name itself, that can be decided in the CMake file and can be easily changed later if we decide to make a shared library.

PR #97, opened 2024-05-24
 tidy(common): Split off slogger from common 

uristdwarf:
This is another split of the common library to help encapsulate and
reduce size/usage of the common library, for both new and future
libraries.

PR #98, opened 2024-05-28
feat(metrics): Add initial prometheus support

uristdwarf:
This commit introduces some initial prometheus support. This is both a
compile and runtime feature. First you must have the prometheus-cpp
library installed for it to compile support for this feature. Second,
you must set ENABLE_PROMETHEUS to 1 in sfsmaster.cfg.

Currently prometheus is only implemented for master, and is simply
collecting packet data sent and received from client. In the future,
more metric data can be added, and hopefully most of CGI metrics can be
replaced with this.

This also introduces some initial support for multi-threading in master
(for prometheus) and adds documentation regarding this new feature.


uristdwarf:
N:B: When reviewing this, please only look at the the last commit. The other commits are dependencies on #96 and #97. They will be dropped when they are merged.

uristdwarf:
Putting to draft, waiting for #97

src/metrics/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+collect_sources(METRICS)
+
+add_library(metrics ${METRICS_SOURCES})
+
+if (PROMETHEUS_CPP_ENABLE_PULL)
+  target_link_libraries(metrics slogger prometheus-cpp::pull)
+endif()
+
+# create_unittest(metrics ${METRICS_TESTS})
+# link_unittest(metrics sfscommon)
```
aNeutrino, line 29: why is it here ?

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
...
+	rx_client_packets = MetricCounter(
+	    &packet_counter.Add({{"protocol", "tcp"}, {"direction", "rx"}}));
+	tx_client_packets = MetricCounter(
+	    &packet_counter.Add({{"protocol", "tcp"}, {"direction", "tx"}}));
+
+}
+
+void prometheus_loop(const char* host) {
+	try {
+		// create an http server running on port 8080
```
aNeutrino, line 84: sometimes 8080 ...
it is defined in cfg file.

aNeutrino, line 84:
```suggestion
		// create an http server
```

aNeutrino (approved):
LGTM

src/master/init.h
```diff
@@ -35,13 +38,27 @@
...
+	           if (cfg_getuint8("ENABLE_PROMETHEUS", 0) != 1) {
+		           safs::log_info(
+		               "Prometheus disabled, no Prometheus metrics will be "
+		               "gathered");
+		           return 0;
+	           }
+	           metrics::metrics_init(
+	               cfg_getstr("PROMETHEUS_HOST", "0.0.0.0:8080"));
+	           eventloop_destructregister(metrics::metrics_destroy);
+	           return 0;
```
rolysr, line 59: Consider splitting the lambda function subsection into smaller function, for code readability. If you consider it useful try something like this:  
```cpp
// Define a function for initializing Prometheus metrics
int init_prometheus_metrics() {
    auto prometheus_host = cfg_getstr("PROMETHEUS_HOST", "0.0.0.0:8080");
    metrics::metrics_init(prometheus_host);
    eventloop_destructregister(metrics::metrics_destroy);
    return 0;
}

// Update the lambda function to use the new helper function
inline const std::vector<RunTab> runTabs = {
    RunTab{[]() -> int {
        if (cfg_getuint8("ENABLE_PROMETHEUS", 0)!= 1) {
            safs::log_info("Prometheus disabled, no Prometheus metrics will be gathered");
            return 0;
        }
        return init_prometheus_metrics();
    }, "prometheus module"}
};```

File: src/metrics/metrics.cc [hunk 4]
```diff
@@ -0,0 +1,105 @@
...
+// This is used to indicate for the metrics thread to stop, it's managed by a
+// single function (metrics_destroy) and read by prometheus_loop
+std::atomic<bool> destroy = false;
+
+std::unique_ptr<std::thread> thread;
+
+// NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+void metrics_destroy() {
+	if (thread != nullptr) {
```
rolysr, line 45: Please consider, if necessary, to change the places in the code related to comparisons to `nullptr` for a constant that improve readability.

src/metrics/metrics.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace metrics {
+
+class MetricCounter {
+public:
+#ifdef HAVE_PROMETHEUS
+	explicit MetricCounter(prometheus::Counter *counter = nullptr) : counter_(counter) {}
+
+	inline void increment(double n = 1) const {
+		if (counter_ != nullptr) {
```
rolysr, line 16: Please consider also checking if changing this `nullptr` for a constant named can improve code readability.

tests/setup_machine.sh
```diff
@@ -91,7 +91,6 @@ common_packages=(
 	socat
 	sudo
 	tidy
-	time
```
rolysr, line 94: Why was it necessary to delete this common package? Just a little doubt.

rolysr (commented):
LGTM. I like this new feature. Please read my minor review suggestions.

tests/setup_machine.sh
uristdwarf: It's a mistake, thanks for noticing :)

src/metrics/metrics.cc
uristdwarf: Not sure how making nullptr a constant named variable should help? In my mind it would be an unnecessary abstraction that's more likely to cause confusion (cause then you need to check what the constant name value is).

src/metrics/metrics.h
uristdwarf: Ditto

src/master/init.h
uristdwarf: Mhm, good suggestion, but I would probably go even further and make it entirely it's own function that is then part of runTabs. I'll think about it and tinker a bit.

src/metrics/CMakeLists.txt
uristdwarf: I was thinking about creating unit tests, but then realized it would be pretty hard to test this without resorting to making HTTP requests and parsing HTML, so the integration tests were much better suited for this.

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
...
+	safs::log_err(
+	    "could not setup prometheus server: Prometheus isn't compiled with "
+	    "this program");
+}
+}
+#else
+
+using namespace prometheus;
+
+inline Family<Counter>& setup_family(const char* name, const char* help, const std::shared_ptr<Registry>& registry) {
```
lgsilva3087, line 62: Please wrap the line to 80 characters.

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
...
+		Exposer exposer{host};
+		const auto registry = std::make_shared<Registry>();
+
+		setup_client_packets(registry);
+
+		exposer.RegisterCollectable(registry);
+		safs::log_info("started prometheus server");
+
+		// Sleep forever this thread
+		while (!destroy) {};
```
lgsilva3087, line 94: This line creates a busy wait without sleep, which can be a high CPU consumer. Please consider something like this to reduce CPU load:
```cpp
while (!destroy) {
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
```
Also consider to use a std::jthread and a std::stop_token instead of a plain thread. This way you don't need to manually join the `std::jthread`, it will automatically join in its destructor.

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
...
+	} catch (std::exception &e) {
+		safs::log_err("could not setup prometheus server: {}", e.what());
+	}
```
lgsilva3087, line 97: Looks like this block is checking for errors about the prometheus server start. Probably the loop can be moved outside the try block.

src/metrics/metrics.h
```diff
@@ -0,0 +1,41 @@
...
+#endif
+};
+
+void metrics_init(const char* host);
+void metrics_destroy();
+
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+
+extern MetricCounter rx_client_packets;
+extern MetricCounter tx_client_packets;
```
lgsilva3087, line 38: Why extern here?

tests/setup_machine.sh
```diff
@@ -224,6 +228,11 @@ python_packages=(
 # determine which OS we are running and choose the right set of packages to be installed
 release="$(lsb_release -si)/$(lsb_release -sr)"
 case "${release}" in
+	Ubuntu/24.04)
+		echo $release
+		apt-get -y install ca-certificates-java # https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1911078.html
+		apt-get -y install "${common_packages[@]}" "${apt_packages[@]}" "${noble_packages[@]}"
```
lgsilva3087, line 234: According to this, prometheus support will only be available in Ubuntu 24.04?

tests/test_suites/SanityChecks/test_prometheus.sh
```diff
@@ -0,0 +1,29 @@
+if ! ldd $(whereis sfsmaster | awk '{print $2}') | grep prometheus; then
+	echo "Prometheus client isn't linked with master, skipping test..."
+	test_end
+fi
```
lgsilva3087, line 4: Nice, but what happens if there is a master without Prometheus support installed for instance in /usr/sbin? Probably the testing framework contains a direct way to know where are the testing binaries.

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
...
+// Received client packets counter
+MetricCounter rx_client_packets;
+// Sent client packets counter
+MetricCounter tx_client_packets;
+
+// This is used to indicate for the metrics thread to stop, it's managed by a
+// single function (metrics_destroy) and read by prometheus_loop
+std::atomic<bool> destroy = false;
+
+std::unique_ptr<std::thread> thread;
```
lgsilva3087, line 40: Consider a std::jthread instead.

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
+#ifdef HAVE_PROMETHEUS
```
lgsilva3087, line 1: Please add the usual Copyright and license text.

lgsilva3087 (commented):
Please see my suggestions.

src/metrics/CMakeLists.txt
aNeutrino: then it must be removed.

src/metrics/metrics.cc
aNeutrino: it will help very much...
because there reason you compare it to nullptr,
and it would be nice you express it explicitly to save time of readers ....
```suggestion
constexpr std::thread* invalidThread = nullptr;
	if (thread != invalidThread) {
```

src/metrics/metrics.cc
```diff
@@ -0,0 +1,105 @@
...
+std::atomic<bool> destroy = false;
+
+std::unique_ptr<std::thread> thread;
+
+// NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+void metrics_destroy() {
+	if (thread != nullptr) {
+		destroy = true;
+		thread->join();
```
aNeutrino, line 47:
```suggestion
		if (thread->joinable()) {
			thread->join();
		}
```

src/metrics/metrics.h
aNeutrino, line 16:
```suggestion
static constexpr prometheus::Counter* UninitializedCounter = nullptr;

	explicit MetricCounter(prometheus::Counter *counter = UninitializedCounter) : counter_(counter) {}

	inline void increment(double n = 1) const {
		if (counter_ != UninitializedCounter) {
```

aNeutrino: it looks same but is not
that is exactly the point...er.
that if we use `nullptr` without naming a reason, despite of using them for different reasons, 
then we decrease readability.
Please think why you use `nullptr` and name it, to make reading of your code smoother.

aNeutrino (approved):

doc/sfsmaster.cfg.5.adoc
```diff
@@ -238,6 +238,13 @@ chunkserver disk usage to determine most fitting chunkserver. Heavy loaded
 chunkservers will be picked for operations less frequently. (default is 0,
 correct values are in range from 0 to 0.5)
 
+*ENABLE_PROMETHEUS*:: Whether to enable Prometheus support and metric
+collection. Note that this requires compiling with Prometheus support. Set to
+either 1 to enable, or 0 to disable (default is 0)
+
+*PROMETHEUS_HOST*:: Host address where Prometheus metric data can be collected,
+must in the format of HOST:PORT (default 0.0.0.0:8080)
```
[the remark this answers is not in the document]
uristdwarf: Done

src/metrics/metrics.cc
uristdwarf: The variables will go out of scope in the try bloc if it exits it. The docs for this library clearly stated it's the caller's responsibility to keep them alive.

src/metrics/metrics.h
uristdwarf: I need other parts of master to access these variables, but if I define them here then it's (potentially) a ODR violation. Hence external here and definition in .cc

tests/setup_machine.sh
uristdwarf: > According to this, prometheus support will only be available in Ubuntu 24.04?

Yes, at least for now.

> It seems that installing on 22.04 is not straightforward https://www.cherryservers.com/blog/install-prometheus-ubuntu

You can still manually compile and have CMake/the executable detect the library (like I did on Arch, which also doesn't have this package), but I can't work on the assumption it might be there.

One idea I had was to download it manually with CMake, but it's a shared library, so it's not going to be easy to setup anyway. Another option is to provide the package in our apt repository.

src/metrics/metrics.h
lgsilva3087: Are inline variables not enough here? My concern is about later usage in shared libraries and plugins, we can end up with variables pointing to different places.

tests/setup_machine.sh
lgsilva3087: I understand, but at this point, looks like we are dropping support for master features on Ubuntu 22.04.

doc/sfsmaster.cfg.5.adoc
```diff
@@ -238,6 +238,13 @@ chunkserver disk usage to determine most fitting chunkserver. Heavy loaded
 chunkservers will be picked for operations less frequently. (default is 0,
 correct values are in range from 0 to 0.5)
 
+*ENABLE_PROMETHEUS*:: Whether to enable Prometheus support and metric
+collection. Note that this requires compiling with Prometheus support. Set to
+either 1 to enable, or 0 to disable (default is 0)
+
+*PROMETHEUS_HOST*:: Host address where Prometheus metric data can be collected,
+must be in the format of HOST:PORT (default 0.0.0.0:8080)
```
lgsilva3087, line 246: Port 8080 is a highly used one. Please consider selecting a default one less prone to conflicts with other applications. Probably 94XX like other saunafs components.

src/metrics/metrics.h
uristdwarf: Inline necessitates a lock:

`extern`:
https://godbolt.org/z/Ync9cPTEv

`inline`:
https://godbolt.org/z/zx963xx33

I'll think about redesigning this a bit, maybe use a switch statement instead of variables

doc/sfsmaster.cfg.5.adoc
uristdwarf: I went with 9499 for now.

src/metrics/metrics.h
uristdwarf: I redesigned how these variables are accessed. It now uses array subscripting with enums instead.

tests/setup_machine.sh
uristdwarf: They can still have the feature, they will just have to install the library and build the packages manually. Our (and other) 22.04 packages simply won't be able to provide the package with Prometheus support unless we figure out a way to provide the library ourselves.

src/metrics/metrics.h
```diff
@@ -0,0 +1,67 @@
...
+	};
+#ifdef HAVE_PROMETHEUS
+	Counter() : counter_(nullptr) {};
+	Counter(const prometheus::Labels &labels, prometheus::Family<prometheus::Counter> *family) : counter_(&family->Add(labels)) {};
+
+	inline void increment(double n = 1) const {
+		if (counter_ != nullptr) {
+			counter_->Increment(n);
+		}
+	}
```
uristdwarf, line 46: Move this to metric.cc, hopefully while keeping inline if possible

doc/sfsmaster.cfg.5.adoc
lgsilva3087, line 246: Please update the port. I think it was changed to 9499.

lgsilva3087 (approved):
Looks ready for initial introduction.

PR #99, opened 2024-05-28
tests: Migrate tests with xor to erasure coding

Antuan-AE:
The xor replication goal will be deprecated, so all tests xor related have been migrated to its equivalent goal representation of ec(k,1).


lgsilva3087 (approved):
LGTM.

PR #101, opened 2024-05-30
fix(saunafs-admin): Fix saunafs-admin for windows

rolysr:
The previous implementation of saunafs-admin was failing on Windows environment when trying to connect to an specific master server instance due to not correctly initializing the socket communication for Windows. This commit fixes that issue.

### DIFF
--- a/src/admin/main.cc
+++ b/src/admin/main.cc
@@ -48,6 +48,7 @@
 #include "common/human_readable_format.h"
 #include "protocol/SFSCommunication.h"
 #include "common/sfserr.h"
+#include "common/sockets.h"
 
 int main(int argc, const char** argv) {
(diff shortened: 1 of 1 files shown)


src/admin/main.cc
```diff
@@ -80,6 +81,7 @@ int main(int argc, const char** argv) {
 		if (argc < 2) {
 			throw WrongUsageException("No command name provided");
 		}
+		socketinit();
```
Antuan-AE, line 84: I can see `socketinit` function implementation is needed on Windows platform only. Please review if as a result you also need to call for function `socketrelease` as well on Windows.

rolysr: Done

Antuan-AE (approved):
LGTM

lgsilva3087 (approved):
LGTM.

PR #102, opened 2024-06-05
fix(chunkserver): Improve plugin related logs

lgsilva3087:
The previous mesages were shown as warnings, but is normal to not have plugins to load. This commit uses less severe messages in the plugin loading sections.


Antuan-AE (approved):
LGTM

ralcolea (approved):
Very good job @lgsilva3087 :+1:

PR #103, opened 2024-06-07
 admin: Add USE_LINEAR_ASSIGNMENT_OPTIMIZER to master default config

dmga44:
The recently added ```USE_LINEAR_ASSIGNMENT_OPTIMIZER``` option to master server was missing in the list of master default options, and should be shown there.

### DIFF
--- a/src/admin/dump_config_command.cc
+++ b/src/admin/dump_config_command.cc
@@ -102,6 +102,7 @@ const static std::unordered_map<std::string, std::string> defaultOptionsMaster =
     {"CHUNKS_LOOP_MIN_TIME", "300"},
     {"CHUNKS_LOOP_PERIOD", "1000"},
     {"CHUNKS_LOOP_MAX_CPU", "60"},
+    {"USE_LINEAR_ASSIGNMENT_OPTIMIZER", "1"},
     {"CHUNKS_SOFT_DEL_LIMIT", "10"},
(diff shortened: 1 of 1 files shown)


uristdwarf:
I'm going on a wild guess that the YAML library is not correctly linked on Windows? I've no idea how this dependency is handled in Windows, since this only applies to master and saunafs-admin. From what I can read, it's using MinGW to build the admin command?

uristdwarf:
I'd try building the library https://github.com/jbeder/yaml-cpp on Windows 10 and getting a simple example to work, and then trying to link that build library to the admin binary.

dmga44:
> From what I can read, it's using MinGW to build the admin binary?

Yes. We are using MinGW to build it everywhere.

dmga44:
I've changed the name of the PR and the description to reflect only the changes introduced by it.

Fixing the saunafs-admin in Windows will take other approach.

uristdwarf (approved):

ralcolea (approved):
LGTM @dmga44. Please, consider updating the commit message with a brief explanation about what the new added option does and the expected behavior by enabling this option.

ralcolea:
> I thought this kind of explanation was given in the commit introducing this option (within #53 changes).
 
I saw the original commit introducing the option and you're right. I think there is no point to add more details. IMHO, the current commit message reflects the change applied.

PR #105, opened 2024-06-10
fix(mount): Fix Windows master connection checking

rolysr:
In the previous implementation, the client was not properly working when
the master server was restarted and first getattr operations where
executed on Windows. This change makes the mount to first try to
properly reconnect to master and if not, directly assume no data was
received from master, thus avoid the mount process to crash when
trying to use the mountpoint.


uristdwarf (approved):
I've no complaints

lgsilva3087 (approved):
LGTM

PR #106, opened 2024-06-10
tidy(tests): Disable failing tests

uristdwarf:
* Move upgrade tests to their own suite
* Move failing tests to their own suite
* Enable Long and Short system tests for branches


tests/test_suites/LongManualTests/test_advanced_readahead_parameters.sh
ralcolea, line 1: Why this test was moved to `LongManualTests` instead of `LongSystemTests`?
I don't see the relationship between the test and the name of the test suite.

ralcolea (approved):
Good job @uristdwarf. My only concern is about the creation of new (maybe unneeded) test suites.

I think that `test_advanced_readahead_parameters` can be included in `LongSystemTests` or even stay in the `ShortSystemTests`, as it was originally.

tests/test_suites/LongManualTests/test_advanced_readahead_parameters.sh
uristdwarf: The problem is this test doesn't work on the CI, and the reason is because it requires running in a privileged container. There was some discussion some time ago between me @ictus4u about privileged containers related to some other tests (that also don't work)

aNeutrino (approved):
LGTM

PR #107, opened 2024-06-10
feat(chunkserver): Introduce DiskEnergyManager

lgsilva3087:
Previously, the disk assignment was done in a function inside the hddspacemgr.cc file, which works fine for basic placement of chunks in the disks, but is not enough if an advanced power management system is required.

This commit introduces the IDiskEnergyManager interface and the DefaultDiskEnergyManager concrete implementation, to provide the extension points for a future more complex energy management system.

So far, the only virtual function is getDiskForNewChunk, but others will appear soon.

The DefaultDiskEnergyManager provides the same distribution as before.

### DIFF
--- a/src/chunkserver/chunkserver-common/disk_interface.h
+++ b/src/chunkserver/chunkserver-common/disk_interface.h
@@ -8,6 +8,8 @@
 #include "common/chunk_part_type.h"
 #include "common/disk_info.h"
 
+#define DiskNotFound nullptr
+
 class IChunk;
 
 /// Represents a data disk in the Chunkserver context.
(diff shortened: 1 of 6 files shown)


ralcolea (approved):
Great job!!! :fire: :rocket: :+1:

src/chunkserver/chunkserver-common/disk_interface.h
```diff
@@ -8,6 +8,8 @@
 #include "common/chunk_part_type.h"
 #include "common/disk_info.h"
 
+#define DiskNotFound nullptr
```
dmga44, line 11: Shouldn't we use ```constexpr``` here?

dmga44 (approved):
LGTM, though I had a question.

rolysr (approved):

src/chunkserver/chunkserver-common/disk_interface.h
lgsilva3087: Yes, but changing it introduce changes not related with this feature. It will be addressed in another change.

PR #108, opened 2024-06-11
fix(tests): Fix test_cgi_validate_html test

uristdwarf:
The test could sometimes fail due to a invalid default host (perhaps because there is already a mount on the server for another host). This change makes it explicit to use localhost as the host name.

### DIFF
--- a/tests/tools/saunafs.sh
+++ b/tests/tools/saunafs.sh
@@ -701,7 +701,8 @@ add_cgi_server_() {
 	saunafs-cgiserver -P "$cgi_server_port" -p "$pidfile"
 	saunafs_info_[cgi_pidfile]=$pidfile
 	saunafs_info_[cgi_port]=$cgi_server_port
-	saunafs_info_[cgi_url]="http://localhost:$cgi_server_port/sfs.cgi?masterport=${saunafs_info_[matocl]}"
+	saunafs_info_[cgi_url]="http://localhost:$cgi_server_port/sfs.cgi?masterhost=localhost&masterport=${saunafs_info_[matocl]}"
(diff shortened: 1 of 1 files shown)


aNeutrino (approved):
LGTM

dmga44 (approved):
LGTM.

Though I have another issue to point out. Maybe it is just my laptop, which might be very slow or probably some wrong setup, but I ran the test a few times and several times it timed out. Have some of you stumbled upon this issue? Maybe we could increase the test time limit to 5 or 6 minutes?

rolysr (approved):

rolysr:
> Though I have another issue to point out. Maybe it is just my laptop, which might be very slow or probably some wrong setup, but I ran the test a few times and several times it timed out. Have some of you stumbled upon this issue? Maybe we could increase the test time limit to 5 or 6 minutes?

From my side, I repeated the test 10 times in a row and all them passed.

uristdwarf:
> Though I have another issue to point out. Maybe it is just my laptop, which might be very slow or probably some wrong setup, but I ran the test a few times and several times it timed out. Have some of you stumbled upon this issue? Maybe we could increase the test time limit to 5 or 6 minutes?

I've never seen this issue to be honest. Would you know where it gets stuck? In some place of the test? Try adding some echo statements and see which part of it gets stuck, it may not be related to the test necessarily, but other parts of the code.

I'll merge it for now, but I would be interested to know what you find out.

dmga44:
> Would you know where it gets stuck? In some place of the test? 

Well, it is not getting stuck anywhere, it is on the edge: it took 233s when it passed (out of 240s).

PR #113, opened 2024-06-17
fix: Fix slogger dependencies compilation

rolysr:
First split off of slogger related code from
common libraries were compiled with shared
libraries approach, making it not portable to
compile in other systems, this commit fixes that.

### DIFF
--- a/src/slogger/CMakeLists.txt
+++ b/src/slogger/CMakeLists.txt
@@ -1,5 +1,4 @@
 collect_sources(SLOGGER)
 
 shared_add_library(slogger ${SLOGGER_SOURCES})
-target_link_libraries(slogger sfserr)
-target_link_libraries(slogger_pic sfserr_pic)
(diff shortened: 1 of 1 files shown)


uristdwarf (approved):

ralcolea (commented):
Despite looking good and compiling well, this change makes fail consistently `test_metadata_dump_period_seconds` from ShortSystemTests suite.

I tested locally and it also fails. However, `dev` branch does not fail with this test. It makes me think that the issue is in the proposed change.

Can you check looking for possible causes to make pass the failing test?

uristdwarf:
> Despite looking good and compiling well, this change makes fail consistently test_metadata_dump_period_seconds from ShortSystemTests suite.

No, this test fails randomly, see #110 for more details.

Just rerun the test in CI again.

aNeutrino (approved):

PR #114, opened 2024-06-19
tests: Update GaneshaTests suite

ralcolea:
This commit applies a patch to the GaneshaTests suite that contains fixes and improvements to parameters like timeout, Lease_Lifetime and iodepth (fio).

The most important changes are:
- Adjust timeout of tests.
- Set Lease_Lifetime equal to Grace_Period because is not recommended to use a Lease_Lifetime less than Grace_Period.
- Replace sleep calls before connecting nfs client with assert eventually macro.
- Set iodepth=1 for fio mixed tests because fio does not allow to use iodepth=4 for these mixed workloads.
- Remove unneeded validation of fio installation for tests using dd.


lgsilva3087 (approved):
LGTM

PR #118, opened 2024-06-21
feat(chunkserver): Add disk manager interface

lgsilva3087:
This PR adds the IDiskManager interface and the extensions to the PluginManager to load plugins not based in DiskPlugin (like disk manager plugins).


ralcolea (approved):
LGTM :fire:

PR #119, opened 2024-06-21
fix(errors): Add explicit std string include

lgsilva3087:
In some configurations, the compiler does not find the std string header automatically. This commit just adds the header to avoid these scenarios.


ralcolea (approved):
LGTM.

PR #125, opened 2024-06-26
fix(mount): help display format

rolysr:
This change improves readability of current
mount possible options.

### DIFF
--- a/src/mount/fuse/mount_config.cc
+++ b/src/mount/fuse/mount_config.cc
@@ -114,20 +114,63 @@ void usage(const char *progname) {
 	printf(
 "usage: %s  [HOST[/PORT]:[PATH]] [options] mountpoint\n"
 "\n", progname);
-
-	printf("general options:\n");
 	fuse_cmdline_help();
 
 	printf(
-"SFS options:\n"
+"\nGeneral options:\n"
 "    -c CFGFILE                  equivalent to '-o sfscfgfile=CFGFILE'\n"
-"    -m   --meta                 equivalent to '-o sfsmeta'\n"
 "    -H HOST                     equivalent to '-o sfsmaster=HOST'\n"
 "    -P PORT                     equivalent to '-o sfsport=PORT'\n"
 "    -B IP                       equivalent to '-o sfsbind=IP'\n"
 "    -S PATH                     equivalent to '-o sfssubfolder=PATH'\n"
+"\n"
+"Security/permissions related options:\n"
+"    -o sfspassword=PASSWORD     authenticate to sfsmaster with password\n"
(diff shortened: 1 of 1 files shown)


uristdwarf:
Please follow this https://www.conventionalcommits.org/en/v1.0.0/ when making PR's, at least with the PR title, but preferably with the commits as well if we want to rebase. This is to help us automate releases in the future.

src/mount/fuse/mount_config.cc
```diff
@@ -114,28 +114,75 @@ void usage(const char *progname) {
...
 "\n", progname);
-
-	printf("general options:\n");
 	fuse_cmdline_help();
 
 	printf(
-"SFS options:\n"
+"Aditional info options\n"
+"    -o sfsdebug                 print some debugging information\n"
+"\n"
```
lgsilva3087, line 122: Consider moving this one to the end. It is usually for debugging and should not be before the General options.

src/mount/fuse/mount_config.cc
```diff
@@ -114,28 +114,75 @@ void usage(const char *progname) {
...
+"Security/permissions related options:\n"
+"    -o sfspassword=PASSWORD     authenticate to sfsmaster with password\n"
+"    -o sfsmd5pass=MD5           authenticate to sfsmaster using directly "
+				"given md5 (only if sfspassword is not defined)\n"
 "    -p   --password             similar to '-o sfspassword=PASSWORD', but "
 				"show prompt and ask user for password\n"
+"    -o askpassword              show prompt and ask user for password\n"
+"    -o sfsdonotrememberpassword do not remember password in memory - more "
+				"secure, but when session is lost then new "
+"\n"
```
lgsilva3087, line 142: We need an extra empty line here.

lgsilva3087 (approved):
Nice change, indeed it looks better. Please see my small comments.

src/mount/fuse/mount_config.cc

src/mount/fuse/mount_config.cc
rolysr: Done

dmga44:
Nice change, @rolysr. Getting things more similar to our WC :)

Though, I think you should check the rule of the alignment of the options. The description should start at some fixed point, or at least have two spaces between the format of the option and the start of the description. For instance, I think it does not hold for:
- sfsdonotrememberpassword
- cacheexpirationtime

I would also like to not have group of options with a single option, but that is kind of more of my likings.

rolysr:
> * cacheexpirationtime

Done

dmga44 (approved):

PR #126, opened 2024-06-27
tidy(INSTALL.md): Update links

uristdwarf:


INSTALL.md
```diff
@@ -4,15 +4,15 @@ You can install SaunaFS from pre-built packages or from source. The
 pre-built packages are currently available for the following platforms:
 
 - Linux (x86_64)
-  - [Ubuntu 24.04 (Noble)](#ubuntu)
-  - [Ubuntu 22.04 (Jammy)](#ubuntu)
+  - [Ubuntu 24.04 (Noble)](#ubuntu-22042404)
```
ralcolea, line 7: For me, there is no reason to have the same link twice pointing to three lines below. I recommend you to remove links and only mention that supported platforms are Ubuntu 22.04 and 24.04.

ralcolea (commented):
@uristdwarf I think there is no point to continue using links.

INSTALL.md
uristdwarf: I was thinking it would be nice to have have the links in case we want to add other distributions as well.

uristdwarf: Sadly, this doesn't work on Github and VS Code.

I'll just remove the links right now, we can add it back later when we add more distribution support

ralcolea (approved):
LGTM

PR #127, opened 2024-06-30
test: Add function to stop hanging ganesha daemon

ralcolea:
Previously, if a Ganesha test failed, very often next tests were not able to continue running because Ganesha daemon became hanging and nfs mounts stayed mounted. Besides, CI containers passed to unhealthy state and after some time, Jenkin nodes needed to be restarted.

This commit adds `stop_ganesha` function to the testing framework to handle hanging ganesha daemon and mounted SaunaFS/nfs mounts. It allows a proper cleanup of resources and handles potential hanging processes. It also should avoid CI containers to pass to an unhealthy state.

### DIFF
--- a/tests/run-test.sh
+++ b/tests/run-test.sh
@@ -53,10 +53,26 @@ unwrap_generators() {
 	fi
 }
 
+stop_ganesha() {
+	# umount all hanging SaunaFS mounts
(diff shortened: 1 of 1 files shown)


uristdwarf (approved):

lgsilva3087 (approved):
Great.

PR #129, opened 2024-07-02
feat(common): Improve the hddSizeParse function

lgsilva3087:
The previous version was completely implemented from the scratch. The proposed version uses modern C++ and std::stod to increase readability and safety. The commit makes it also more flexible with different suffixes for the size and improves the error handling.

The function initially was in the hddspacemgr.cc in the chunkserver, but it is very general and will be used from the plugins soon, so it makes sense to move it to the common library.

An unit test file was also added to verify different combinations.

### DIFF
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -2510,97 +2510,6 @@ void hddTerminate(void) {
 	gDisks.clear();
 }
 
-int hddSizeParse(const char *str, uint64_t *ret) {
-	TRACETHIS();
-	uint64_t val, frac, fracdiv;
-	double drval, mult;
-	int f;
-	val = 0;
-	frac = 0;
-	fracdiv = 1;
-	f = 0;
-	while (*str >= '0' && *str <= '9') {
-		f = 1;
-		val *= 10;
-		val += (*str - '0');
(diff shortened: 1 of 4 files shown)


src/chunkserver/hddspacemgr.cc
```diff
@@ -2510,97 +2510,6 @@ void hddTerminate(void) {
...
-	while (*str >= '0' && *str <= '9') {
-		f = 1;
-		val *= 10;
-		val += (*str - '0');
-		str++;
-	}
-	if (*str == '.') { // accept format ".####" (without 0)
-		str++;
-		while (*str >= '0' && *str <= '9') {
-			fracdiv *= 10;
```
dmga44, line 2531: Maybe low priority issue, but ```frac``` and ```fracdiv``` may overflow when the number of digits after the dot increases to around 20 (the value of ```fracdiv``` should be 10^20 which overflows in uint64_t).

Probably you could use ```long double``` for large precission or other floating point types to get through it.

dmga44 (approved):
Nice job, @lgsilva3087 !

rolysr (approved):
LGTM

ralcolea (approved):
Great job!!! :fire: :+1: :rocket:

PR #130, opened 2024-07-02
feat(chunkserver): Relocate hddDisksReinit

lgsilva3087:
This change moves the hddDisksReinit function to the chunkserver-common library, to make it available for the plugins. DiskManagers should reimplement the virtual functions to provide custom behavior.

Driven by the movement, other related changes were applied:
- Rename the function to reloadDisksFromCfg which is more consistent.
- Move the gPluginManager to be visible from plugins, because it knows how to create the concrete disks. In this case, CmrDisk was needed as it is the default.
- The includes were also updated due to the movements.
- The DefaultDiskManager was also moved, to provide a base behavior.

The logic of the functions related to the disks loading was not modified. They belong now to the current disk manager.


src/chunkserver/chunkserver-common/default_disk_manager.cc
```diff
@@ -0,0 +1,305 @@
...
+
+			delete currentDisk;  // Also deletes the lock files (smart pointer)
+			return 1;
+		}
+	}
+
+	gDisks.emplace_back(currentDisk);
+	gResetTester = true;
+
+	return 2;
```
rolysr, line 143: Consider to return constant names instead of values. Or add a little comment about their meaning or possible values returned by the function if you consider it necessary

ralcolea (approved):
LGTM

src/chunkserver/chunkserver-common/default_disk_manager.h
```diff
@@ -47,6 +47,11 @@ class DefaultDiskManager : public IDiskManager {
 	/// Virtual destructor needed for correct polymorphism
 	virtual ~DefaultDiskManager() = default;
 
+	virtual int parseCfgLine(std::string hddCfgLine);
+
```
uristdwarf, line 51: nit: If DefaultDiskManager only uses this, maybe it should be private?

uristdwarf (approved):

src/chunkserver/chunkserver-common/default_disk_manager.h
lgsilva3087: The derived classes will reimplement it. The unit tests also will use it.

src/chunkserver/chunkserver-common/default_disk_manager.cc
lgsilva3087: There is no change in the logic in this PR. I agree with the comment, but this function will change soon. I did not change the original content.

uristdwarf:
Also, if the function didn't change, then it's probably a refactor instead of a feat?

PR #131, opened 2024-07-03
ci(.github): Add check for Conventional Commits

uristdwarf:
This is to make sure all future commits are written according to this document https://www.conventionalcommits.org/en/v1.0.0/. This will allow us to automate release note generations


ralcolea (approved):
Great change @uristdwarf!! :fire: :fire: :fire: 
This new check will help us to automate release notes!!!

PR #132, opened 2024-07-04
feat(tests): Allow custom sfshdd content in tests

lgsilva3087:
This change extends the testing framework to support custom sfshdd.cfg content directly from the tests. For example, to set the content for the chunkserver with index 3, provide the following line before calling the setup_local_empty_saunafs function:

CHUNKSERVER3_HDD_CONTENT="YOUR CUSTOM HDD CONTENT HERE"

This feature facilitates testing of new sfshdd formats and adds flexibility for tests like faking broken drives or using existing real devices.

The existing function 'create_sfshdd_cfg_' was also modified to only create the content. 'create_emulated_zoned_devices_for_cs_number_' will handle now the emulated devices creation, which is more consistent with their names.

### DIFF
--- a/tests/tools/saunafs.sh
+++ b/tests/tools/saunafs.sh
@@ -449,48 +449,78 @@ prepare_metalogger_() {
 	saunafs_info_[metalogger_cfg]="$etcdir/sfsmetalogger.cfg"
 }
 
+baseMetadataPath="/mnt/ramdisk/metadata/sauna_nullb"
+baseDataPath="/mnt/zoned/sauna_nullb"
+
+# Creates the needed emulated zoned devices for a given chunkserver number.
+#
+# The function also creates the needed directories for the metadata and data
+# paths of the chunkserver. The metadata path is created in the ramdisk and the
+# data path is created in the zoned device. The function also formats and mounts
+# the zoned device using the zonefs filesystem.
+#
+# Parameters:
+# - chunkserver_index (optional): The chunkserver number.
+#   Defaults to "chunkserver_id".
+#
+# Usage:
+# create_emulated_devices_for_cs_number_ [chunkserver_index]
+#
+create_emulated_zoned_devices_for_cs_number_() {
+	if [[ ! $use_zoned_disks || ! $use_ramdisk ]]; then
+        return
+    fi
+
+	local chunkserver_index=${1:-chunkserver_id}
+	local n=${disks_per_chunkserver}
+	local zoned_prefix="zonefs:"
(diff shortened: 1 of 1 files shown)


tests/tools/saunafs.sh
```diff
@@ -449,48 +449,75 @@ prepare_metalogger_() {
...
+		# Mount point
+		if [ ! -d "${dataPath}" ]; then
+			sudo mkdir -pm 777 "${dataPath}"
+		fi
 
-			metadataPath="/mnt/ramdisk/metadata/sauna_nullb${disk_number}"
-			dataPath="/mnt/zoned/sauna_nullb${disk_number}"
+		sudo mount -t zonefs "${devPath}" "${dataPath}"
+	done
+}
```
uristdwarf, line 507: I'd try to minimize root usage as much as possible. Why not put the dataPath under the /tmp/ directory or directories that saunafs-tests already uses?

tests/tools/saunafs.sh
```diff
@@ -449,48 +449,75 @@ prepare_metalogger_() {
...
-			fi
+	if [[ $use_zoned_disks && $use_ramdisk ]]; then
+		zoned_prefix="zonefs:"
+		local firstDisk=$(echo "${chunkserver_id} * ${n}" | bc)
+		local lastDisk=$(echo "${firstDisk} + ${n} - 1" | bc)
 
-			sudo mount -t zonefs "${devPath}" "${dataPath}"
+		for disk_number in $(seq -w ${firstDisk} ${lastDisk}); do
+			local metadataPath="/mnt/ramdisk/metadata/sauna_nullb${disk_number}"
+			local dataPath="/mnt/zoned/sauna_nullb${disk_number}"
```
uristdwarf, line 520: Potential for bugs here, since if the paths differ from the function above, this may not be accurate information (e.g a later change changes the paths in the function above, but not here, or vice versa).

uristdwarf (approved):
No blockers from me, two comments but feel free to ignore/close them, they are rather minor.

tests/tools/saunafs.sh
lgsilva3087: Nice catch, thanks. Done.

tests/tools/saunafs.sh
lgsilva3087: I agree, but probably that's for another change. In this change I am just separating the functions and using similar places as the setup_machine uses for ramdisk, loop devices, etc.

rolysr (approved):
LGTM

PR #133, opened 2024-07-08
feat(master): Allow localhost connections to master

uristdwarf:
Since MooseFS, localhost couldn't be used to connect to master in certain circumstances. Since this is a valid use case (especially for testing), this commit removes the checks disallowing connections from localhost.


lgsilva3087:
> Since MooseFS, localhost couldn't be used to connect to master in certain circumstances. Since this is a valid use case (especially for testing), this commit removes the checks disallowing connections from localhost.

Do you know why it was forbidden before? Just to know.

lgsilva3087 (approved):
LGTM.

uristdwarf:
> Do you know why it was forbidden before? Just to know.
AFAIK, it was because of mix-up of 127 and 172

PR #134, opened 2024-07-09
fix(build): Fix cmake syntax warning

rolysr:
The content of file cmake/Packing.cmake was producing a syntax warning during fresh installation due to a syntax issue, this commit solves that.


lgsilva3087 (approved):
LGTM

PR #135, opened 2024-07-10
fix(mount): Fix random case option parsing

rolysr:
For allowing the client users to input option with random case formats, this change fixes that on the client argument processing.


src/mount/fuse/main.cc
```diff
@@ -427,6 +427,59 @@ static size_t count_colons_in_str(const char *str, size_t len) {
 	return colons_count;
 }
 
+/**
```
ralcolea, line 430: Consider this Copilot suggestions as possible improvements for your code. This version uses modern `std::string` and STL iterators to make easier and more readable the code.

Please, feel free to test and make the changes as you consider more correct.
```
#include <cctype> // for std::tolower
#include <cstring> // for strlen, strcmp
#include <string> // for std::string

// Normalize a single argument's casing
void normalize_argument_casing(char *arg) {
    std::string argument(arg);
    bool should_apply_case_change = true;

    for (auto &c : argument) {
        if (c == ',') {
            should_apply_case_change = true;
        } else if (c == '=') {
            should_apply_case_change = false;
        } else if (should_apply_case_change) {
            c = std::tolower(c);
        }
    }

    // Copy the modified string back to the original C-string argument
    std::copy(argument.begin(), argument.end(), arg);
    arg[argument.size()] = '\0'; // Null-terminate the modified C-string
}

// Normalize options casing
void normalize_options_casing(fuse_args &args) {
    bool is_previous_dash_o = false;

    for (int index = 1; index < args.argc; ++index) {
        std::string current_arg(args.argv[index]);
        if (is_previous_dash_o || (current_arg.length() > 2 && current_arg[0] == '-' &&
                                   (current_arg[1] == 'o' || current_arg[1] == '-'))) {
            if (!is_previous_dash_o) {
                // Skip the initial "-o" or "--" prefix
                current_arg = current_arg.substr(2);
                normalize_argument_casing(&args.argv[index][2]);
            } else {
                normalize_argument_casing(args.argv[index]);
            }
            is_previous_dash_o = false;
        } else {
            is_previous_dash_o = (current_arg == "-o");
        }
    }
}
```

ralcolea (approved):
LGTM. Please see my suggestion as possible improvements for your code if you consider it.

src/mount/fuse/main.cc
```diff
@@ -427,6 +427,59 @@ static size_t count_colons_in_str(const char *str, size_t len) {
...
+		}
+		current_char++;
+	}
+}
+
+/**
+ * @brief Normalize options casing by converting uppercase characters in the
+ * options names to its lowercase equivalent.
+ */
+void normalize_options_casing(fuse_args &args) {
```
lgsilva3087, line 464: This function is a good candidate for unit testing. Please consider it.

lgsilva3087 (commented):
Please see my suggestions.

src/mount/fuse/main.cc
rolysr: Done, now is more readable thanks to this approach.

src/mount/fuse/main.cc
```diff
@@ -427,6 +427,59 @@ static size_t count_colons_in_str(const char *str, size_t len) {
...
+		if (*current_char == ',') {
+			should_apply_case_change = true;
+		} else if (*current_char == '=') {
+			should_apply_case_change = false;
+		} else if (should_apply_case_change) {
+			to_lowercase_if_uppercase(*current_char);
+		}
+		current_char++;
+	}
+}
```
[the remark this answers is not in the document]
rolysr: I considered to use this approach but for the specifics of the options strings, I need to check the limits and exact places for ',' and '=' to correctly separate variable names from values. If I know beforehand the exact range where I should apply the transform function, then it was needed before to make at least one iteration check for this. This is necessary as I should avoid to affect variable values.

src/mount/fuse/main.cc
rolysr: Done, I added a unittest. Please, check it in case some other improve could be necessary.

lgsilva3087 (approved):

aNeutrino (approved):
LGTM

src/mount/option_casing_normalization.h
```diff
@@ -0,0 +1,75 @@
...
+ * @brief Normalize options casing by converting uppercase characters in the
+ * options names to its lowercase equivalent.
+ */
+void normalize_options_casing(struct fuse_args &args) {
+	bool is_previous_dash_o = false;
+	for (int index = 1; index < args.argc; index++) {
+		size_t current_arg_len = strlen(args.argv[index]);
+		if (is_previous_dash_o ||
+		    (current_arg_len > 2 && args.argv[index][0] == '-' &&
+		     (args.argv[index][1] == 'o' || args.argv[index][1] == '-'))) {
```
aNeutrino, line 64: @rolysr please do not use ' ' for ident uset TAB instead ALWAYS
and setup your editor corectly for this.
```suggestion
			(current_arg_len > 2 && args.argv[index][0] == '-' &&
				(args.argv[index][1] == 'o' || args.argv[index][1] == '-'))) {
```

aNeutrino (changes requested):

src/mount/option_casing_normalization.h
rolysr: Done

aNeutrino (approved):
LGTM

PR #136, opened 2024-07-10
Improvements to uRaft

uristdwarf:
This is a patch series. _**DO NOT MERGE OR SQUASH THIS, DO REBASE INSTEAD**_

Generally refactoring and fixing clang-tidy warnings/hints. This is in preparation for some fundamental changes to how uRaft works. It also improves logging (by logging to both syslog _and_ `stderr`) and fixes a bug with key-values not being correctly parsed from the uRaft helper script.


src/slogger/slogger.cc
```diff
@@ -44,6 +47,54 @@ static safs::log_level::LogLevel log_level_from_syslog(int priority) {
...
+	case LogLevel::info:
+		return "info";
+	case LogLevel::warn:
+		return "warn";
+	case LogLevel::err:
+		return "err";
+	case LogLevel::critical:
+		return "crit";
+	case LogLevel::off:
+		break;
```
Antuan-AE, line 93: Consider adding `default` clause to avoid possible future errors in case an integer is cast to LogLevel type

wg1k (dismissed):

src/slogger/slogger.cc
uristdwarf: I think in this case it should be an assert, as if it's anything, it's most likely a bug in the program

src/uraft/uraftcontroller.cc
```diff
@@ -1,41 +1,43 @@
...
+      check_node_status_timer_(ios),
+      cmd_timeout_timer_(ios),
+      command_pid_(-1),
+      command_type_(kCmdNone),
+      force_demote_(false),
+      node_alive_(false) {
+
+	opt_.check_node_status_period = DEFAULT_CHECK_NODE_STATUS_PERIOD;
+	opt_.check_cmd_status_period  = DEFAULT_CMD_STATUS_PERIOD;
+	opt_.getversion_timeout       = DEFAULT_GETVERSION_TIMEOUT;
```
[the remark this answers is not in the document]
uristdwarf: This was done by clang format, not me

uristdwarf:
> @uristdwarf The github rebase button seems to do the opposite to what we want (it brings the changes in the opposite direction, leaving a weird shape in the graph), please, rebase from your computer if it is your will.

I actually want it to be fast-forwarded, so that all commits are preserved in dev. But maybe a merge will suffice.

PR #137, opened 2024-07-10
feat(chunkserver): Extend getDiskForNewChunk

lgsilva3087:
The functions based on IDiskManager::getDiskForNewChunk now accept a ChunkPartType as argument, to allow disk assignments based on the this type. The default implementation does not use it yet, but it is needed for the plugins.


rolysr (approved):
LGTM.

ralcolea (approved):
LGTM

PR #140, opened 2024-07-15
fix: Update only error/warning messages to stdout

rolysr:
Some messages from the system are printed to stderr when they do not represent a warning or an error. This commit fixes that issue.


uristdwarf:
stderr isn't only for errors, but for logging and diagnostics as well. Essentially, stderr should be used for things that are not considered as output of the program. It seems to be the case here?

uristdwarf:
https://unix.stackexchange.com/a/331808

lgsilva3087 (approved):
LGTM for this particular case.

lgsilva3087:
Looks like this change will not be needed.

PR #141, opened 2024-07-16
test: Remove Ganesha test for master failover

ralcolea:
Commit 661c15e added a test to detect file corruption in IO operations using NFS mount during master failover. This test has started failing by timeout in CI, preventing the approval of PRs to the dev branch and future releases.

This commit removes the test until we can identify and resolve the cause of the failure.


uristdwarf (approved):
The test should be fixed before another release is made

uristdwarf:
>
The test should be fixed before another release is made

Although to be fair, this test wasn't in the last release

ralcolea:
> The test should be fixed before another release is made

I agree that the test should be fixed. However, I don't know why the test fails in CI.
The test is passing for me locally. It looks like Ganesha in CI does not behave equal, and considering there are other higher priorities now, I think the best is to postpone the work to fix this test.

lgsilva3087 (approved):

PR #144, opened 2024-07-18
feat: Allow master to load legacy metadata

Antuan-AE:
The master server now loads legacy metadata `mfs` if no actual metadata is found. This allows easy migration from legacy systems where metadata is store in legacy format. Shadow server also can download legacy metadata and convert it to new format after service is restarted or dump is trigger by admin tool. Loading of actual metadata takes precedence over legacy and if both are present, actual is loaded instead and user is warned to delete legacy.

### DIFF
--- a/src/common/metadata.cc
+++ b/src/common/metadata.cc
@@ -41,6 +41,9 @@ const char kMetadataFilename[] = METADATA_FILENAME_TEMPL;
 const char kMetadataTmpFilename[] = METADATA_FILENAME_TEMPL ".tmp";
 const char kMetadataEmergencyFilename[] = METADATA_FILENAME_TEMPL ".emergency";
 #undef METADATA_FILENAME_TEMPL
+#define METADATA_LEGACY_FILENAME_TEMPL "metadata.mfs"
+const char kMetadataLegacyFilename[] = METADATA_LEGACY_FILENAME_TEMPL;
+#undef METADATA_LEGACY_FILENAME_TEMPL
 #define METADATA_ML_FILENAME_TEMPL "metadata_ml.sfs"
 const char kMetadataMlFilename[] = METADATA_ML_FILENAME_TEMPL;
 const char kMetadataMlTmpFilename[] = METADATA_ML_FILENAME_TEMPL ".tmp";
@@ -96,7 +99,7 @@ uint64_t metadataGetVersion(const std::string& file) {
 
 	if (signature == sfsSignature || signature == sauSignature) {
 		memcpy(eofmark,"[SFS EOF MARKER]",16);
-	} else if (memcmp(chkbuff, legacySignature.data(), legacySignature.size()) == 0) {
+	} else if (signature == legacySignature) {
 		safs_pretty_syslog(LOG_WARNING,
 		                   "Legacy metadata section header %s, was detected in the metadata file %s",
 		                   legacySignature.c_str(), file.c_str());
--- a/src/common/metadata.cc
+++ b/src/common/metadata.cc
@@ -92,9 +92,15 @@ uint64_t metadataGetVersion(const std::string& file) {
 	std::string signature = std::string(chkbuff, 8);
 	std::string sfsSignature = std::string(SFSSIGNATURE "M 2.9");
 	std::string sauSignature = std::string(SAUSIGNATURE "M 2.9");
+	std::string legacySignature = std::string("LIZM 2.9");
 
 	if (signature == sfsSignature || signature == sauSignature) {
 		memcpy(eofmark,"[SFS EOF MARKER]",16);
+	} else if (memcmp(chkbuff, legacySignature.data(), legacySignature.size()) == 0) {
+		safs_pretty_syslog(LOG_WARNING,
+		                   "Legacy metadata section header %s, was detected in the metadata file %s",
+		                   legacySignature.c_str(), file.c_str());
+		memcpy(eofmark,"[MFS EOF MARKER]",16);
 	} else {
 		close(fd);
 		throw MetadataCheckException("Bad EOF MARKER in the metadata file.");
--- a/src/master/filesystem.cc
+++ b/src/master/filesystem.cc
@@ -216,9 +216,28 @@ int fs_loadall(void) {
 	changelogsMigrateFrom_1_6_29("changelog");
 	if (fs::exists(kMetadataTmpFilename)) {
 		throw MetadataFsConsistencyException(
-				"temporary metadata file (" + std::string(kMetadataTmpFilename) + ") exists, metadata directory is in dirty state");
+		    "temporary metadata file (" + std::string(kMetadataTmpFilename) + ") exists,"
+		    " metadata directory is in dirty state");
 	}
-	if ((metadataserver::isMaster()) && !fs::exists(kMetadataFilename)) {
+	std::string metadataFile;
+	bool metadataFileExists = fs::exists(kMetadataFilename);
+	bool legacyMetadataFileExists = fs::exists(kMetadataLegacyFilename);
+	if (metadataFileExists) {
+		metadataFile = kMetadataFilename;
+	}
+	if(metadataFileExists && legacyMetadataFileExists) {
+		metadataFile = kMetadataFilename;
+		safs_pretty_syslog(LOG_WARNING, "There are two metadata files in the data path: %s and %s."
+		                   " Please remove the legacy one (%s) to avoid damage to your storage.",
+		                   kMetadataFilename, kMetadataLegacyFilename, kMetadataLegacyFilename);
+	}
+	if (!metadataFileExists && legacyMetadataFileExists) {
+		metadataFile = kMetadataLegacyFilename;
+		safs_pretty_syslog(LOG_WARNING, "Only Legacy metadata file %s found and will be loaded instead."
+		                   " You should delete legacy metadata %s on next restart after new metadata %s is created ",
+		                   metadataFile.c_str(), kMetadataLegacyFilename, kMetadataFilename);
+	}
+	if (metadataserver::isMaster() && !metadataFileExists && !legacyMetadataFileExists) {
 		fs_unlock();
 		std::string currentPath = fs::getCurrentWorkingDirectoryNoThrow();
 		throw FilesystemException("can't open metadata file "+ currentPath + "/" + kMetadataFilename
@@ -229,7 +248,7 @@ int fs_loadall(void) {
 
 	{
 		auto scopedTimer = util::ScopedTimer("metadata load time");
-		fs_loadall(kMetadataFilename, 0);
+		fs_loadall(metadataFile, 0);
 	}
 
 	bool autoRecovery = fs_can_do_auto_recovery();
--- a/src/master/matoclserv.cc
+++ b/src/master/matoclserv.cc
@@ -674,6 +674,7 @@ void matoclserv_store_sessions() {
 	}
 }
 
+#define MFSSIGNATURE "MFS"
 int matoclserv_load_sessions() {
 	session *asesdata;
 	uint32_t ileng;
@@ -685,7 +686,6 @@ int matoclserv_load_sessions() {
 	uint32_t i,statsinfile;
 	int r;
 	FILE *fd;
-	std::string legacySessionSignature = std::string("MFS");
 
 	fd = fopen(kSessionsFilename, "r");
 	if (fd==NULL) {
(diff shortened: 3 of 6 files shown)


src/master/matoclserv.cc
```diff
@@ -700,19 +701,23 @@ int matoclserv_load_sessions() {
 		fclose(fd);
 		return -1;
 	}
-	if (memcmp(hdr,SFSSIGNATURE "S 1.5",8)==0) {
+	if (memcmp(hdr,SFSSIGNATURE "S 1.5",8)==0 ||
+	    memcmp(hdr, (legacySessionSignature + "S 1.5").c_str(), 8) == 0) {
```
lgsilva3087, line 705: Please, consider to use the same style as SFSSIGNATURE, I mean a macro MFSSIGNATURE, to use the same style until both are modernized.

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Exit mount point and save metadata
+cd "${TEMP_DIR}"
+assert_success saunafs_master_daemon stop
+
+# Expected Values
+assert_equals "SFSM 2.9" "$(head -c8 ${metadata_file})"
+assert_equals "[SFS EOF MARKER]" "$(tail -c16 ${metadata_file})"
+
+# Modify metadata signature for legacy and reload metadata by staring master
```
lgsilva3087, line 29: Posible typo in starting

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
```diff
@@ -0,0 +1,46 @@
...
+echo -n "LIZ" | dd of=${metadata_file} bs=1 seek=0 count=3 conv=notrunc
+echo -n "MFS" | dd of=${metadata_file} bs=1 seek=$(( $(stat --print="%s" ${metadata_file}) - 15)) count=3 conv=notrunc
+assert_equals "LIZM 2.9" "$(head -c8 ${metadata_file})"
+assert_equals "[MFS EOF MARKER]" "$(tail -c16 ${metadata_file})"
+
+mv -iv "${info[master_data_path]}/metadata.sfs" "${info[master_data_path]}/metadata.mfs"
+
+# Start master server to reload metadata legacy format
+assertlocal_success saunafs_master_daemon start
+# Restart master server to load metadata in new format when both are present
```
lgsilva3087, line 39: Does It means that metadata will be always converted to new format and saved to metadata.sfs?
Why there are two metadatas here with different formats?

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
```diff
@@ -0,0 +1,46 @@
+timeout_set 30 seconds
+master_cfg+="MAGIC_DEBUG_LOG = $TEMP_DIR/syslog|LOG_FLUSH_ON=DEBUG"
+touch "$TEMP_DIR/syslog"
+
+CHUNKSERVERS=0 \
+	MASTERSERVERS=1 \
```
lgsilva3087, line 6: Please consider adding a shadow and a metalogger for the final version, to check that everything is compatible.

lgsilva3087 (commented):
Nice work. Please see my suggestions and doubts.

ralcolea (approved):

src/master/filesystem.cc
```diff
@@ -218,7 +218,24 @@ int fs_loadall(void) {
 		throw MetadataFsConsistencyException(
 				"temporary metadata file (" + std::string(kMetadataTmpFilename) + ") exists, metadata directory is in dirty state");
 	}
-	if ((metadataserver::isMaster()) && !fs::exists(kMetadataFilename)) {
+	std::string metadataFile = std::string("");
+	if (fs::exists(kMetadataFilename)) {
```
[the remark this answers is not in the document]
Antuan-AE: This checking even when might look redundant at first, enforce priority for existing actual metadata file over the possibility that there might also be an old legacy metadata file that was left hanging.

src/master/filesystem.cc
```diff
@@ -218,7 +218,24 @@ int fs_loadall(void) {
...
+		metadataFile = kMetadataFilename;
+	}
+	if(fs::exists(kMetadataFilename) && fs::exists(kMetadataLegacyFilename)) {
+		metadataFile = kMetadataFilename;
```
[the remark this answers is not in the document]
Antuan-AE: You are right, this assignment is redundant. I did it intentionally to enforce that if both files are present, system will load actual metadata file and ignore legacy metadata.

src/common/metadata.cc
```diff
@@ -92,9 +95,15 @@ uint64_t metadataGetVersion(const std::string& file) {
 	std::string signature = std::string(chkbuff, 8);
 	std::string sfsSignature = std::string(SFSSIGNATURE "M 2.9");
 	std::string sauSignature = std::string(SAUSIGNATURE "M 2.9");
+	std::string legacySignature = std::string("LIZM 2.9");
 
 	if (signature == sfsSignature || signature == sauSignature) {
 		memcpy(eofmark,"[SFS EOF MARKER]",16);
+	} else if (memcmp(chkbuff, legacySignature.data(), legacySignature.size()) == 0) {
```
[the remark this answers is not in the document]
Antuan-AE: applied

src/common/metadata.cc
```diff
@@ -92,9 +95,15 @@ uint64_t metadataGetVersion(const std::string& file) {
 	std::string signature = std::string(chkbuff, 8);
 	std::string sfsSignature = std::string(SFSSIGNATURE "M 2.9");
 	std::string sauSignature = std::string(SAUSIGNATURE "M 2.9");
+	std::string legacySignature = std::string("LIZM 2.9");
```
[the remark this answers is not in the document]
Antuan-AE: This is intended to be a temporary feature to allow easy migration. Its not intended to stay for long time. This is why i consider to scope this change only to the part of the code that is using it rather that expose it to global `SFSCommunication.h` level. Also use of legacy is intended not to reference explicitly the brand (company) from which we are supporting the migration at this point.

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
Antuan-AE: fixed

src/master/filesystem.cc
```diff
@@ -218,7 +218,24 @@ int fs_loadall(void) {
...
+				kMetadataFilename, kMetadataLegacyFilename, kMetadataLegacyFilename);
+	}
+	if (!fs::exists(kMetadataFilename) && fs::exists(kMetadataLegacyFilename)) {
+		metadataFile = kMetadataLegacyFilename;
+		safs_pretty_syslog(LOG_WARNING, "Only Legacy metadata file %s found and will be loaded instead. "
+		                   "You should delete legacy metadata %s on next restart after new metadata %s is created ",
+				metadataFile.c_str(), kMetadataLegacyFilename, kMetadataFilename);
+	}
+	if (metadataserver::isMaster() &&
+	    (!fs::exists(kMetadataFilename) && !fs::exists(kMetadataLegacyFilename))) {
```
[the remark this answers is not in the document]
Antuan-AE: applied

src/master/filesystem.cc
```diff
@@ -218,7 +218,24 @@ int fs_loadall(void) {
...
+		safs_pretty_syslog(LOG_WARNING, "There are two metadata files in the data path: %s and %s. "
+		                   "Please remove the legacy one (%s) to avoid damage to your storage.",
+				kMetadataFilename, kMetadataLegacyFilename, kMetadataLegacyFilename);
+	}
+	if (!fs::exists(kMetadataFilename) && fs::exists(kMetadataLegacyFilename)) {
+		metadataFile = kMetadataLegacyFilename;
+		safs_pretty_syslog(LOG_WARNING, "Only Legacy metadata file %s found and will be loaded instead. "
+		                   "You should delete legacy metadata %s on next restart after new metadata %s is created ",
+				metadataFile.c_str(), kMetadataLegacyFilename, kMetadataFilename);
+	}
```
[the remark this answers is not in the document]
Antuan-AE: applied

src/master/filesystem.cc
```diff
@@ -218,7 +218,24 @@ int fs_loadall(void) {
...
+	std::string metadataFile = std::string("");
+	if (fs::exists(kMetadataFilename)) {
+		metadataFile = kMetadataFilename;
+	}
+	if(fs::exists(kMetadataFilename) && fs::exists(kMetadataLegacyFilename)) {
+		metadataFile = kMetadataFilename;
+		safs_pretty_syslog(LOG_WARNING, "There are two metadata files in the data path: %s and %s. "
+		                   "Please remove the legacy one (%s) to avoid damage to your storage.",
+				kMetadataFilename, kMetadataLegacyFilename, kMetadataLegacyFilename);
+	}
```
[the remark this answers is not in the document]
Antuan-AE: applied

src/common/metadata.cc
```diff
@@ -92,9 +95,15 @@ uint64_t metadataGetVersion(const std::string& file) {
...
 
 	if (signature == sfsSignature || signature == sauSignature) {
 		memcpy(eofmark,"[SFS EOF MARKER]",16);
+	} else if (memcmp(chkbuff, legacySignature.data(), legacySignature.size()) == 0) {
+		safs_pretty_syslog(LOG_WARNING,
+		                   "Legacy metadata section header %s, was detected in the metadata file %s",
```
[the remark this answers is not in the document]
Antuan-AE: applied

src/master/filesystem.cc
```diff
@@ -218,7 +218,24 @@ int fs_loadall(void) {
 		throw MetadataFsConsistencyException(
 				"temporary metadata file (" + std::string(kMetadataTmpFilename) + ") exists, metadata directory is in dirty state");
 	}
-	if ((metadataserver::isMaster()) && !fs::exists(kMetadataFilename)) {
+	std::string metadataFile = std::string("");
```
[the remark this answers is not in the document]
Antuan-AE: applied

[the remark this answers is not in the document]
ralcolea: I agree, but the logic it's the same as initializing with empty file. Anyways, the checks are evaluated and the final result is the same.

src/master/matoclserv.cc
Antuan-AE: I remember our last discussion about code style was, because we never find time to apply a full code refactor we would try to introduce new code using latest style even if it means having mixed styles in some part of the code. This is why i avoid to use c-style macros and use std::string type. I also wanted to scope the changes only to the part of the code that is using it rather that expose it to global `SFSCommunication.h` because this feature is intended to be temporal and will probably be deprecated after we drop support for migration at some point.

src/master/filesystem.cc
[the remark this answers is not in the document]
ralcolea: My point is that first check is not need, because the next checks already guarantee that the right scenario will be handled and the result is the same.

Anyways, the logic will work OK. These are only suggestions :)

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
Antuan-AE: Yes, so the logic is like this. If only `metadata.mfs` is found, it loads it, and on the next metadata dump the new `metadata.sfs` is created. This means at this point both files will be present until system admin manually deletes the legacy `metadata.mfs`. The system detect if both files are present and loads the actual `metadata.sfs` in this case. Also when SaunaFS shadows connect to legacy master cluster, it downloads the metadata content to the file called `metadata.sfs` but with wrong tags internally until next dump occurs, then internal tags are properly written to file. This how ever was not yet tested in this integration tests, because it requires the testing framework to support mixing actual and legacy binaries and goes beyond the scope of this PR.

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
Antuan-AE: To properly test integration with Shadow and Metaloggers we need to mixed current binaries to legacy binaries and the testing framework doesn't support it currently. So this tests will be added as part of another PR in the future.

src/master/matoclserv.cc
lgsilva3087: I agree that this is temporal. But, please notice that you are still doing the same, using memcmp and converting the string to an array to compare it. The lines are just longer.

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
lgsilva3087: Clear now.

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
lgsilva3087: Not exactly, by adding another shadow and a metalogger you will tests that this versions does not have compatibility issues even if the metadata is mfs or sfs. No need for legacy versions at this point, you are cracking the metadata on purpose.

lgsilva3087 (approved):
Please see my last suggestions.

src/master/matoclserv.cc
Antuan-AE: applied

tests/test_suites/ShortSystemTests/test_metadata_load_legacy.sh
Antuan-AE: Not exactly, because the data flow goes from Shadows/Metalogger to Master, not the other way arround, So even when cracking the metadata on purpose, Only master will notice when loading it. Shadow and Metalogger will never know. Since Shadows will be downloading from master the already corrected metadata after master initialized. In any case i have added a Shadow and Metalogger to the test

PR #145, opened 2024-07-19
feat(mount): Add readcachemaxsize option

rolysr:
For allowing the client process to limit the expected memory in system for reading operations, an option was added so that the client also waits for availability of system memory, avoiding the client process to crash.


src/common/system_memory.h
```diff
@@ -0,0 +1,61 @@
...
+    GlobalMemoryStatusEx(&memInfo);
+    return memInfo.ullTotalPhys;
+#elif __linux__
+    struct sysinfo memInfo;
+    sysinfo(&memInfo);
+    return (memInfo.totalram * memInfo.mem_unit);
+#else
+    return 0; // Unsupported platform
+#endif
+}
```
uristdwarf, line 61: Maybe create an [interface](https://stackoverflow.com/questions/1216750/how-can-i-simulate-interfaces-in-c) for these functions, so that both windows/linux client (and maybe in the future other clients) can implement this how they like, instead of using ifdefs

rolysr: Done, please check the interface implementation on src/mount/memory_info.h and src/mount/linux_memory_info.h. Review if it is ok that way.

src/mount/linux_memory_info.h
```diff
@@ -0,0 +1,43 @@
...
+        sysinfo(&memInfo);
+        return (memInfo.freeram * memInfo.mem_unit);
+    }
+
+    unsigned long long getTotalMemory() override {
+        struct sysinfo memInfo;
+        sysinfo(&memInfo);
+        return (memInfo.totalram * memInfo.mem_unit);
+    }
+};
```
dmga44, line 43: Please remember adding an extra line ending here.

src/mount/readdata_cache.h
```diff
@@ -286,8 +286,8 @@ class ReadCache {
 	 * \return cache query result
 	 */
 	void query(Offset offset, Size size, ReadCache::Result &result,
-	           bool insertPending = true) {
-		collectGarbage();
+	           uint32_t usedReadCacheSize, bool insertPending = true) {
```
dmga44, line 289: I think your idea was to modify the value of gUsedReadCacheMemory considering the changes of this variable through collectGarbage, maybe you somehow need to pass it as a reference?

src/mount/linux_memory_info.h
rolysr: Done

src/mount/readdata_cache.h
rolysr: Done

src/mount/readdata_cache.h
```diff
@@ -381,11 +382,12 @@ class ReadCache {
 		return entries_.insert(it, *e);
 	}
 
-	void collectGarbage(unsigned count = 4) {
+	void collectGarbage(uint32_t usedReadCacheMemoryMb, unsigned count = 4) {
```
dmga44, line 385: Same comment as before here, were you not trying to update the value of gUsedReadCacheMemory through the changes of usedReadCacheSize in query function which comes down here?

Also, is the value stored in usedReadCacheMemoryMb representing megabytes? Because there is a substraction a bit down here which is substracting bytes (the buffer size).

rolysr: Done

src/mount/sauna_client.h
```diff
@@ -73,6 +73,7 @@ struct FsInitParams {
 	static constexpr unsigned kDefaultChunkserverTotalReadTo = 2000;
 	static constexpr unsigned kDefaultCacheExpirationTime = 1000;
 	static constexpr unsigned kDefaultReadaheadMaxWindowSize = 65536;
+	static constexpr unsigned kDefaultReadCacheMaxSize = 16384;
```
lgsilva3087, line 76: Make it consistent with documentation.

lgsilva3087 (commented):
Please see my suggestions.

src/mount/memory_info.h
```diff
@@ -0,0 +1,70 @@
...
+        GlobalMemoryStatusEx(&memInfo);
+        return memInfo.ullTotalPhys;
+    }
+};
+#elif __linux__
+class LinuxMemoryInfo : public IMemoryInfo {
+public:
+    unsigned long long getAvailableMemory() override {
+        struct sysinfo memInfo;
+        sysinfo(&memInfo);
```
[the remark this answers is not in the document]
lgsilva3087: Looks like not. Let's try with this.

src/mount/readdata.cc
```diff
@@ -583,6 +656,18 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
+
+			uint32_t read_cache_bytes_to_reserve =
+				getBytesToBeReadFromCS(reader.index(), offset_in_chunk,
+										size_in_chunk, reader.fileLength());
+			std::unique_lock lock(gReadCacheMemoryMutex);
+			bool successfulWait =
+			    waitForEnoughSystemMemory(read_cache_bytes_to_reserve);
+			if (!successfulWait) {
+				break;
+			}
```
lgsilva3087, line 668: After the break, SAUNAFS_STATUS_OK will be returned, which is not correct.
Following the rest of the function logic, looks like we need to throw an exception, and then the logic of retries is executed.
According to this, we need to evaluate if the new retry logic in the waitForEnoughSystemMemory is really needed.
Sorry, I did not saw it in my first review.

src/mount/readdata.cc
```diff
@@ -57,6 +58,57 @@
...
+			bytesToReadLeft > avalaibleSystemMemory;
+}
+
+void increaseUsedReadCacheMemory(size_t bytesToReadLeft) {
+	std::unique_lock lock(gUsedReadCacheMemoryMutex);
+	gUsedReadCacheMemory += bytesToReadLeft;
+}
+
+void waitForEnoughSystemMemory(size_t bytesToReadLeft) {
+	while (readShouldWaitForSystemMemory(bytesToReadLeft)) {
```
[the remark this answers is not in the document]
lgsilva3087: Please see my last comment abour the complete function logic and the retries.

lgsilva3087 (commented):
Almost ready. Please see my comment.

src/mount/readdata.cc
rolysr: Done, integrated corresponding wait result with retry logic already implemented in code.

src/mount/readdata.cc
```diff
@@ -56,6 +57,60 @@
...
+bool waitForEnoughSystemMemory(size_t bytesToReadLeft) {
+	bool readShouldWait = false;
+	if(readShouldWaitForSystemMemory(bytesToReadLeft)) {
+		safs_pretty_errlog(
+		    LOG_ERR,
+		    "Not enough system memory available for reading");
+		readShouldWait = true;
+	}
+	return readShouldWait;
+}
```
lgsilva3087, line 96: Looks like this function is probably not needed anymore. Seems that it was used to log the error, but now you throw the exception in the function call.

lgsilva3087 (approved):
Please see my last comment.

doc/sfsmount.1.adoc
```diff
@@ -162,6 +162,9 @@ disables cache (default: 0).
 *-o readaheadmaxwindowsize=*'KB'::
 Set max value of readahead window per single descriptor in kibibytes (default: 16384).
 
+*-o readcachemaxsize=*'MB'::
```
aNeutrino, line 165: please use camelcase in our documentation.... everywhere.
```suggestion
*-o rearCacheMaxSize=*'MB'::
```

lgsilva3087: Probably this change should be in a separate PR changing all arguments.

src/mount/readdata.cc

doc/sfsmount.1.adoc
rolysr: We better let this for a dedicated PR improving sfsmount doc readability

dmga44 (approved):
LGTM.

PR #148, opened 2024-07-20
fix: Release compilation on GCC 14

uristdwarf:
When building with "Release" CMAKE_BUILD_TYPE, the compilation fails on Arch Linux with GCC 14.1.1, complaining about uninitialized array's.

This small commit fixes that and two linter errors.


Antuan-AE (approved):
LGTM

PR #151, opened 2024-07-21
fix: Update services to wait for network

aNeutrino:
network.target is not giving warranty that IP addresses are set
network-online.target gives such assurances
https://systemd.io/NETWORK_ONLINE/

Co-authored-by: Antuan <antuan@leil.io>
Co-authored-by: Crash <crash@leil.io>


uristdwarf (approved):

lgsilva3087 (approved):
LGTM

PR #152, opened 2024-07-22
feat(metrics): Add rest of master chartdata metrics

uristdwarf:
This commit finishes adding metric data from master chartsdata. Since all the chartsdata was counters in Prometheus terminology, there was no need to add other types, however a a small refactor was needed to accomodate master metrics and in the future other service metrics. Notably removed was CPU usage and memory, since there are better ways to gather this data, including with a prometheus server itself (which will scrape this data).

This commit deprecates master chartsdata, and is the first step in deprecating the CGI monitoring

### DIFF
--- a/src/master/chartsdata.cc
+++ b/src/master/chartsdata.cc
@@ -18,6 +18,9 @@
    along with SaunaFS  If not, see <http://www.gnu.org/licenses/>.
  */
 
+
+/* DEPRECATED (master/chartsdata.h): Use src/metrics instead */
+
 #include "common/platform.h"
 
 #include "master/chartsdata.h"
--- a/src/master/matoclserv.cc
+++ b/src/master/matoclserv.cc
@@ -5268,6 +5268,7 @@ void matoclserv_read(matoclserventry *eptr) {
 		}
 		eptr->inputpacket.startptr+=i;
 		eptr->inputpacket.bytesleft-=i;
+		metrics::Counter::increment(metrics::Counter::Master::CLIENT_RX_BYTES, i);
 		stats_brcvd+=i;
 
 		if (eptr->inputpacket.bytesleft>0) {
@@ -5303,7 +5304,7 @@ void matoclserv_read(matoclserventry *eptr) {
 			eptr->inputpacket.startptr = eptr->hdrbuff;
 			matoclserv_gotpacket(eptr,type,eptr->inputpacket.packet,size);
 			stats_prcvd++;
-			metrics::Counter::increment(metrics::Counter::CLIENT_RX_PACKETS);
+			metrics::Counter::increment(metrics::Counter::Master::CLIENT_RX_PACKETS);
 
 			if (eptr->inputpacket.packet) {
 				free(eptr->inputpacket.packet);
@@ -5339,13 +5340,14 @@ void matoclserv_write(matoclserventry *eptr) {
 		}
 		pack->startptr+=i;
 		pack->bytesleft-=i;
+		metrics::Counter::increment(metrics::Counter::Master::CLIENT_TX_BYTES, i);
 		stats_bsent+=i;
 		if (pack->bytesleft>0) {
 			return;
 		}
 		free(pack->packet);
 		stats_psent++;
-		metrics::Counter::increment(metrics::Counter::CLIENT_TX_PACKETS);
+		metrics::Counter::increment(metrics::Counter::Master::CLIENT_TX_PACKETS);
 		eptr->outputhead = pack->next;
 		if (eptr->outputhead==NULL) {
 			eptr->outputtail = &(eptr->outputhead);
--- a/src/metrics/master.cc
+++ b/src/metrics/master.cc
@@ -1,0 +1,187 @@
+/*
+
+   Copyright 2024 Leil Storage OÜ
+
+   This file is part of SaunaFS.
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "metrics/master.h"
+#include "metrics.h"
+#include "metrics/utils.h"
+
+#ifdef HAVE_PROMETHEUS
+using namespace metrics;
+
+Master::Master(std::shared_ptr<prometheus::Registry> &registry) {
+	// clang-format off
+	// NOLINTBEGIN(cppcoreguidelines-prefer-member-initializer)
+	packetClientCounter = &setupFamily(
+		"metadata_observed_packets_client_total",
+		"Number of observed packets from and for client", registry);
+	byteClientCounter = &setupFamily(
+		"metadata_observed_bytes_client_total",
+		"Number of observed bytes from and for client", registry);
+	filesystemCounter = &setupFamily(
+		"metadata_stats_total",
+		"Number of observed filesystem operations", registry);
+	chunkCounter= &setupFamily(
+		"metadata_chunk_operations_total",
+		"Number of chunk operations", registry);
+	// NOLINTEND(cppcoreguidelines-prefer-member-initializer)
+
+	// A very hacky way to allow compile time checking if metrics have been
+	// set. Any enum value not used will throw a compile time error.
+	using Master = Counter::Master;
+	Master start = Master::KEY_START;
+	switch (start) {
+		case Master::KEY_START:
+			[[fallthrough]];
+		case Master::CHUNK_DELETE:
+			masterCounters[static_cast<unsigned int>(Master::CHUNK_DELETE)] =
+				Counter(
+					{{"chunk", "operations"}, {"operation", "delete"}},
+					chunkCounter);
+			[[fallthrough]];
+		case Master::CHUNK_REPLICATE:
+			masterCounters[static_cast<unsigned int>(Master::CHUNK_REPLICATE)] =
+				Counter(
+					{{"chunk", "operations"}, {"operation", "replicate"}},
+					chunkCounter);
+			[[fallthrough]];
+		case Master::FS_STATFS:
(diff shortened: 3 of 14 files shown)


lgsilva3087 (dismissed):
LGTM.
I see you are almost not using anymore #include "common/platform.h". I don't remember the last decision about it, can you please remaind it.

src/metrics/utils.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
rolysr, line 1: Consider adding corresponding copyright license to new created files.

src/master/chartsdata.cc
```diff
@@ -18,6 +18,9 @@
    along with SaunaFS  If not, see <http://www.gnu.org/licenses/>.
  */
 
+
+/* DEPRECATED: Use src/metrics instead */
```
ralcolea, line 22: I think this comment is somehow ambiguous. 
If the comment is for file `master/charsdata.h`, the comment should be placed beside the file which is being deprecated. Otherwise, it should be explicitely mentioned the name of file(s) that are being deprecated.

src/metrics/master.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include "metrics/metrics.h"
+
+#ifdef HAVE_PROMETHEUS
+#include <prometheus/counter.h>
+#include <prometheus/registry.h>
+
+namespace metrics::master {
+
+struct Master {
```
ralcolea, line 10: Why use `struct` instead of `class`?

src/metrics/master.h
```diff
@@ -0,0 +1,25 @@
...
+public:
+	Master() = default;
+	Master(std::shared_ptr<prometheus::Registry>& registry);
+
+	// Metric(s)
+	prometheus::Family<prometheus::Counter> *packet_client_counter{nullptr};
```
ralcolea, line 16: Maybe `prometheus::Family<prometheus::Counter>` can be defined as a custom data type with a shorter name by using C++ `using ` instruction and then use this data type in all functions.

```cpp
using MetricsCounter = prometheus::Family<prometheus::Counter>;

// Metric(s)
MetricsCounter *packet_client_counter{nullptr};
MetricsCounter *byte_client_counter{nullptr};
MetricsCounter *filesystem_counter{nullptr};
MetricsCounter *chunk_counter{nullptr};
```

src/metrics/master.cc
```diff
@@ -0,0 +1,166 @@
...
+		    "metadata_chunk_operations_total",
+		    "Number of chunk operations", registry);
+		// NOLINTEND(cppcoreguidelines-prefer-member-initializer)
+
+		// clang-format off
+		// A very hacky way to allow compile time checking if metrics have been
+		// set. Any enum value not used will throw a compile time error.
+		master::Counters start = KEY_START;
+		switch (start) {
+        case KEY_START:
```
ralcolea, line 28: Why the alignment is different for `case KEY_START:` ?

src/metrics/master.cc
```diff
@@ -0,0 +1,166 @@
...
+				{{"filesystem", "operations"}, {"operation", "OPEN"}},
+			    filesystem_counter);
+			[[fallthrough]];
+		case FS_READ:
+			master_counters[FS_READ] =
+			    Counter(
+				{{"filesystem", "operations"}, {"operation", "READ"}},
+			    filesystem_counter);
+			[[fallthrough]];
+        case FS_WRITE:
```
ralcolea, line 131: Same here about different alignment. Please check the rest of the function.

src/metrics/metrics.cc
```diff
@@ -64,56 +81,32 @@ prometheus::Family<prometheus::Counter> &setup_family(
...
 	std::shared_ptr<prometheus::Registry> registry;
-
-	// Metric(s)
-	prometheus::Family<prometheus::Counter> *packet_counter;
-
-	// All counters
-	std::array<Counter, Counter::Key::KEY_END + 1> counters;
 };
-Counters counters;  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+PrometheusMetrics prom_metrics;  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
```
ralcolea, line 103: Why not `prometheus_metrics` instead of `prom_metrics`? 
I think is more clear.

src/metrics/metrics.cc
```diff
@@ -34,8 +34,25 @@
 #endif
 
 #include "metrics.h"
+#include "metrics/master.h"
 #include "slogger/slogger.h"
 
+// Quick summary on the current implementation from 17/07/24
```
ralcolea, line 40: I think in the final version, there should not be implementation details referencing specific dates.

ralcolea (dismissed):
Very good job @uristdwarf :+1: :fire: :rocket: 

I added some minor suggestions to improve readability and other aspects, but they are not stoppers for merging.

uristdwarf:
> LGTM. I see you are almost not using anymore #include "common/platform.h". I don't remember the last decision about it, can you please remaind it.

For new libraries, I think we are not adding it but for existing/refactored ones we are

src/metrics/master.h
uristdwarf: struct is public by default compared to class, which I generally prefer if I have few/no members that are private and no inheritance is intended for it (i.e it's a simple data structure).

src/metrics/metrics.cc
uristdwarf: My worry is that comments go out of date fast, but I guess that one can look at git blame instead.

ralcolea (dismissed):
LGTM :+1: :fire: :rocket: 
The only suggestion I have before merging (if not done) is to run `test_prometheus.sh` with `ENABLE_PROMETHEUS = 0` to confirm everything works as expected.

lgsilva3087 (changes requested):
Requesting changes about bytes sent and received. The call to increment should use the second argument (by default it is 1).

lgsilva3087:
> > LGTM. I see you are almost not using anymore #include "common/platform.h". I don't remember the last decision about it, can you please remaind it.
> 
> For new libraries, I think we are not adding it but for existing/refactored ones we are

Maybe this kind of decisions should be communicated to all devs.

src/master/matoclserv.cc
```diff
@@ -5233,6 +5233,7 @@ void matoclserv_read(matoclserventry *eptr) {
 		}
 		eptr->inputpacket.startptr+=i;
 		eptr->inputpacket.bytesleft-=i;
+		metrics::Counter::increment(metrics::master::CLIENT_RX_BYTES);
```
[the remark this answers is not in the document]
uristdwarf: Nice catch, fixed

src/master/matoclserv.cc
```diff
@@ -5304,13 +5305,14 @@ void matoclserv_write(matoclserventry *eptr) {
 		}
 		pack->startptr+=i;
 		pack->bytesleft-=i;
+		metrics::Counter::increment(metrics::master::CLIENT_TX_BYTES);
```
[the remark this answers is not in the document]
uristdwarf: Also true

src/metrics/master.h
```diff
@@ -0,0 +1,43 @@
...
+
+struct Master {
+	Master() = default;
+	Master(std::shared_ptr<prometheus::Registry>& registry);
+
+	// Metric(s)
+	CounterFamily *packet_client_counter{nullptr};
+	CounterFamily *byte_client_counter{nullptr};
+	CounterFamily *filesystem_counter{nullptr};
+	CounterFamily *chunk_counter{nullptr};
```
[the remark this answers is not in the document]
uristdwarf: Done. I think it's mostly because C++ library functions are snake_case, and I get confused sometimes

src/metrics/metrics.h
```diff
@@ -27,35 +27,62 @@
...
+
+using CounterFamily = prometheus::Family<prometheus::Counter>;
+
 #endif
 
 namespace metrics {
 
+namespace master {
+
+enum Counters {
```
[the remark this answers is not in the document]
uristdwarf: This was more of a pain than I realized, as I used implicit conversion a lot more than expected. Still, I think it's better now because it prevents programmer error from implicitly passing a non-enum to the interface functions.

src/metrics/master.cc
```diff
@@ -0,0 +1,185 @@
...
+		"Number of observed filesystem operations", registry);
+	chunk_counter= &setup_family(
+		"metadata_chunk_operations_total",
+		"Number of chunk operations", registry);
+	// NOLINTEND(cppcoreguidelines-prefer-member-initializer)
+
+	// A very hacky way to allow compile time checking if metrics have been
+	// set. Any enum value not used will throw a compile time error.
+	master::Counters start = KEY_START;
+	switch (start) {
```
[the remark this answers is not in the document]
uristdwarf: Not sure how. My idea is that adding enum members will raise compile warnings here, so you don't forget about this place (and try to figure why your metric isn't working)

tests/test_suites/SanityChecks/test_prometheus.sh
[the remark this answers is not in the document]
uristdwarf: I've for now only extended them to the operations in metadata, because I'm unsure about the other operations. I've checked the CGI test and all it does is check if the data is populated.

lgsilva3087 (dismissed):
LGTM

tests/test_suites/ShortSystemTests/test_prometheus_metadata.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Check if any value is zero
+input_data=$(grep -E 'metadata_stats_total\{filesystem="operations",operation="[A-Z]+"\}' <<< "$(curl --fail "${prometheus_host}/metrics")")
+echo "$input_data" | while read -r line; do
+	echo "${line}"
+	# BUG: Setting symlinkcachetimeout to 0, it still uses client cache
+	if grep -q "READLINK" <<< "$line"; then
+		continue
+	fi
+    # Extract the numeric value from each line
```
ralcolea, line 32: Please, check the indentation of lines 32-35. These lines uses spaces instead of tabs.

ralcolea (dismissed):
Great job @uristdwarf :+1: :fire: :rocket:
I shared some minor suggestions for your consideration, but none of them are stoppers to merge.

However, I had compilation errors when building the project without installing `prometheus-cpp-dev` package. Can you confirm whether this scenario works as expected?

ralcolea (dismissed):
Great job @uristdwarf :+1: :fire: :rocket:

lgsilva3087 (approved):

lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM :clap:

PR #153, opened 2024-07-23
Fix service uraft restart on failure

Antuan-AE:
Current saunafs-uarft binary is deploying a service file `saunafs-uraft.service` that if service fails, it doesn't allow to be automatically reloaded.


ralcolea (approved):
LGTM :+1: :fire: :rocket:

lgsilva3087 (approved):
LGTM.

PR #154, opened 2024-07-29
fix(mount): Fix IIS permissions mess

dmga44:
This PR aims to provide the expected behavior of our system while
dealing with the IIS tool. This tool apparently uses a very specific
user (17) to access the underlaying files. The user 17 was not expected
to perform these kinds of operations, thus causing the permission denied
errors. Solution was to make a breach at some checks to allow it.

### DIFF
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -233,6 +233,7 @@ uint8_t get_session_flags() { return session_flags; }
 bool isMasterDisconnected() { return gIsDisconnectedFromMaster.load(); }
 
 #define LOCAL_USERS_THRESHOLD 0x30000
+#define IIS_USER 17
 static int32_t mounting_uid = USE_LOCAL_ID;
 static int32_t mounting_gid = USE_LOCAL_ID;
 uint32_t winfsp_uid = 0;
@@ -242,13 +243,13 @@ std::mutex winfsp_context_mutex;
 void update_last_winfsp_context(const uint32_t uid, const uint32_t gid) {
 	std::lock_guard<std::mutex> winfsp_context_lock(winfsp_context_mutex);
 
-	if (uid > LOCAL_USERS_THRESHOLD) {
+	if (uid > LOCAL_USERS_THRESHOLD || uid == IIS_USER) {
 		if (debug_mode && winfsp_uid != uid) {
 			safs::log_debug("set WinFSP user id to {}", uid);
 		}
 		winfsp_uid = uid;
 	}
-	if (gid > LOCAL_USERS_THRESHOLD) {
+	if (gid > LOCAL_USERS_THRESHOLD || gid == IIS_USER) {
 		if (debug_mode && winfsp_gid != gid) {
 			safs::log_debug("set WinFSP group id to {}", uid);
 		}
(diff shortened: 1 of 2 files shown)


rolysr (approved):
LGTM

src/mount/sauna_client.cc
```diff
@@ -233,6 +233,7 @@ uint8_t get_session_flags() { return session_flags; }
 bool isMasterDisconnected() { return gIsDisconnectedFromMaster.load(); }
 
 #define LOCAL_USERS_THRESHOLD 0x30000
+#define IIS_USER 17
```
Antuan-AE, line 236: Are we sure IIS is always using this hard coded `17` value. Consider making this a parameter that could be tweak on the  fly.

Antuan-AE (approved):
LGTM

src/mount/sauna_client.cc
ralcolea, line 236: I suggest to use a more generic name, like `ALLOWED_USERS` to use this feature not only for IIS in Tsukaeru, but for other customers who could have the same issue.

I agree with @aNeutrino that if we use a list of users, is even more flexible and generic the solution.
A definition like
`std::unordered_set<int> ALLOWED_USERS;`
can do the job.

src/mount/sauna_client.cc
```diff
@@ -242,13 +243,13 @@ std::mutex winfsp_context_mutex;
 void update_last_winfsp_context(const uint32_t uid, const uint32_t gid) {
 	std::lock_guard<std::mutex> winfsp_context_lock(winfsp_context_mutex);
 
-	if (uid > LOCAL_USERS_THRESHOLD) {
+	if (uid > LOCAL_USERS_THRESHOLD || uid == IIS_USER) {
```
ralcolea, line 246: With the new suggestion, the condition can be updated:
```cpp
 if (uid > LOCAL_USERS_THRESHOLD || ALLOWED_USERS.contains(uid)) {
   ...
 }
```

src/mount/sauna_client.cc
```diff
@@ -242,13 +243,13 @@ std::mutex winfsp_context_mutex;
...
 		if (debug_mode && winfsp_uid != uid) {
 			safs::log_debug("set WinFSP user id to {}", uid);
 		}
 		winfsp_uid = uid;
 	}
-	if (gid > LOCAL_USERS_THRESHOLD) {
+	if (gid > LOCAL_USERS_THRESHOLD || gid == IIS_USER) {
```
ralcolea, line 252: Similar change here, but using `gid`.

src/mount/sauna_client.cc
ralcolea: I agree with @Antuan-AE.
It would be very nice to support updating this value at the configuration level.

Is it possible to define an option for the client?
For example:
`-o allowedUsers=17,123,456`

ralcolea (commented):
I think the idea is very good @dmga44.
Please, see my suggestions.

aNeutrino:
this PR must be rebased - because now it include duplicated commits from release 4.4.0

dmga44:
Well, I'm gonna implement this feature like ```-o allowedUsers=17:123:456```.

At first I was going to have it like @Antuan-AE suggested, but the way suggested by @ralcolea seems a bit more generic for future patch ups, though I think we could not have it exactly the way he proposed because the commas are used to separate command line options in our clients. In other words, the commas are used in ```-o readWorkers=10,writeWorkers=15``` to separate the options, so we could not used it inside the same option value. Beside this change, I'll implement almost exactly the rest of the suggestion.

src/mount/sauna_client.cc
```diff
@@ -267,12 +268,12 @@ void convert_winfsp_context_to_master_context(uint32_t &uid, uint32_t &gid) {
...
-	     (st.st_mode & S_IWOTH) ||
-	     (st.st_gid ==
-	          (mounting_gid == USE_LOCAL_ID ? winfsp_gid : mounting_gid) &&
-	      (st.st_mode & S_IWGRP)))) {
+	if ((st.st_uid ==
+	     (mounting_uid == USE_LOCAL_ID ? winfsp_uid : mounting_uid)) ||
+	    (st.st_mode & S_IWOTH) ||
+	    (st.st_gid ==
+	         (mounting_gid == USE_LOCAL_ID ? winfsp_gid : mounting_gid) &&
+	     (st.st_mode & S_IWGRP))) {
```
dmga44, line 276: This is not necessary for the PR (it is just removing an extra pair of parenthesis), but it is not so unrelated.

ralcolea: I noticed :sweat_smile:

ralcolea (approved):
Great job @dmga44 :+1: :fire: :rocket:

PR #155, opened 2024-07-30
feat(admin): Add list-disk-groups command

lgsilva3087:
The saunafs-admin was extended to show the disk groups configuration from the chunkservers. The default disk manager does not support this feature yet and prints 'Not supported', but it is already available (and used) from the plugins by reimplementing the default behavior.

### DIFF
--- a/src/admin/list_disk_groups_command.cc
+++ b/src/admin/list_disk_groups_command.cc
@@ -1,0 +1,98 @@
+/*
+   Copyright 2023 Leil Storage OÜ
+
+   This file is part of SaunaFS.
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "common/platform.h"
+
+#include <yaml-cpp/yaml.h>
+#include <iostream>
+
+#include "admin/list_chunkservers_command.h"
+#include "admin/list_disk_groups_command.h"
+#include "common/saunafs_version.h"
+#include "common/serialization.h"
+#include "common/server_connection.h"
+
+std::string ListDiskGroupsCommand::name() const { return "list-disk-groups"; }
+
+void ListDiskGroupsCommand::usage() const {
+	std::cerr << name() << " <master ip> <master port>\n";
+	std::cerr << "    Prints disk groups configuration in chunkservers.\n";
+}
+
+void ListDiskGroupsCommand::run(const Options &options) const {
+	if (options.arguments().size() != 2) {
+		throw WrongUsageException(
+		    "Expected <master ip> and <master port> for " + name());
+	}
+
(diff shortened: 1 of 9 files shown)


src/protocol/SFSCommunication.h
```diff
@@ -1751,3 +1751,11 @@ enum class SugidClearMode {
 // 0x0259
 #define CSTOCL_HDD_LIST_V2 (PROTO_BASE+601)
 // N*[ entrysize:16 path:NAME flags:8 errchunkid:64 errtime:32 used:64 total:64 chunkscount:32 bytesread:64 usecread:64 usecreadmax:64 byteswriten:64 usecwrite:64 usecwritemax:64]
+
+// 0x025D
+#define SAU_CLTOCS_ADMIN_LIST_DISK_GROUPS (PROTO_BASE + 605)
+/// -
+
+// 0x025E
+#define SAU_CSTOCL_ADMIN_LIST_DISK_GROUPS (PROTO_BASE + 606)
```
Antuan-AE, line 1760: Please assign this values consistently. All commands with prefix `SAU` are shifted by `1000U` instead of `PROTO_BASE`

src/admin/list_disk_groups_command.cc
```diff
@@ -0,0 +1,98 @@
...
+			yaml << YAML::Value << info;
+		}
+
+		yaml << YAML::EndMap;  // end chunkserver map
+	}
+
+	yaml << YAML::EndSeq;  // end chunkservers list
+	yaml << YAML::EndMap;  // end root map
+
+	std::cout << yaml.c_str() << "\n";
```
Antuan-AE, line 97: Consider the following:
```suggestion
	std::cout << yaml.c_str() << std::endl;
```

Antuan-AE (approved):
LGTM

src/chunkserver/network_worker_thread.cc
```diff
@@ -887,6 +887,18 @@ void worker_hdd_list_v2(csserventry *eptr, const uint8_t *data,
 	hddSerializeAllDiskInfosV2(ptr); // unlock
 }
 
+void worker_list_disk_groups(csserventry *eptr,
+                             [[maybe_unused]] const uint8_t *data,
+                             [[maybe_unused]] uint32_t length) {
+	TRACETHIS();
+
+	std::string diskGroups = hddGetDiskGroups();
+	static constexpr uint8_t kPadding = 5;
```
rolysr, line 896: Please, consider adding a little comment here about why value 5 was selected here to kPadding variable

rolysr (approved):
LGTM

src/chunkserver/network_worker_thread.cc
ralcolea: I have the same doubt :sweat_smile: 
Why adding `kPadding value = 5` to `diskGroups.size()` is needed?

ralcolea (approved):
Good job @lgsilva3087 :+1: :fire: :rocket:

src/admin/list_disk_groups_command.cc
lgsilva3087: There is no need for that. std::endl causes a flush in the stream, which is not needed here.

src/protocol/SFSCommunication.h
lgsilva3087: I removed the SAU_ prefix, because the packet is handled using the legacy way, which is simpler in this case. Now should be consistent.

src/chunkserver/network_worker_thread.cc
lgsilva3087: I renamed the variable to be kSerializedSizePlusNullChar. 5 was needed by the serialize function:
image

PR #159, opened 2024-08-01
fix(tests): Allow to change tests data path

ictus4u:
When the binary for tests was moved to a different path it could not find the tests bash script. 
This commit adds an environment variable to allow the path change in run time.


lgsilva3087 (approved):

rolysr (approved):

uristdwarf (approved):
Perhaps the only thing I would ask is to document this env variable in the README of the tests folder

Antuan-AE (approved):
LGTM

PR #164, opened 2024-08-12
fix(uraft): Fix missing floating ip address

uristdwarf:
If something removes the floating ip address from the device, saunafs-uraft will not bring it back. For example, on Ubuntu systems, if systemd-networkd is restarted, it will reapply netplan config, which inadvertently remove the floating ip address, and uraft has no way of adding this back.

This fix adds a way for uraft to bring the IP address back. First by exporting the IP address as an environment variable in the main program, and then by using in the script to query iproute2 for the IP address, and add it when it's missing.

Note that if two IP addresses with different subnet masks are added, then it won't re-add the IP address with the correct subnet mask.


lgsilva3087 (approved):
LGTM

uristdwarf:
The only failing test worked locally, likely the CI VM was overloaded with the build tests

image

So I'll merge it since it's very unlikely this change broke anything.

PR #168, opened 2024-08-19
test: Improve `stop_ganesha()` function

ralcolea:
The original `stop_ganesha()` function unmounted mountpoints and attempted to kill the Ganesha daemon only once. If the `pkill` operation failed, the Ganesha daemon could remain hanging, leading to Ganesha CI containers becoming "unhealthy" and requiring Jenkins nodes to be restarted.

This commit enhances the `stop_ganesha()` function to handle failed `pkill` attempts. If the initial `pkill` fails, a new Ganesha daemon is started, which should terminate the hanging one. This change aims to ensure proper cleanup of resources and prevent Ganesha CI containers from entering an "unhealthy" state, thus avoiding unnecessary Jenkins nodes restarts.


lgsilva3087 (approved):
LGTM.

PR #169, opened 2024-08-21
feat(chunkserver): Add IDiskManager::getDiskForGC

lgsilva3087:
This commit adds and extension point for the disk managers to be asked about the next disk for garbage collection (GC). The default disk manager returns the next zoned device in a round-robin approach.

It is up to the disk manager plugins to provide custom implementation for this behavior.

### DIFF
--- a/src/chunkserver/chunkserver-common/default_disk_manager.cc
+++ b/src/chunkserver/chunkserver-common/default_disk_manager.cc
@@ -304,3 +304,31 @@ IDisk *DefaultDiskManager::getDiskForNewChunk(
 
 	return bestDisk;
 }
+
+IDisk *DefaultDiskManager::getDiskForGC() {
+	TRACETHIS();
+	IDisk *bestDisk = DiskNotFound;
+
+	std::lock_guard disksLockGuard(gDisksMutex);
+
+	if (gDisks.empty()) {
+		return DiskNotFound;
+	}
+
+	auto diskCount = gDisks.size();
+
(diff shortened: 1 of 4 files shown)


src/chunkserver/chunkserver-common/default_disk_manager.cc
```diff
@@ -304,3 +304,31 @@ IDisk *DefaultDiskManager::getDiskForNewChunk(
...
+	for (size_t i = 0; i < diskCount; ++i) {
+		size_t index = (lastDiskIndexForGC + i) % diskCount;
+		const auto &disk = gDisks[index];
+
+		if (!disk->isZonedDevice() || !disk->isSelectableForNewChunk()) {
+			continue;
+		}
+
+		bestDisk = disk.get();
+		lastDiskIndexForGC = (index + 1) % diskCount;
```
ralcolea, line 329: The update of `lastDiskIndexForGC` equals to `index + 1` suggests to me that the stored value is actually the `nextDiskIndexForGC`, i.e., the initial index considered in the next iteration this function will be called, because the returned disk is the one at `index` position.

ralcolea (approved):
LGTM

src/chunkserver/chunkserver-common/default_disk_manager.cc
lgsilva3087: Nice catch, thanks. Renamed to nextDiskIndexForGC_.

uristdwarf (approved):

PR #173, opened 2024-08-25
fix(tests): Reserve one core in build tests

lgsilva3087:
The build tests inside the saunafs mount point were using all available cores (clamped to 16). In the meantime other tasks need the processor. Sometimes these other tasks are timing out.

This change ensures that there is at least one core to do these other tasks.


ralcolea (approved):
LGTM

wg1k (approved):

PR #174, opened 2024-08-25
feat(chunkserver): Get chunk to test from manager

lgsilva3087:
This commit adds and extension point for the disk managers to be asked about the next chunk to test. The default disk manager uses the same approach as before.

It is up to the disk manager plugins to provide custom implementation for this behavior.


ralcolea (approved):
LGTM :+1: :fire: :rocket:

rolysr (approved):
LGTM

PR #184, opened 2024-09-06
fix(tools): infinite recursion in sfstools.sh

fogti:
Fixes https://github.com/lizardfs/lizardfs/issues/886, requested in https://github.com/lizardfs/lizardfs/pull/887#issuecomment-2333801710
This commit is a port of that fix for LizardFS to SaunaFS.

An example error message was:

```console
$ LC_ALL=C bash -x /usr/bin/mfssetgoal --help
++ basename /usr/bin/mfssetgoal
+ tool=mfssetgoal
+ mfssetgoal --help
bash: warning: shell level (1000) too high, resetting to 1
```

### DIFF
--- a/src/tools/sfstools.sh
+++ b/src/tools/sfstools.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
 
-tool=$(basename $0)
-
-${tool/saunafs/saunafs } "$@"
+tool="$(basename $0)"
+exec saunafs "${tool#sfs}" "$@"
--- a/tests/test_suites/SanityChecks/test_sfstools.sh
+++ b/tests/test_suites/SanityChecks/test_sfstools.sh
@@ -1,0 +1,60 @@
+CHUNKSERVERS=4 \
+	SFSEXPORTS_EXTRA_OPTIONS="allcanchangequota" \
+	# MASTER_CUSTOM_GOALS="2 2: _" \
+	USE_RAMDISK=YES \
+	setup_local_empty_saunafs info
+
+cd "${info[mount0]}"
+
+touch test
+mkdir dir
+
+goal=$(sfsgetgoal dir)
+
+assert_equals "$goal" "dir: 1"
+
+goal=$(sfssetgoal 2 dir)
+
+assert_equals "$goal" "dir: 2"
+
+
+time=$(sfsgettrashtime test)
+assert_equals "$time" "test: 86400"
+sfssettrashtime 0 test
+time=$(sfsgettrashtime test)
+assert_equals "$time" "test: 0"
+
+attrs=$(sfsgeteattr test)
+assert_equals "$attrs" "test: -"
+sfsseteattr -f noattrcache test
+attrs=$(sfsgeteattr test)
+assert_equals "$attrs" "test: noattrcache"
+sfsdeleattr -f noattrcache test
+attrs=$(sfsgeteattr test)
+assert_equals "$attrs" "test: -"
+
+check=$(sfscheckfile test)
+assert_equals "$check" "test:"
+check=$(sfsfileinfo test)
+assert_equals "$check" "test:"
+
+echo "foo" > test
+echo "bar" > test2
+
+sfsappendchunks test2 test
+# We aren't checking the result, since it can take some time. Another test
+# should be check if it's working with (with the saunafs command)
+
+assert_equals "$(sfsdirinfo dir)" "$(saunafs dirinfo dir)"
+
+sfsfilerepair test
+
+sfsmakesnapshot dir dir_snapshot
+
+quota=$(sfsrepquota -d dir)
+assert_equals "$quota" "# User/Group ID/Directory; Bytes: current usage, soft limit, hard limit; Inodes: current usage, soft limit, hard limit;"
+
+sfssetquota -d 1000000 2000 1000000 2000 dir
+
+quota=$(sfsrepquota -d dir)
+assert_equals "$quota" "$(saunafs repquota -d dir)"


src/tools/sfstools.sh
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
-tool=$(basename $0)
-
-${tool/saunafs/saunafs } "$@"
+tool="$(basename $0)"
+case "$tool" in
+  (lizardfs*) tool="${tool#lizardfs}" ;;
+  (mfs*) tool="${tool#mfs}" ;;
+  (saunafs*) tool="${tool#saunafs}" ;;
```
fogti, line 7: It is not clear to me if this should also catch `sfs*`, see also: https://github.com/leil-io/saunafs/commit/64f3f27c1ee7550156192367cbcedeeaac00039d

uristdwarf: It actually should. `sfs` is what `mfs` used to stand for in the code base. `saunafs` is the renamed `lizardfs` tool, which is the binary this script will call. So I don't think there is even a need to add a switch between `saunafs` (`lizardfs`) and `sfs` (`mfs`) as calling `saunafs` directly won't invoke this script at all.

Could you also please remove the `mfs` and `lizardfs`? We don't want to conflict with either MooseFS or LizardFS (as we renamed everything to not conflict, these will not be an issue anymore).

uristdwarf (changes requested):
Generally looks good, consider my suggestions above and I'll try this out.

Also, if it's possible please sign the commit and add a Signed-off message (https://elinux.org/Developer_Certificate_Of_Origin). We are probably introducing DCO in the near future, and it may take a while for this to merge (because we are dealing with issues with the CI)

src/tools/sfstools.sh
fogti: I warn you therefore, you should know that this changes the behavior of the script to strip `sfs` instead of `saunafs` as before. Do you have any release notes where this should be mentioned for package maintainers?

uristdwarf: > you should know that this changes the behavior of the script to strip sfs instead of saunafs as before

This must have been some kind of regression. It should have replaced mfs with lizardfs in LizardFS, but now I see that in LizardFS it's replacing lizardfs with lizardfs!

Digging deeper, the original commit for this looked like this: https://github.com/leil-io/saunafs/commit/23e3af5ac794c0cb790abefa712a6ec40eabd1d9#diff-61ffda4854ab1bec0eabeee8ef0cb7fb6ba6a7c41dd34845e97442dc7fcff00aR1

But this changed with this commit
https://github.com/leil-io/saunafs/commit/4684bb21667b126f14f4c11333e46af46488f1d8#diff-61ffda4854ab1bec0eabeee8ef0cb7fb6ba6a7c41dd34845e97442dc7fcff00aR5

I can only assume this was a mistake, since there's no reason to replace lizardfs with lizardfs, the mfs set of tools (now sfs) are simply this script file.

Why this was never caught I'll never know.

> Do you have any release notes where this should be mentioned for package maintainers?

We have a changelog that automatically fetches commits since previous stable release when we release, and we try to avoid stuff that breaks packaging in minor/patch versions (i.e 4.4.0, the middle and right number should not break anything, including packaging, however the left one could.). I don't believe this could affect package maintainers (unless they fixed this themselves), but perhaps for breaking changes, we could potentially use ReleaseNotes for that since that hasn't been updated since LizardFS (outside of accidentally renaming LizardFS in there to SaunaFS).

fogti: there is a reason to replace `lizardfs` with `lizardfs ` (notice the extra space), and it's also exactly what the original script (in the patch this was ported from) did. Are you sure that you have no more references to `mfs` in the code base (which caused issues in LizardFS)?

uristdwarf: > there is a reason to replace `lizardfs` with `lizardfs ` (notice the extra space)

Ah I see now. But it still doesn't explain why the match was changed from `mfs` to `lizardfs` (see second commit), as how would that help considering calling, for example mfssetgoal, would not replace anything in this case and would call mfssetgoal again, causing an infinite bash recursion (hence the error message). Seems like someone in the past did find and replace but didn't check if this would break the intended behavior here.

The CMake file that generates all the sfs* tools: https://github.com/leil-io/saunafs/blob/main/src/tools/CMakeLists.txt

Notice that all it does is create links for the sfs* tools to this script file. The saunafs binary doesn't touch this script at all. This is the same for LizardFS with mfs*.

https://github.com/lizardfs/lizardfs/blob/master/src/tools/CMakeLists.txt

> Are you sure that you have no more references to mfs in the code base (which caused issues in LizardFS)?

Positive

image

And neither LizardFS references either

image

uristdwarf: In fact, now that I think about it, this mfs to lizardfs change was never released in stable (since they stopped I think around 2018 making any new stable releases, and the commit date for the change was 2019), so this technically is a regression from LizardFS.

uristdwarf: Either way, thanks for taking the time to fix this. We are going to probably remove the sfs* links in 5.0.0 or later and we encourage using saunafs instead. However, while it's still here, might as well fix it as some people coming from LizardFS are still used to it. I was going to fix it myself, but then I remembered there was a hanging PR for this in LizardFS, so I thought I would ask you.

I'll try this out tomorrow and if it works, I think we can merge it (after it passes CI, which might take some time).

uristdwarf (approved):
LGTM. I've added a commit that adds a test to help catch any regressions with these tools in the future.

One final request: Can you please base this on dev branch, as that's our development branch? Thanks again for helping out.

fogti:
isn't that test a bit large? I think it might make more sense to just call `sfsgetgoal --help` and check that it returns successful.

uristdwarf:
> isn't that test a bit large? I think it might make more sense to just call sfsgetgoal --help and check that it returns successful.

I want to make sure that all the tools are working as expected, not only fix this regression.

lgsilva3087 (approved):
LGTM

uristdwarf (approved):

PR #189, opened 2024-09-12
feat(tests): Add RPC service recovery function

ralcolea:
Previously, Ganesha tests failed when RPC became unavailable during the test suite execution. There was also some logic repeated across the suite, such as the creation of Ganesha PID file.

Key changes:
- Introduced a recovery mechanism for the RPC service to handle failures during the test suite execution.
- Refactored the Ganesha PID file creation by encapsulating the logic into a new `create_ganesha_pid_file` function, which is now used across all suite to avoid redundancy.


lgsilva3087 (approved):
LGTM

uristdwarf (approved):

PR #190, opened 2024-09-12
fix(mount): Fix tcptoread-write error message

rolysr:
When trying to start client connection to master, sometimes error receiving data from master was shown as successful, which is not correct. This change improves readability of error by specifying a type of it when logging error message from client.

### DIFF
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -671,10 +671,12 @@ int fs_connect(bool verbose) {
 		return -1;
 	}
 	if (tcptoread(fd,regbuff,8,1000)!=8) {
+		int tcplasterr = tcpgetlasterror();
+		const auto* errorMessage = (tcplasterr != 0) ? strerr(tcplasterr) : strerr(TCPNORESPONSE);
 		if (verbose) {
-			fprintf(stderr,"error receiving data from sfsmaster: %s\n",strerr(tcpgetlasterror()));
+			fprintf(stderr,"error receiving data from sfsmaster: %s\n", errorMessage);
 		} else {
-			safs_pretty_syslog(LOG_WARNING,"error receiving data from sfsmaster: %s",strerr(tcpgetlasterror()));
+			safs_pretty_syslog(LOG_WARNING,"error receiving data from sfsmaster: %s", errorMessage);
 		}
 		tcpclose(fd);
 		fd=-1;
--- a/src/common/sockets.h
+++ b/src/common/sockets.h
@@ -41,6 +41,7 @@
 #define TCPENOTSUP ENOTSUP
 #define TCPEINVAL EINVAL
 #define TCPETIMEDOUT ETIMEDOUT
+#define TCPNORESPONSE EIO
 #endif
 
 #include <vector>


src/common/sockets.cc
```diff
@@ -606,11 +606,15 @@ int32_t tcptoread(int sock, void *buff, uint32_t leng, int msecto) {
...
 		if (tcppoll(pfd, msecto) < 0) {
+			tcpsetlasterror(TCPETIMEDOUT);
 			return -1;
 		}
 		if (pfd.revents & POLLIN) {
 			i = tcprecv(sock, ((uint8_t *)buff) + rcvd, leng - rcvd, 0);
 			if (i == 0 || (i < 0 && tcpgetlasterror() != TCPEAGAIN)) {
+				if (i == 0) {
+					tcpsetlasterror(TCPETIMEDOUT);
+				}
```
uristdwarf, line 617: `tcprecv` seems to be just a wrapper function for `recv`

From `man recv`

```
RETURN VALUE
       Upon successful completion, recv() shall return the length of  the  mes‐
       sage  in bytes. If no messages are available to be received and the peer
       has performed an orderly shutdown, recv() shall return 0. Otherwise,  -1
       shall be returned and errno set to indicate the error.
```

Is it then accurate to set this as ETIMEDOUT as the error value? It did not timeout, it just responded with nothing.

Maybe it should EAGAIN (but this might have consequences in other parts of the code if it's checking that). Have a look at `man errno` and see what error may best fit in this situation.

src/common/sockets.cc
```diff
@@ -632,11 +636,15 @@ int32_t tcptowrite(int sock, const void *buff, uint32_t leng, int msecto) {
 	while (sent < leng) {
 		pfd.revents = 0;
 		if (tcppoll(pfd, msecto) < 0) {
+			tcpsetlasterror(TCPETIMEDOUT);
 			return -1;
 		}
```
uristdwarf, line 641: From `man poll` (again, this just seems to be wrapper function to poll, but the `npfd` parameter hard coded to 1):

```
RETURN VALUE
       Upon successful completion, poll() shall return a non-negative value.  A
       positive value indicates the total number of pollfd structures that have
       selected  events  (that  is,  those for which the revents member is non-
       zero). A value of 0 indicates that the call timed out and  no  file  de‐
       scriptors  have  been selected. Upon failure, poll() shall return -1 and
       set errno to indicate the error.

ERRORS
       The poll() function shall fail if:

       EAGAIN The allocation of internal data structures failed  but  a  subse‐
              quent request may succeed.

       EINTR  A signal was caught during poll().

       EINVAL The  nfds  argument  is greater than {OPEN_MAX}, or one of the fd
              members refers to a STREAM or multiplexer  that  is  linked  (di‐
              rectly or indirectly) downstream from a multiplexer.
```

Judging the possible errors, the previous code is correct that the only thing left to do in this case is return failure with `-1`. Setting the error value is wrong here because it hides the actual error that happened and this timeout is covered in a case below.

src/common/sockets.cc
```diff
@@ -632,11 +636,15 @@ int32_t tcptowrite(int sock, const void *buff, uint32_t leng, int msecto) {
...
 		if (pfd.revents & POLLOUT) {
 			i = tcpsend(sock, ((uint8_t *)buff) + sent, leng - sent, 0);
 			if (i == 0 || (i < 0 && tcpgetlasterror() != TCPEAGAIN)) {
+				if (i == 0) {
+					tcpsetlasterror(TCPETIMEDOUT);
+				}
```
uristdwarf, line 647: If i == 0, then that means while the communication was successful, no data was sent, which may be either correct (as it could not send all the data, so you must resend it again) or a bug (why would we be sending empty data?). Setting ETIMEDOUT feels wrong here as well, as that's not the case really.

uristdwarf (changes requested):
Please be aware that this isn't just affecting client, but anyone who uses common library (pretty much everything). I would do this kind of error checking/setting on the client application, rather than in the shared common library

src/common/sockets.cc
rolysr: At the end I decided it to be `TCPNORESONSE` which is a definition of `EIO` generic _input/ouput_ error. This way I show a more generic reason. And also not affecting the common library as you suggest. Please check my lastest commit for this.

src/common/sockets.cc
rolysr: Decided to delete all code affected from common and let error management for mount.

src/common/sockets.cc
rolysr: Same here, decided to not set last tcp error from master.

src/mount/mastercomm.cc
```diff
@@ -671,10 +671,11 @@ int fs_connect(bool verbose) {
 		return -1;
 	}
 	if (tcptoread(fd,regbuff,8,1000)!=8) {
+		int tcplasterr = tcpgetlasterror();
 		if (verbose) {
-			fprintf(stderr,"error receiving data from sfsmaster: %s\n",strerr(tcpgetlasterror()));
+			fprintf(stderr,"error receiving data from sfsmaster: %s\n",tcplasterr?strerr(tcplasterr):strerr(TCPNORESPONSE));
```
ralcolea, line 676: Considering refactoring the code to avoid duplicated logic and longer lines:
```cpp
int tcplasterr = tcpgetlasterror();
const auto* errorMessage = (tcplasterr != 0) ? strerr(tcplasterr) : strerr(TCPNORESPONSE);

if (verbose) {
	fprintf(stderr,"error receiving data from sfsmaster: %s\n", errorMessage);
} else {
	safs_pretty_syslog(LOG_WARNING,"error receiving data from sfsmaster: %s", errorMessage);
}
```

ralcolea (dismissed):
LGTM.
Please, see my suggestion to make more readable the code.

uristdwarf (approved):

ralcolea (approved):
LGTM.

Issue #192, opened 2024-09-16
IP added to every uraft node.


uristdwarf:
Version: 4.5.0

Commit 161f889 tried to fix an issue where the IP was lost if network was restarted on Ubuntu 24.04. However, due to an oversight, it did not correctly check if the node was leader or not, causing all nodes to add the floating IP address.

As a quick and dirty fix, you can remove these 5 lines from the script https://github.com/leil-io/saunafs/commit/161f889c10e438b2762610747755e790ca22767b#diff-707307719d66efb2d48c2007e62561cb85b508e0e97131cafffa51759f58eec3R158-R162

uristdwarf:
I'll introduce a hotfix shortly, and 4.5.1 should be out this week with this fix.

PR #193, opened 2024-09-16
Release 4.5.1

uristdwarf:
SaunaFS 4.5.1
 
 - Fix critical issue with uraft assigning floating IP address to all nodes.
   Note that this fix reverts the previous fix for lost IP addresses after
   network restart.


uristdwarf:
Fixes #192

NEWS
```diff
@@ -1,5 +1,10 @@
 This file lists noteworthy changes in SaunaFS.
 
+ * SaunaFS (4.5.1) (2024-09-16)
+ - Fix critical issue with uraft assigning floating IP address to all nodes.
```
ralcolea, line 4: If Ganesha improvements are included in the pull request, they shoud also be included in the `NEWS` and in the `changelog` files.

debian/changelog
```diff
@@ -1,3 +1,10 @@
+saunafs (4.5.1) stable; urgency=high
+  * Fix critical issue with uraft assigning floating IP address to all nodes.
```
ralcolea, line 2: Same here.

ralcolea (approved):
LGTM. Please, see my minor suggestions about update of `NEWS` and `changelog` files.

lgsilva3087 (dismissed):
LGTM. Please see Crash's comments about adding the ganesha changes to the changelog and news.

NEWS
uristdwarf: I was thinking about leaving them out (since they are only for the testing framework), but you're probably right, should include them for consistency with the commits at least.

However, small comment: You should not use feat for test because otherwise our current tool will bump the minor version, which we don't want (since it's not changing/adding functionality)

ralcolea: I understand, I did not know.
I used **feat** because the commit adds a feature at the tests level :)
But it's not a problem, I will not do it in the future.

lgsilva3087 (approved):
LGTM.

ralcolea (approved):
LGTM :+1: 
Please, consider to update the branch with the base branch before merging.

aNeutrino (approved):
LGTM

PR #200, opened 2024-09-25
ci: Limit amount of artifacts

uristdwarf:
Due to the current large size of the artifacts being build, strict limits must be put in place to not overwhelm the storage of Jenkins.


ralcolea (approved):
Good job @uristdwarf :+1:

lgsilva3087 (approved):
LGTM

PR #202, opened 2024-09-27
fix(mount): Fix read cache max size as percentage

rolysr:
## Summary

This PR introduces several enhancements and fixes to the read cache mechanism, particularly focusing on handling corrupted files and optimizing read cache memory usage. The changes aim to improve the reliability and efficiency of read operations under various conditions, including low read cache memory scenarios and when encountering corrupted file segments.

### Changes

- **Test for Reading Corrupted Files (`133f7187`):** Added a test to verify that files with corrupted segments can still be read multiple times, despite initial failures that fill the read cache. This ensures that the read cache can release part of its memory over time, allowing continued read operations.

- **Test for Low Read Cache Size Percentage (`4a954e14`):** Introduced a test to confirm that the client does not freeze or fail read operations due to low read cache memory relative to the system's total memory. This test ensures the client's resilience in low-memory conditions.

- **Fix for Read Cache Recovery on Failure (`2fb805ea`):** Addressed an issue where read operations that failed due to insufficient read cache memory did not properly trigger a retry process. Previously, retries continued until reaching the maximum limit without effectively releasing memory. This fix ensures that read requests can complete successfully after memory is freed.

- **Improvements to Read Cache Configuration (`5f972100`, `b37a26e1`):** Implemented enhancements to the read cache configuration, including clearer references to the read cache max size percentage in documentation and help text, and correcting the read cache max size option to represent memory size limits dynamically based on system memory capabilities.

### Impact

These changes significantly improve the system's ability to handle read operations under adverse conditions, such as when dealing with corrupted files or when operating with limited read cache memory. By ensuring that the read cache can efficiently manage and recover from failures, these updates enhance the overall stability and performance of the system.

### Testing

Comprehensive tests have been added to cover the new functionality and fixes, ensuring that the system behaves as expected in scenarios involving corrupted files and low read cache memory. These tests validate the robustness of the read cache mechanism and its ability to adapt to different system conditions.

### Documentation

Updates to the documentation and help text have been made to reflect the changes in read cache configuration options, providing clearer guidance for users on how to optimize read cache settings for their specific system configurations.


doc/sfsmount.1.adoc
```diff
@@ -162,8 +162,8 @@ disables cache (default: 1000).
 *-o readaheadmaxwindowsize=*'KB'::
 Set max value of readahead window per single descriptor in kibibytes (default: 16384).
 
-*-o readcachemaxsize=*'MB'::
-Set max value of read cache size in mebibytes (default: 16384).
+*-o readcachemaxsizepercentage=*'P'::
+Set percentage of system memory used for max value of read cache size (default: 60).
```
lgsilva3087, line 166: Consider ...(default: 60%)

src/mount/fuse/mount_config.cc
```diff
@@ -155,8 +155,8 @@ void usage(const char *progname) {
 				"of a xor chunk\n"
 "    -o bandwidthoveruse=N       define ratio of allowed bandwidth overuse "
 				"when fetching data (default: %.2f)\n"
-"    -o readcachemaxsize=MB      define the maximum cache memory size "
-				"for reading operations in mebibytes (default: %u)\n"
+"    -o readcachemaxsizepercentage=P  define the maximum cache memory size "
+				"percentage from system for reading operations (default: %u)\n"
```
lgsilva3087, line 159: Consider:
Specifies the maximum percentage of system memory for the read cache (default: %u%%)

lgsilva3087 (changes requested):
This change is very important. Please consider adding at least one test. The test should set the percentage to a very low level, like 1%, and then read more than 1% of the total memory. The client should respect the limit but not freeze.

ralcolea (commented):
I agree with @lgsilva3087 about the need of adding the test to validate the change works as expected.

doc/sfsmount.1.adoc
rolysr: Done

src/mount/fuse/mount_config.cc

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,25 @@
...
+		`|cacheexpirationtime=10000" \
+    setup_local_empty_saunafs info
+
+cd "${info[mount0]}"
+
+# Calculate total system memory in MB
+total_memory_mb=$(free --mega | awk '/^Mem:/ {print $2}')
+
+# Calculate 2% of total memory
+two_percent_memory=$(echo "$total_memory_mb * 0.02" | bc)
```
antuan96314, line 15:
```suggestion
two_percent_memory=$(echo "${total_memory_mb} * 0.02" | bc | awk '{printf "%.0f", $0}')
```

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Since 'dd' deals with whole numbers, round the required 
+# size in MB for 2% of memory to the nearest whole number.
+required_size_mb=$(printf "%.0f" "$two_percent_memory")
```
antuan96314, line 19:
```suggestion
required_size_mb=$(printf "%.0f" "${two_percent_memory}")
```

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
antuan96314: Consider rounding to integer using `awk` instead of `printf` since the latest doesn't handle floating point numbers very well by default and might fail.

rolysr: Done

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
rolysr: Done

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Generate a file that is 2% of total memory size
+dd if=/dev/urandom of=test_data bs=1M count=$required_size_mb
```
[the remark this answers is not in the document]
rolysr: Done

lgsilva3087 (dismissed):
LGTM.

tests/test_suites/ShortSystemTests/test_read_cache_ec_data_loss.sh
```diff
@@ -0,0 +1,61 @@
+timeout_set 15 minutes
+
+USE_RAMDISK=YES \
+	CHUNKSERVERS=5 \
+    MOUNT_EXTRA_CONFIG="sfscachemode=NEVER`
+			`|readcachemaxsizepercentage=1`
+			`|sfsioretries=1`
+			`|maxreadaheadrequests=0`
```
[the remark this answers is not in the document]
dmga44: The implications are the following:
- the readahead algorithm is working, which means each read operation is sent to the read workers to work on, resulting in an extra context switch per read operation, but it is not that much overhead.
- no readahead requests are scheduled, meaning that per each read operation, only the explicitly requested data is actually pulled from the chunkservers.
- the read cache is being filled as expected with all the data read as usual for as long we set.

The most important reason for having this option is that we can control which data is queried (only the explicitly requested one) while having the read filled.

tests/test_suites/ShortSystemTests/test_read_cache_ec_data_loss.sh
```diff
@@ -0,0 +1,61 @@
...
+rm "$chunk"
+
+# Delete data in part 3 of the chunk
+csid=$(find_first_chunkserver_with_chunks_matching \
+	"chunk_ec2_3_of_4*${chunk_metadata_extension}")
+hdd=$(get_metadata_path "${info[chunkserver${csid}_hdd]}")
+chunk=$(find "$hdd" -name "chunk_ec2_3_of_4_*${chunk_metadata_extension}")
+rm "$chunk"
+
+# Delete data in part 4 of the chunk
```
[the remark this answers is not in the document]
rolysr: Yes, the chunk is lost but for the tests purposes it is just necessary to be able to read at least the first part. The goal of the test was to proof that after the damage of some file in many of its chunks, and then tried to read this files many times as our client retries, this could fill the read cache, and we are checking that after these retries that fill the cache, then we wait and we are able to make at least one read operation as it is expected the cache to be cleaned correctly and avoid any read cache memory no availability errors.

tests/test_suites/ShortSystemTests/test_read_cache_ec_data_loss.sh
```diff
@@ -0,0 +1,61 @@
+timeout_set 15 minutes
```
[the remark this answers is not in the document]
rolysr: Done. I reduced timeouts to proper values. We can keep it on shorts.

src/mount/readdata.cc
```diff
@@ -81,10 +80,13 @@ bool readShouldWaitForSystemMemory(size_t bytesToReadLeft) {
 }
 
 void increaseUsedReadCacheMemory(size_t bytesToReadLeft) {
-	std::lock_guard lock(gUsedReadCacheMemoryMutex);
 	gUsedReadCacheMemory += bytesToReadLeft;
 }
 
+void decreaseUsedReadCacheMemory(size_t bytesToReadLeft) {
```
lgsilva3087, line 86: Please inline this short function.

src/mount/readdata.cc
```diff
@@ -81,10 +80,13 @@ bool readShouldWaitForSystemMemory(size_t bytesToReadLeft) {
 }
 
 void increaseUsedReadCacheMemory(size_t bytesToReadLeft) {
```
lgsilva3087, line 82: inline

lgsilva3087 (dismissed):
LGTM.

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,30 @@
...
+		`|sfsioretries=14`
+		`|cacheexpirationtime=10000" \
+    setup_local_empty_saunafs info
+
+cd "${info[mount0]}"
+
+# Calculate total system memory in MB
+total_memory_mb=$(free --mega | awk '/^Mem:/ {print $2}')
+
+# Calculate 2% of total memory
```
ralcolea, line 15: What about testing a bigger percent between `readcachemaxsizepercentage` and `FILE_SIZE`?
What if we define `required_size_mb` equals to 10% bigger than `readcachemaxsizepercentage`?

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,30 @@
+timeout_set 2 minutes
+
+MOUNT_EXTRA_CONFIG="sfscachemode=NEVER`
+		`|readcachemaxsizepercentage=1`
+		`|maxreadaheadrequests=0`
+		`|sfsioretries=14`
```
ralcolea, line 6: Why was changed default value of `sfsioretries`? 
Is it measuring something specific?

ralcolea (dismissed):
LGTM :+1: :fire: :rocket: 
I’ve added a few minor suggestions for your consideration, but they’re not blockers for merging.

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Generate a file that is 2% of total memory size
+# dd if=/dev/urandom of=test_data bs=1M count=${required_size_mb}
+FILE_SIZE=${required_size_mb}M BLOCK_SIZE=1M file-generate test_data
+
+# Create a copy of the file to validate generated data
+dd if=test_data of=test_data_copy bs=1M
+
+# Perform file validation on copied test data
+file-validate test_data_copy
```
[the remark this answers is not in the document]
ralcolea: I ran again the same test multiple times and I could not reproduce this issue again.
I'm marking this conversation as resolved.

src/mount/readdata.cc
rolysr: Done

src/mount/readdata.cc
rolysr: Done

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Calculate 2% of total memory
+two_percent_memory=$(echo "${total_memory_mb} * 0.02" | bc | awk '{printf "%.0f", $0}')
+
+# Since 'dd' deals with whole numbers, round the required 
+# size in MB for 2% of memory to the nearest whole number.
+required_size_mb=$(printf "%.0f" "${two_percent_memory}")
+
+# Generate a file that is 2% of total memory size
+# dd if=/dev/urandom of=test_data bs=1M count=${required_size_mb}
```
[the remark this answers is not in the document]
rolysr: Done

tests/test_suites/ShortSystemTests/test_read_corrupted_files.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ $iterations -gt $max_iterations ]; then
+  iterations=$max_iterations
+fi
+
+# Read the file in a loop
+for ((i=0; i<iterations; i++)); do
+  assert_failure pv -B ${SAUNAFS_BLOCK_SIZE} -E -Z ${SAUNAFS_BLOCK_SIZE} -Y ${dir}/file &> /dev/null
+done
+
+sleep 10
```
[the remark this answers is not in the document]
rolysr: The reason for this sleep was to ensure that the `cacheexpirationtime` passes and then the entries in the ReadCache data structure could release the expired entries and avoiding no enough read cache memory for reading exception. After we read the file on a loop many times this make the `ReadCache` to be filled and consequently cause the no enough memory for reading exception, then we wait for a time for releasing this memory in use and then it is possible to perform a read operation with no problem.

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
rolysr: In this test we tried to check that even if the read cache memory percentage is lower than a given file that we would like to read, it could handle the situation of no enough read cache memory by retrying the read operation  and releasing corresponding memory on the `ReadCache`, allowing the client to correctly finish the reading operation and not to freeze or not finishing the read operation. For that specific case of 10% file size bigger that the `readcachemaxsizepercentage` that you suggest, it would be necessary to increase the `sfsioretries` option in order to increase the number of times the read process releases memory from the `ReadCache` and then increase the probability of completing the read request successfully.

tests/test_suites/ShortSystemTests/test_read_cache_low_level_percentage.sh
rolysr: This is because it is just necessary for this test to have a low number of retries for releasing the expected memory to continue with the reading process with no problem. A lower number could not ensure that the memory from `ReadCache` could be enough released. And a higher number won't be necessary as this test should stop before that maximum.

lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM :+1: :fire: :rocket:

PR #206, opened 2024-10-01
Refactor chunkserverentry use std list and smart pointers

lgsilva3087:
Partial refactor of ChunkserverEntry related structs.
Main goals:
- Avoid C-like linked lists and use std::list.
- Simplified memory management (malloc and free was used before).
- Reuse of already allocated memory.
- Use of move semantics.
- Simplified destruction and cleanup.
- Shorter code.
- Remove void* and casts.

### DIFF
--- a/src/chunkserver/network_worker_thread.cc
+++ b/src/chunkserver/network_worker_thread.cc
@@ -153,8 +153,6 @@ std::unique_ptr<PacketStruct> worker_create_detached_packet_with_output_buffer(
 ChunkserverEntry::~ChunkserverEntry() {
 	if (sock >= 0) { tcpclose(sock); }
 	if (fwdSocket >= 0) { tcpclose(fwdSocket); }
-
-	outputPackets.clear();
 }
 
 void ChunkserverEntry::attachPacket(std::unique_ptr<PacketStruct> &&packet) {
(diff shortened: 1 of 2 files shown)


src/chunkserver/network_worker_thread.cc
```diff
@@ -1633,35 +1585,17 @@ void NetworkWorkerThread::operator()() {
 void NetworkWorkerThread::terminate() {
 	TRACETHIS();
 	job_pool_delete(bgJobPool_);
+
```
rolysr, line 1588: Should be considered to keep these extra spaces in the changes?

rolysr (approved):
LGTM. Just one suggestion to consider.

src/chunkserver/network_worker_thread.cc
lgsilva3087: Hi Roly, if you mean the empty lines, they are intentional, to visually separate the code blocks. IMO it is easier to read this way.

ralcolea (dismissed):
Great job @lgsilva3087 :+1: :fire: :rocket:

ralcolea (approved):
LGTM :+1: :fire: :rocket:

rolysr (approved):

PR #207, opened 2024-10-02
ci: Update sfstests to v0.2.1

uristdwarf:
Fixes logs not showing due to a race condition with saunafs-tests reading /var/log/syslog and the file actually being created

https://github.com/leil-io/sfstests/compare/v0.2.0...v0.2.1

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
                 sh '''
                     git clone "https://github.com/leil-io/sfstests"
                     cd sfstests
-                    git checkout v0.2.x
+                    git checkout v0.2.1
                     go build -o $WORKSPACE/sfstests
                     '''
             }


Jenkinsfile
```diff
@@ -18,7 +18,7 @@ pipeline {
                 sh '''
                     git clone "https://github.com/leil-io/sfstests"
                     cd sfstests
-                    git checkout v0.2.x
+                    git checkout v0.2.1
```
lgsilva3087, line 21: Does it means that every time you change the version you need to create a PR in saunafs?

lgsilva3087 (approved):
Not a stopper, but please see my comment.

Jenkinsfile
uristdwarf: Correct, the idea being if something breaks it can be reverted to a previous version, and also allowing me to test newer ones without breaking things.

ralcolea (approved):
LGTM.

PR #209, opened 2024-10-04
test: Apply multiplier regardless if Valgrind enabled or not

uristdwarf:
Some functions were only applying the multiplier if Valgrind was enabled. This fixes that by ensuring it's applied regardless of that

It also applies the multiplier to the default timeout of 15 seconds in assert functions, with it being adjustable through an environment variable.


uristdwarf:
Should fix the pread random test failures in CI

lgsilva3087 (approved):
LGTM.

ralcolea (approved):
LGTM :+1:

PR #210, opened 2024-10-07
Refactor CS move functions to ChunkserverEntry

lgsilva3087:
[MUST be merged after PR#206]

This commit moves most of the hanging functions in
network_worker_thread.cc to the ChunkserverEntry class. Where they make
more sense due to their responsibilities. Also:
- The code now is shorter and easier to read.
- There is no need to pass a pointer to the ChunkserverEntry every time
  for almost all functions.
- Lines are also shorter.
- Every function was documented.
- Small code style aplied close to the related lines.
- Better function names.

This is a partial refactor, the next step is to separate the
ChunkserverEntry into its own files.

### DIFF
--- a/src/chunkserver/network_worker_thread.cc
+++ b/src/chunkserver/network_worker_thread.cc
@@ -33,6 +33,7 @@
 #include <unistd.h>
 #include <atomic>
 #include <cassert>
+#include <cstdint>
 #include <memory>
 #include <mutex>
 #include <set>
@@ -129,14 +130,18 @@ MessageSerializer* MessageSerializer::getSerializer(PacketHeader::Type type) {
 	}
 }
 
-std::unique_ptr<PacketStruct> worker_create_detached_packet_with_output_buffer(
+std::unique_ptr<PacketStruct>
+ChunkserverEntry::createDetachedPacketWithOutputBuffer(
     const std::vector<uint8_t> &packetPrefix) {
 	TRACETHIS();
+
 	PacketHeader header;
 	deserializePacketHeader(packetPrefix, header);
+
 	uint32_t sizeOfWholePacket = PacketHeader::kSize + header.length;
 	std::unique_ptr<PacketStruct> outPacket = std::make_unique<PacketStruct>();
 	passert(outPacket);
+
 	outPacket->outputBuffer = getReadOutputBufferPool().get(sizeOfWholePacket);
 
 	if (outPacket->outputBuffer->copyIntoBuffer(packetPrefix) !=
@@ -144,6 +149,7 @@ std::unique_ptr<PacketStruct> worker_create_detached_packet_with_output_buffer(
 		if (outPacket->outputBuffer) {
 			getReadOutputBufferPool().put(std::move(outPacket->outputBuffer));
 		}
+
 		return nullptr;
 	}
 
@@ -164,8 +170,7 @@ void ChunkserverEntry::preserveInputPacket() {
 	writePacket->packet = std::move(inputPacket.packet);
 }
 
-void worker_create_attached_packet(ChunkserverEntry *eptr,
-                                   std::vector<uint8_t> &packet) {
+void ChunkserverEntry::createAttachedPacket(std::vector<uint8_t> &packet) {
 	TRACETHIS();
 	std::unique_ptr<PacketStruct> outpacket = std::make_unique<PacketStruct>();
 	passert(outpacket);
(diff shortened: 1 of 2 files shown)


src/chunkserver/network_worker_thread.cc
```diff
@@ -176,103 +181,101 @@ void worker_create_attached_packet(ChunkserverEntry *eptr,
...
 	outpacket->startPtr = outpacket->packet.data();
 
-	eptr->attachPacket(std::move(outpacket));
+	attachPacket(std::move(outpacket));
 }
 
-uint8_t *worker_create_attached_packet(ChunkserverEntry *eptr, uint32_t type,
-                                       uint32_t size) {
+uint8_t *ChunkserverEntry::createAttachedPacket(uint32_t type,
+                                                uint32_t opSize) {
```
antuan96314, line 188:
```suggestion
                                                uint32_t outPacketSize) {
```
Consider full descriptive naming for this variable.

src/chunkserver/network_worker_thread.h
antuan96314, line 1: Consider declaring `ChunkserverEntry` as a `class` instead of `struct` type

src/chunkserver/network_worker_thread.h
```diff
@@ -168,14 +168,206 @@ struct ChunkserverEntry {
...
+	/// Creates an attached packet from the given vector.
+	/// The function takes ownership of the vector.
+	void createAttachedPacket(std::vector<uint8_t> &packet);
+
+	/// Creates an attached packet with the given type and operation size.
+	///
+	/// @param type The type of the packet.
+	/// @param opSize The size of the operation.
+	/// @return Pointer to the created packet data.
+	uint8_t *createAttachedPacket(uint32_t type, uint32_t opSize);
```
antuan96314, line 190:
```suggestion
	const uint8_t *createAttachedPacket(uint32_t type, uint32_t opSize);
```
Consider returning const pointer to prevent caller from accidentally sending the wrong data.

src/chunkserver/network_worker_thread.h
```diff
@@ -168,14 +168,206 @@ struct ChunkserverEntry {
...
+	static void readFinishedCallback(uint8_t status, void *entry);
+	/// Callback after delayed close operations.
+	static void delayedCloseCallback(uint8_t status, void *entry);
+	/// Callback for when a write operation finishes.
+	static void writeFinishedCallback(uint8_t status, void *entry);
+	/// Callback for legacy chunk block retrieval completion.
+	static void sauGetChunkBlocksFinishedLegacyCallback(uint8_t status,
+	                                                    void *entry);
+	/// Callback for chunk block retrieval completion.
+	static void sauGetChunkBlocksFinishedCallback(uint8_t status, void *entry);
```
antuan96314, line 306: Why is prefix `sau` on this two particular methods. Consider removing or applying it to all others for consistency

src/chunkserver/network_worker_thread.h
```diff
@@ -168,14 +168,206 @@ struct ChunkserverEntry {
...
+
+	/// Retrieves chunk blocks from the given information using the new way.
+	void sauGetChunkBlocks(const uint8_t *data, uint32_t length);
+
+	/// Retrieves the list with the HDDs information.
+	void hddListV2(const uint8_t *data, uint32_t length);
+
+	/// Lists the disk groups (if the DiskManager supports it).
+	void listDiskGroups([[maybe_unused]] const uint8_t *data,
+	                    [[maybe_unused]] uint32_t length);
```
antuan96314, line 321: Consider using the same rule for `unused` parameter. Either `maybe_unused` or `(void) var;` inside the function...

src/chunkserver/network_worker_thread.cc
```diff
@@ -285,125 +288,128 @@ void worker_delayed_close(uint8_t status, void *e) {
...
 		}
-		eptr->readJobId = job_read(
-		    eptr->workerJobPool, worker_read_finished, eptr, eptr->chunkId,
-		    eptr->chunkVersion, eptr->chunkType, eptr->offset, thisPartSize,
-		    maxReadBehindBlocks, readAheadBlocks,
-		    eptr->readPacket->outputBuffer.get(), !eptr->isChunkOpen);
-		if (eptr->readJobId == 0) {
-			eptr->state = ChunkserverEntry::State::Close;
+
+		readJobId = job_read(workerJobPool, readFinishedCallback, this, chunkId,
```
antuan96314, line 385: Consider use lambda expression to avoid passing this as argument.
eg:
```
writeJobId = job_write(workerJobPool, 
    [this](uint8_t status, void *entry) { this->writeFinishedCallback(status, entry); },
    nullptr, opChunkId, chunkVersion, chunkType, blocknum, opOffset, opSize, crc, dataToWrite);
```

antuan96314 (dismissed):
LGTM

ralcolea (dismissed):
Great job @lgsilva3087 :+1: :fire: :rocket:

src/chunkserver/network_worker_thread.cc
```diff
@@ -176,103 +181,101 @@ void worker_create_attached_packet(ChunkserverEntry *eptr,
...
 	outPacket->startPtr = outPacket->packet.data();
 
-	eptr->attachPacket(std::move(outPacket));
+	attachPacket(std::move(outPacket));
 
 	return ptr;
 }
 
-void worker_fwderror(ChunkserverEntry *eptr) {
+void ChunkserverEntry::fwdError() {
```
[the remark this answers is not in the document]
dmga44: After checking the rest of the code, I think that applying similar changes may be tiresome, also going a little bit ahead of the PR goals and may not clarify much the idea of those functions. So, I think this suggestion can be regarded as extremely optional.

dmga44 (dismissed):
LGTM. 

I have some very minor suggestion, but it is also worth to point out the suggestions made by @rolysr in the PR [217](https://github.com/leil-io/saunafs/pull/217). These suggestions can be applied both in this PR, the 217 or some future one.

src/chunkserver/network_worker_thread.cc
lgsilva3087: Applied

src/chunkserver/network_worker_thread.cc
```diff
@@ -176,103 +181,101 @@ void worker_create_attached_packet(ChunkserverEntry *eptr,
...
-	status = tcpnumconnect(eptr->fwdSocket, eptr->fwdServer.ip,
-	                       eptr->fwdServer.port);
+
+	status = tcpnumconnect(fwdSocket, fwdServer.ip, fwdServer.port);
 	if (status < 0) {
 		safs_pretty_errlog(LOG_WARNING, "connect failed, error");
-		tcpclose(eptr->fwdSocket);
-		eptr->fwdSocket = -1;
+		tcpclose(fwdSocket);
+		fwdSocket = -1;
```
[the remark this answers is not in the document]
lgsilva3087: I will do it in another change not related to just moving the functions.

src/chunkserver/network_worker_thread.h
lgsilva3087: There is no advantage of doing this at this point. The only difference is the default visibility, but this change is not making the members private, it is just moving the functions.

src/chunkserver/network_worker_thread.h
lgsilva3087: This implies changes in other files like the hddspacemgr, which are not related to this change. Also at the end the returned buffer is serialized, needing a const_cast which is also not recommended.

ralcolea (approved):
LGTM :+1: :fire: :rocket:

dmga44 (approved):

PR #213, opened 2024-10-09
fix(tests): Fix call to timeout_rescale

lgsilva3087:
The previous call was expanding to two arguments, possibly causing non-deterministic errors in some tests.


ralcolea (approved):
LGTM.

PR #216, opened 2024-10-12
feat(master): Improve openedfiles per session

dmga44:
These changes target improving the performance of sessions which have
already opened many files. Previous implementation of the openedfiles
member was a linked list sorted by inode which caused that searching,
inserting, counting how many files are open and most operations were
spending time proportional to the already opened files, i.e O(n) if n
is the number of opened files.

The solution proposed changes the type of this member to std::set and
makes complexity of all operations O(logn).

This change also improves readability of this part of the code.


uristdwarf:
I remember doing something similar with https://github.com/leil-io/saunafs/commit/3bbde7864f225656fba5e8cb828a111b5d1d760e but with vector. I think I tried std::set, but either there were implementation issues or performance (sadly I don't remember which). Could you compare the performance of this commit to that commit I posted? std::vector may still be yet faster.

dmga44:
> I remember doing something similar with [3bbde78](https://github.com/leil-io/saunafs/commit/3bbde7864f225656fba5e8cb828a111b5d1d760e) but with vector. I think I tried std::set, but either there were implementation issues or performance (sadly I don't remember which). Could you compare the performance of this commit to that commit I posted? std::vector may still be yet faster.

Mate, I'll comment a few things that I notice on your commit:

uristdwarf:
> * you are anyway traversing the vector, which means you are doing O(n) operations for those operations. In the Tsukaery cluster we have sessions with 7*10^5 open files. So you are through all of them which is not good. The std::set complexity for insertion, deletion, find, and almost everything is O(logn), which means that for these case the cost of insertions and openings should be improved be a 10^3 factor guaranteed. I'm most definately sure of it, but I'd love to create a benchmark for it

I mean it makes complete sense that set should be faster, but I don't remember what problems I faced with set exactly. I did test it on a smaller amount of files I believe (20000 files I believe?)

dmga44:
> Benchmark is a good idea though 👍

Do you still have the script you used for those tests?

uristdwarf:
I used this

https://github.com/kofemann/mdbench

dmga44:
@uristdwarf  Look, but I don't think that this benchmark script is keeping all those files open until the very end, which is what we are currently worried about. I'll create our own very simple benchmark tool, and let's see.

src/master/matoclserv.cc
```diff
@@ -823,34 +819,21 @@ int matoclserv_load_sessions() {
...
-	while ((ofptr=*ofpptr)) {
-		if (ofptr->inode==inode) {
-			return SAUNAFS_STATUS_OK;       // file already acquired - nothing to do
-		}
-		if (ofptr->inode>inode) {
-			break;
-		}
-		ofpptr = &(ofptr->next);
+int matoclserv_insert_openfile(session *cr, uint32_t inode) {
+	if (cr->openedfiles.count(inode) > 0) {
```
lgsilva3087, line 823: Consider using contains instead of count.

ralcolea, line 823: As we are using the most recent C++ version, consider to replace `std::set::count()` by `std::set::contains`.
Both functions have the same time complexity (`O(log n)`), but IMHO `contains()` in this context is more readable and closer to what you're verifying in the function.

src/master/matoclserv.cc
```diff
@@ -863,21 +846,10 @@ void matoclserv_add_open_file(uint32_t sessionid,uint32_t inode) {
...
-	ofpptr = &(asesdata->openedfiles);
-	while ((ofptr=*ofpptr)) {
-		if (ofptr->inode==inode) {
-			return;
-		}
-		if (ofptr->inode>inode) {
-			break;
-		}
-		ofpptr = &(ofptr->next);
+	if (asesdata->openedfiles.count(inode) > 0) {
```
lgsilva3087, line 849: The same here.

lgsilva3087 (approved):
LGTM.

src/master/matoclserv.cc
```diff
@@ -1946,68 +1909,38 @@ void matoclserv_fuse_reserved_inodes(matoclserventry *eptr,const uint8_t *data,u
...
-		if (length) {
-			length--;
-			inode = get32bit(&ptr);
-		} else {
-			inode=0;
+
+	for (const auto &inode_to_reserve : inodes_to_reserve) {
+		if (fs_acquire(context, inode_to_reserve, eptr->sesdata->sessionid) ==
+		    SAUNAFS_STATUS_OK) {
+			// insert the ones that are not already open missing
```
ralcolea, line 1940: This comment can be improved. Maybe the `missing` at the end is not needed?
```cpp
// Insert successfully reserved inodes into the opened files set
```

ralcolea (approved):
Great job @dmga44 :+1: :fire: :rocket: 
Please, see my minor suggestions. However, none of them are stoppers to merge this important improvement.

src/master/matoclserv.cc
dmga44: Applied.

src/master/matoclserv.cc
dmga44: Applied.

src/master/matoclserv.cc
dmga44: Applied.

src/master/matoclserv.cc
dmga44: Applied.

utils/open_close_files_benchmark.cc
```diff
@@ -0,0 +1,133 @@
...
+		          << std::endl;
+		return 1;
+	} else {
+		std::cout << "Created folder '" << folder_path << "' successfully."
+		          << std::endl;
+	}
+
+	auto start = std::chrono::high_resolution_clock::now();
+
+	std::vector<int> test_files_fds(number_of_files);
```
lgsilva3087, line 79: Consider using std::array if already know the number of elements and them will not change, to avoid the heap allocation.

utils/open_close_files_benchmark.cc
```diff
@@ -0,0 +1,133 @@
...
+	exit(status);
+}
+
+int main(int argc, char **argv) {
+	if (argc < 3 || argc > 4) {
+		show_help_message_and_exit(argv[0], 1);
+	}
+
+	std::string testing_path(argv[1]);
+	int number_of_files = atoi(argv[2]);
```
lgsilva3087, line 52: At some point we considered to change the snake casing to semi camel casing for new code. Please consider it.

lgsilva3087 (dismissed):
Great change. Please see my minor suggestions.

src/mount/writedata.cc
```diff
@@ -256,42 +255,31 @@ void write_cb_wait_for_block(inodedata* id, Glock& glock) {
...
-			return id;
-		}
+inodedata *write_get_inodedata(uint32_t inode, Glock &) {
+	if (inodedataMap.contains(inode)) {
+		return inodedataMap[inode];
 	}
-	id = new inodedata(inode);
-	id->next = idhash[idh];
-	idhash[idh] = id;
+	inodedata *id = new inodedata(inode);
```
ralcolea, line 269: Consider using `auto` here to avoid duplicating `inodedata` type.

src/mount/writedata.cc
```diff
@@ -256,42 +255,31 @@ void write_cb_wait_for_block(inodedata* id, Glock& glock) {
...
-		}
-		idp = &(id->next);
+void write_free_inodedata(inodedata *fid, Glock &) {
+	uint32_t inode = fid->inode;
+	if (!inodedataMap.contains(inode)) {
+		return;
 	}
+
+	inodedata *id = inodedataMap[inode];
+	delete id;
```
ralcolea, line 281: Consider using smart pointers to avoid explicitly delete the pointers.

utils/big_session_metadata_benchmark.cc
```diff
@@ -0,0 +1,158 @@
...
+		show_help_message_and_exit(argv[0], 1);
+	}
+
+	std::string folder_path =
+	    testing_path + "/big_sessions_metadata_benchmark_" + std::to_string(time(0));
+	if (!std::filesystem::create_directory(folder_path)) {
+		std::cerr << "Failed to create folder '" << folder_path << "'."
+		          << std::endl;
+		return 1;
+	} else {
```
ralcolea, line 84: Redundant `else` after `return`.

ralcolea (dismissed):
Great change @dmga44 :+1: :fire: :rocket: 
Please consider my minor suggestions, which are not stoppers to merge.

utils/open_close_files_benchmark.cc
dmga44: Well, the ```number_of_files``` variable which is the size of the array needs to be parsed, thus it can not be a constant. Therefore it can not be the size of the std::array because it must be a constant. Can you provide a way to have it done?

Also, I will move the start time just after initializing these vectors to make sure the allocation time is not measured.

utils/open_close_files_benchmark.cc
dmga44: Applied.

src/master/matoclserv.cc
```diff
@@ -890,15 +862,8 @@ void matoclserv_remove_open_file(uint32_t sessionid, uint32_t inode) {
...
-	filelist *ofptr = NULL;
-	filelist** ofpptr = &(asesdata->openedfiles);
-	while ((ofptr = *ofpptr)) {
-		if (ofptr->inode == inode) {
-			*ofpptr = ofptr->next;
-			free(ofptr);
-			break;
-		}
-		ofpptr = &(ofptr->next);
+	if (asesdata->openedfiles.count(inode) > 0) {
```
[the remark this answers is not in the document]
dmga44: Done, I missed a few ones from the previous suggestion, my bad. Thank you :)

src/mount/writedata.cc
dmga44: Applied.

src/mount/writedata.cc
dmga44: This specific change was applied in another branch. I think it can wait until it is finally merged :)

utils/big_session_metadata_benchmark.cc
dmga44: Applied.

lgsilva3087 (approved):
LGTM.

src/mount/writedata.cc
```diff
@@ -256,42 +255,32 @@ void write_cb_wait_for_block(inodedata* id, Glock& glock) {
...
-		}
-		idp = &(id->next);
+void write_free_inodedata(inodedata *fid, Glock &) {
+	uint32_t inode = fid->inode;
+	if (!inodedataMap.contains(inode)) {
+		return;
 	}
+
+	auto id = inodedataMap[inode];
+    sassert(id == fid);
```
ralcolea, line 281: Check the indentation of this line. It looks different.

ralcolea (approved):
Great job @dmga44 :+1: :fire: :rocket:

src/mount/writedata.cc
dmga44: Applied.

ralcolea (approved):
LGTM :+1: :fire: :rocket:

lgsilva3087 (approved):
LGTM

PR #218, opened 2024-10-13
fix: Clarify error offset in file-validate

dmga44:
Current version of the code, when showing the error offset in a
file-validate check many times say a wrong offset inside the file. This
is because it is showing the error offset inside the buffer reading
from the checked file.

The real offset may be obtained from the expected data, but it would be
much easier to show it as a first result.


rolysr (approved):
LGTM

ralcolea (approved):
LGTM :+1: :fire: :rocket:

PR #219, opened 2024-10-14
feat: Configure judy use via directives

dmga44:
This change adds a couple of possible directives for the CMake to
configure the judy library use:
- DISABLE_JUDY_FOR_SPARSEBITSET: disables the use of judy for
SparseBitset class (used in IdPoolDetainer class).
- DISABLE_JUDY_FOR_FSCONTAINERS: disables the use of judy_map for
several filesystem metadata containers.

The motivation for this change is to easily configure the use of judy
because it was used by default otherwise, and some use cases (unlink
speed) show that the replacement structure may perform better.


antuan96314 (approved):
LGTM

rolysr (approved):
LGTM.

PR #228, opened 2024-10-25
feat(mound): Add ignoreflush to tweaks variables

rolysr:
This change allows the users to modify the sfsignoreflush option during runtime execution of the client.

### DIFF
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -215,7 +215,7 @@ static unsigned gDirEntryCacheMaxSize = 100000;
 
 static int debug_mode = 0;
 static int usedircache = 1;
-static bool ignore_flush = false;
+static std::atomic<bool> gIgnoreFlush = false;
 static int keep_cache = 0;
 static double direntry_cache_timeout = 0.1;
 static double entry_cache_timeout = 0.0;
@@ -2408,7 +2408,7 @@ BytesWritten write(Context &ctx, Inode ino, const char *buf, size_t size, off_t 
 }
 
 void flush(Context &ctx, Inode ino, FileInfo* fi) {
-	if (ignore_flush) {
+	if (gIgnoreFlush) {
 		oplog_printf(ctx, "flush (%lu): OK",
 				(unsigned long int)ino);
 		return;
@@ -3411,7 +3411,7 @@ void init(int debug_mode_, int keep_cache_, double direntry_cache_timeout_, unsi
 	mounting_gid = mounting_gid_;
 	allowed_users = allowed_users_;
 #endif
-	ignore_flush = ignore_flush_;
+	gIgnoreFlush = ignore_flush_;
 	debug_mode = debug_mode_;
 	keep_cache = keep_cache_;
 	direntry_cache_timeout = direntry_cache_timeout_;
@@ -3443,6 +3443,7 @@ void init(int debug_mode_, int keep_cache_, double direntry_cache_timeout_, unsi
 			getAcl));
 
 	gTweaks.registerVariable("DirectIO", gDirectIo);
+	gTweaks.registerVariable("IgnoreFlush", gIgnoreFlush);
 	gTweaks.registerVariable("AclCacheMaxTime", acl_cache->maxTime_ms);
 	gTweaks.registerVariable("AclCacheHit", acl_cache->cacheHit);
 	gTweaks.registerVariable("AclCacheExpired", acl_cache->cacheExpired);
--- a/src/mount/writedata.cc
+++ b/src/mount/writedata.cc
@@ -215,6 +215,7 @@ static InodeDataMap inodedataMap;
 
 static uint32_t gWriteWindowSize;
 static uint32_t gChunkserverTimeout_ms;
+static std::atomic<bool> gIgnoreFlush;
 
 // percentage of the free cache (1% - 100%) which can be used by one inode
 static uint32_t gCachePerInodePercentage;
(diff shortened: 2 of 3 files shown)


src/mount/writedata.cc
```diff
@@ -215,6 +215,7 @@ static InodeDataMap inodedataMap;
 
 static uint32_t gWriteWindowSize;
 static uint32_t gChunkserverTimeout_ms;
+static std::atomic<bool> gIgnoreFlush;
```
uristdwarf, line 218: Outside of registering it for tweaks, it doesn't seem this variable is used elsewhere? I'm assuming it's updated when the tweaks file is updated, but will this be used to determine whether to flush or not?

uristdwarf (commented):
Please see my question on gIgnoreFlush

src/mount/sauna_client.cc
```diff
@@ -3496,7 +3496,8 @@ void fs_init(FsInitParams &params) {
 			params.prefetch_xor_stripes,
 			std::max(params.bandwidth_overuse, 1.));
 	write_data_init(params.write_cache_size, params.io_retries, params.write_workers,
-			params.write_window_size, params.chunkserver_write_timeout_ms, params.cache_per_inode_percentage);
+			params.write_window_size, params.chunkserver_write_timeout_ms,
```
ralcolea, line 3499: Please, remove the trailing space at the end of this line, i.e., extra whitespace after `params.chunkserver_write_timeout_ms, `

src/mount/writedata.cc
```diff
@@ -675,7 +676,8 @@ void* write_worker(void*) {
 
 /* API | glock: INITIALIZED,UNLOCKED */
 void write_data_init(uint32_t cachesize, uint32_t retries, uint32_t workers,
-		uint32_t writewindowsize, uint32_t chunkserverTimeout_ms, uint32_t cachePerInodePercentage) {
+		uint32_t writewindowsize, uint32_t chunkserverTimeout_ms,
```
ralcolea, line 679: Please, remove the trailing space after `uint32_t chunkserverTimeout_ms, `

src/mount/writedata.cc
```diff
@@ -703,6 +706,7 @@ void write_data_init(uint32_t cachesize, uint32_t retries, uint32_t workers,
 	pthread_attr_destroy(&thattr);
 
 	gTweaks.registerVariable("WriteMaxRetries", maxretries);
+	gTweaks.registerVariable("IgnoreFlush", gIgnoreFlush);
```
ralcolea, line 709: I see this change does not include an integration test. Please, consider adding an integration test like `SanityChecks/test_mount_directio.sh` to verify that modifying `ignoreflush` option on the fly works correctly. 

Another idea can be to create a test specifically for options that can be tweaked. In that way, every time a new option is added to the tweaks variable, we add the respective check for the new variable.

src/mount/sauna_client.cc
ralcolea: Please, fix also the typo in the commit message: `mound` instead of `mount`.

ralcolea (changes requested):
The change in general looks good. However, I would like to add an integration test to verify the tweak of the option. Please, see my comment.

src/mount/writedata.cc
rolysr: Yes, you are right. The variable was not correctly being updated when its value changed in the tweaks file, the solution for this was to set the `gIgnoreFlush` variable in the tweaks in the `init` function.

src/mount/sauna_client.cc
rolysr: Done

src/mount/writedata.cc
rolysr: Done

src/mount/writedata.cc
rolysr: At the end, I decided to add that requested integration test to the test `SanityChecks/test_mount_tweaks.sh` which is a substitution for `SanityChecks/test_mount_directio.sh` for testing `directio` and `ignoreflush` variables.

tests/master.fs.checksum.mismatch:/tmp/saunafs_error_dir/debug_log_errors_chunkserver_0.log
```diff
@@ -0,0 +1,2905 @@
+10/31/24 00:12:05.812 [info] [21702:21702] : changed working directory to: /tmp/SaunaFS-autotests/saunafs/var/chunkserver_0
```
ralcolea, line 1: I assume this file was incorrectly added?

tests/master.fs.checksum.mismatch:/tmp/saunafs_error_dir/debug_log_errors_chunkserver_1.log
```diff
@@ -0,0 +1,3091 @@
+10/31/24 00:12:05.884 [info] [21803:21803] : changed working directory to: /tmp/SaunaFS-autotests/saunafs/var/chunkserver_1
```
ralcolea, line 1: I assume this file was incorrectly added?

tests/master.fs.checksum.mismatch:/tmp/saunafs_error_dir/debug_log_errors_chunkserver_2.log
```diff
@@ -0,0 +1,1539 @@
+10/31/24 00:12:05.949 [info] [21905:21905] : changed working directory to: /tmp/SaunaFS-autotests/saunafs/var/chunkserver_2
```
ralcolea, line 1: I assume this file was incorrectly added?

ralcolea (approved):
LGTM :+1: 
Please, remove all the `.log` files, that probably were added by mistake.

uristdwarf (approved):

PR #232, opened 2024-10-30
chore: Change how PACKAGE_VERSION string is build

uristdwarf:
Git commits are not guaranteed to be available or reliable on build time (for example, some packaging systems prefer without a VCS or have their own image of the repository with their own commits). Build time may be useful, but it should not be part of the PACKAGE_VERSION, instead it should be generated seperately and used seperately when -v is called. Also, the version should not depend on an outside variable, the CMakeLists.txt file should be the final source for this.

This commit removes the above git commit and build time variables (build time will be re-introduced in another commit) and simplifies the version string, with one exception: A compile time option called VERSION_SUFFIX, which can be used to identify the build (e.g, 'unofficial', 'dev', 'rc' or nothing at all). By default this will be dev when building from source, or 'unofficial' when using ./configure. Package maintainers may use their own suffix to distinguish building processes or add additional version information.


lgsilva3087 (commented):
I agree that we can simplify the package naming. But:
- You are dropping very useful information for debugging.
- Maybe the information should be added to the --version option before dropping it.
- 'dev' is a very generic suffix that could be interpreted as it was generated from the 'dev' branch.

This is how vscode show the related information:
image

uristdwarf:
> You are dropping very useful information for debugging.

I do agree we should preserve some of the info, e.g build date, that is useful for differentiating same versions with different builds.

> I agree with @lgsilva3087. Key information about the packages, such as the commit, should be available in the package name or through an alternative method. This information is essential, particularly for supporting customers, and should not be removed.

However we can't rely on git commits for a couple of reasons:
1. They may only download the source tar, with no git. Even if they get it through git, they may wish to remove the VCS (e.g Debian does this), so what we end up is an empty string.
2.  Even if they do have a git commit, it may very well wrongly point to some other repository, debian for example tracks source code with git also, but they have their own different repostiory 

https://salsa.debian.org/med-team/neo
https://salsa.debian.org/med-team/neo/-/commits/master?ref_type=heads
Compare to upstream
https://github.com/NeuralEnsemble/python-neo/commits/master/

3. Even if they include the right commit (perhaps through a CMake definition), there's no guarantee this commit will be 1:1 with the source code in this repository, as there may have been patches added to it. Example from above

https://salsa.debian.org/med-team/neo/-/tree/master/debian/patches?ref_type=heads

But also see LizardFS

https://salsa.debian.org/debian/lizardfs/-/tree/master/debian/patches?ref_type=heads

This last rule also applies to our own customers, as we may wish to apply customized patches without hosting them here (e.g branding related).

In short, there's no guarantee the source code is not tampered with during the build process, making the commit pointless in terms of reproduction.  You need either the binary(ies) or the build process. This suffix should allow builders to distinguish themselves so it can be figured out how it was build (and from what source with what patches) if needed.

> Maybe the information should be added to the --version option before dropping it.

I agree that some info should be preserved. This was the minimum needed for the new deployment process (distinguish builds between dev, staging and production builds) but -v should produce the build time at least. I'll mark it as a draft.

While we are discussing this, I think it should also include some of the compile time definitions, e.g those related to Judy, Prometheus and other stuff.

> 'dev' is a very generic suffix that could be interpreted as it was generated from the 'dev' branch

Perhaps `source` would be better?

uristdwarf:
I could add CMake definition to optionally include a commit hash that we use in our official (and unofficial) builds

uristdwarf:
No git commit/branch:

image

With git commit/branch

image

The git commit will be extracted in the [package.sh script](https://github.com/leil-io/debian-package/blob/dev/package.sh) for our own packages, but I will do this when these commits are approved

This will need to be provided with the -DGIT_COMMIT or -DGIT_BRANCH

I've also added missing version options to the rest of the binaries.

src/main/main.cc
```diff
@@ -42,9 +42,12 @@
 #include <algorithm>
 #include <atomic>
 #include <fstream>
+#include <iostream>
 #include <list>
 #include <memory>
+#include <ostream>
 
+#include "common/version.h"
```
lgsilva3087, line 50: Please sort the includes.

src/main/main.cc
```diff
@@ -42,9 +42,12 @@
 #include <algorithm>
 #include <atomic>
 #include <fstream>
+#include <iostream>
 #include <list>
 #include <memory>
+#include <ostream>
```
lgsilva3087, line 48: Should not be needed if iostream or fstream is already included.

CMakeLists.txt
```diff
@@ -218,6 +217,14 @@ if(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
...
 endif()
 
+if(GIT_COMMIT)
+    add_compile_definitions(GIT_COMMIT="${GIT_COMMIT}")
+endif()
+
+if(GIT_BRANCH)
+    add_compile_definitions(GIT_BRANCH="${GIT_BRANCH}")
+endif()
+
```
lgsilva3087, line 227: Why not to try to get them as before if they are not provided?
The worse scenario is that they could be empty.

lgsilva3087 (commented):
Nice improvements. Please check my comments.

CMakeLists.txt
uristdwarf: Maybe it could be as an extra development option

lgsilva3087: IMO it is even better, we will know that it was not generated from our official repo.

rolysr (dismissed):
LGTM

src/tools/main.cc
```diff
@@ -20,11 +20,15 @@
 #include "common/platform.h"
 
 #include <cassert>
+#include <cstdlib>
+#include <cstring>
 #include <iostream>
+#include <ostream>
 #include <limits.h>
 #include <unistd.h>
 #include <stdio.h>
```
rolysr, line 29: As a cosmetic suggestion, consider changing the order of the imports here to be alphabetically ordered.

src/common/version.h
```diff
@@ -0,0 +1,42 @@
...
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "common/platform.h"
+
+#include <string>
+
+namespace common {
+
+inline std::string version() {
```
rolysr, line 25: Consider if it is more readable to do this function like this for flexibility and maintainability
```
inline std::string version() {
    std::string version_info = "Version: " SAUNAFS_PACKAGE_VERSION "\n"
                               "Build time: " BUILD_TIME "\n";

    #ifdef GIT_COMMIT
        version_info += "Git commit: " GIT_COMMIT "\n";
    #else
        version_info += "Git commit: N/A\n";
    #endif

    #ifdef GIT_BRANCH
        version_info += "Git branch: " GIT_BRANCH;
    #else
        version_info += "Git branch: N/A";
    #endif

    return version_info;
}```

rolysr (dismissed):
LGTM. My suggestions are not an stopper for merging. Probably at some point we could make an integration test to somehow check the correct setting of some compiler directives to its expected values.

lgsilva3087 (dismissed):
LGTM.

src/common/version.h
uristdwarf: For me it's more readable without the variable addition each time

aNeutrino (approved):
LGTM

lgsilva3087 (approved):
LGTM

PR #235, opened 2024-11-03
fix(client): Fix client using localhost for chunkservers

uristdwarf:
When master gives information regarding chunks and chunkservers, it may give the localhost ip addresses of chunkserver to client, causing the client to search for the chunks on their localhost. This commit fixes that by interpreting localhost from master to mean the master IP address.

### DIFF
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -59,7 +59,9 @@
 #include "protocol/SFSCommunication.h"
 #include "protocol/packet.h"
 
-const uint32_t localhost = 0x7F000001;
+bool isLoopbackAddress(uint32_t ip) {	
+	return ((ip & 0xFF000000) == 0x7F000000); // ip in range (127.0.0.0...127.255.255.255)
+}
 
 struct threc {
 	pthread_t thid;
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -2253,7 +2253,7 @@ uint8_t fs_sauwritechunk(uint32_t inode, uint32_t chunkIndex, uint32_t &lockId,
 	for (auto& server : chunkservers) {
 		// If 127.0.0.1, let's assume it's the same as master
 		if (server.address.ip == localhost) {
-			server.address.ip=masterip;
+			server.address.ip = masterip;
 			safs::log_debug("changing chunkserver ip address 127.0.0.1 to {}", htonl(masterip));
 		}
 	}
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -22,6 +22,7 @@
 #include "mount/mastercomm.h"
 
 #include <limits.h>
+#include <netinet/in.h>
 #include <pthread.h>
 #include <signal.h>
 #include <stdio.h>
@@ -57,6 +58,8 @@
 #include "protocol/matocl.h"
 #include "protocol/SFSCommunication.h"
 #include "protocol/packet.h"
+
+const uint32_t localhost = 0x7F000001;
 
 struct threc {
 	pthread_t thid;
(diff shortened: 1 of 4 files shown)


src/mount/mastercomm.cc
```diff
@@ -2118,6 +2121,13 @@ uint8_t fs_saureadchunk(std::vector<ChunkTypeWithAddress> &chunkservers, uint64_
 		setDisconnect(true);
 		return SAUNAFS_ERROR_IO;
 	}
+	for (auto& server : chunkservers) {
+		// If 127.0.0.1, let's assume it's the same as master
+		if (server.address.ip == localhost) {
+			server.address.ip=masterip;
+			safs::log_debug("changing chunkserver ip address 127.0.0.1 to {}", htonl(masterip));
+		}
+	}
```
rolysr, line 2130: As a minor suggestion consider for readability to encapsulate this logic into a function for reusing it.

src/mount/mastercomm.cc
```diff
@@ -2213,6 +2223,13 @@ uint8_t fs_sauwritechunk(uint32_t inode, uint32_t chunkIndex, uint32_t &lockId,
 		setDisconnect(true);
 		return SAUNAFS_ERROR_IO;
 	}
+	for (auto& server : chunkservers) {
+		// If 127.0.0.1, let's assume it's the same as master
+		if (server.address.ip == localhost) {
+			server.address.ip=masterip;
```
rolysr, line 2229: Consider adding a space in this line to both sides of the `=` .

rolysr (commented):
LGTM. But consider my minor suggestions for code readabilty. On the other hand, did you considered if it is necessary to add a test for checking the fix of the issue that was happening before? Just for checking there is not edge case failing for this.

src/mount/mastercomm.cc
antuan96314, line 2229:
```suggestion
			server.address.ip = masterip;
```

antuan96314 (dismissed):

aNeutrino (changes requested):

src/mount/mastercomm.cc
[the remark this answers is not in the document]
uristdwarf: > and some test ?

I wasn't sure how to actually do this test, because our testing framework doesn't use local ip addresses (master never supported them until recently), but I'll try the test you send out. If it fails, I'll just write a unit test with some master mocking (but it might take a while).

> why MASTER is returning locahost on the FIRSTPLACE ?

Because master stores the chunkserver IP address used to connect and gives this directly to client, and that's it. Previously chunkserver failed to connect at all if it was on localhost, but that has changed.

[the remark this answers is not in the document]
uristdwarf: > to be honest I do not know what is a need/goal of your code change ....

Simple, allow client to connect to chunk servers that use the same localhost IP address as master.

uristdwarf: Done

src/mount/mastercomm.cc
uristdwarf: Done

ralcolea (approved):
LGTM. Please, consider adding an automated test where at least one chunkserver use localhost as IP to validate the change.

uristdwarf:
> LGTM. Please, consider adding an automated test where at least one chunkserver use localhost as IP to validate the change.

I tried adding one, but it isn't simple (with our current testing framework). This problem occurs because of a client on a different host.

You have chunkserver/master on host A, and client on host B. Master returns the chunkserver ip address relative to it's own address on host B. However with the testing framework everything is tested on the single host, where this bug doesn't happen. If I researched this more I'm sure I could find a way to emulate this behavior, but I don't think it's worth the time or effort

uristdwarf:
> If I researched this more I'm sure I could find a way to emulate this behavior, but I don't think it's worth the time or effort

At least not for this PR, which is my personal contribution I made because I found and fixed it in my spare time.

antuan96314 (approved):

PR #240, opened 2024-11-08
build: Require Judy for building

### LINKED ISSUE #239: Master loads metadata very slowly without libjudy
## Version

4.6.0-rc3

## Description

When upgrading an existing installation from 4.5.0 master starts very slowly. It took about 19 hours to finish loading the metadata. 4.5.0 doesn't have this issue, as downgrading it is much faster (around a few minutes on our test cluster)

It seems the final section of the metadata (FREE 1.0) takes the longest, as everything else is much faster.

This is also related to uraft, as uraft won't assign a floating IP address until master loads the metadata and starts.

## Reproduction

1. Setup a 4.5.0 cluster and write a lot of files.
2. Upgrade to 4.6.0
3. Start master from 4.6.0

uristdwarf:
If Judy is not present, the master could take a very long time to load (depending on FREE section size). This fixes that by requiring Judy for building.

Fixes #239

### DIFF
--- a/cmake/Libraries.cmake
+++ b/cmake/Libraries.cmake
@@ -137,7 +137,7 @@ endif()
 
 # Find Judy
 
-find_package(Judy)
+find_package(Judy REQUIRED)
 if(JUDY_FOUND)
   set(SAUNAFS_HAVE_JUDY YES)
   set(SAUNAFS_HAVE_WORKING_JUDY1 ${JUDY_HAVE_WORKING_JUDY1})


uristdwarf:
Note that it's a temporary fix, we still need to figure out why it takes so much time loading (CC @dmga44)

ralcolea (approved):
LGTM

antuan96314 (approved):

uristdwarf:
Not to break any builds for other architectures, I don't think is needed, but it's a lower priority

PR #241, opened 2024-11-14
Fix readcache memory release

rolysr:
### Summary of Changes

- **test(mount): Add test for multi read low cache**
  - Added a test to ensure the system can handle multiple read processes with low read cache memory.

- **feat(mount): Add threads for read records garbage**
  - Introduced threads for each active read record to manage their own read cache garbage collection.

- **feat(mount): Add cache expiration time thread**
  - Added a thread to dynamically update the read cache expiration time for better handling of multiple parallel reads.

- **feat(mount): Fix window read cache suggested size**
  - Adjusted the read ahead window size distribution among read workers to balance the load during multiple parallel reads.

- **feat(mount): Add limitation to extra read requests**
  - Disabled additional read requests when read cache memory is almost full to prevent memory release issues.

- **update(mount): Update collectGarbage function**
  - Improved the `collectGarbage` function to ensure all expired and fully processed entries are released, enhancing read performance.

Note: In the performance benchmark that we've made there aren't no noticeable changes when enough memory is provided. 

Co-authored-by: Dave <dave@leil.io>

### DIFF
--- a/src/mount/readdata.cc
+++ b/src/mount/readdata.cc
@@ -81,10 +81,16 @@ bool readShouldWaitForSystemMemory(size_t bytesToReadLeft) {
 
 inline void increaseUsedReadCacheMemory(size_t bytesToReadLeft) {
 	gUsedReadCacheMemory += bytesToReadLeft;
+	gReadCacheMemoryAlmostExceeded =
+	    gUsedReadCacheMemory >=
+	    static_cast<uint64_t>(0.8 * gReadCacheMaxSize.load());
 }
 
 inline void decreaseUsedReadCacheMemory(size_t bytesToReadLeft) {
 	gUsedReadCacheMemory -= bytesToReadLeft;
+	gReadCacheMemoryAlmostExceeded =
+	    gUsedReadCacheMemory >=
+	    static_cast<uint64_t>(0.8 * gReadCacheMaxSize.load());
 }
 
 uint32_t getBytesToBeReadFromCS(uint32_t index, uint32_t offset, uint32_t size,
@@ -293,6 +299,10 @@ RequestConditionVariablePair *ReadaheadOperationsManager::addRequest_(
 void ReadaheadOperationsManager::addExtraRequests_(
     ReadRecord *rrec, uint64_t currentOffset, uint64_t satisfyingSize,
     uint64_t maximumRequestedOffset) {
+	if (gReadCacheMemoryAlmostExceeded.load()) {
+		return;
+	}
+
 	if (!rrec->readaheadRequests.empty()) {
 		maximumRequestedOffset = rrec->readaheadRequests.lastPendingRequest()
 		                             ->requestPtr->endOffset();
--- a/src/mount/readdata.cc
+++ b/src/mount/readdata.cc
@@ -60,6 +60,8 @@ inline std::condition_variable readOperationsAvailable;
 inline std::mutex gReadaheadRequestsContainerMutex;
 inline int kMaxReadCacheRequestRetries = 10;
 inline uint32_t kMaxWindowConsideringMaxReadCacheSize;
+inline std::atomic<uint64_t> timesRequestedMemory = 0;
+inline std::atomic<uint64_t> successfulTimesRequestedMemory = 0;
 
 std::unique_ptr<IMemoryInfo> createMemoryInfo() {
     std::unique_ptr<IMemoryInfo> memoryInfo;
@@ -393,7 +395,8 @@ void* read_data_delayed_ops(void *arg) {
 	(void)arg;
 
 	pthread_setname_np(pthread_self(), "readDelayedOps");
-
+	
+	int ticksSinceLastUpdate = 0;
 	for (;;) {
 		gReadConnectionPool.cleanup();
 		std::unique_lock gMutexLock(gMutex);
@@ -421,7 +424,35 @@ void* read_data_delayed_ops(void *arg) {
 				++readRecordIt;
 			}
 		}
+
 		gMutexLock.unlock();
+		ticksSinceLastUpdate++;
+		if (ticksSinceLastUpdate == 180) {
+			std::unique_lock lock(gReadCacheMemoryMutex);
+			if (gOriginalCacheExpirationTime_ms.load() == 0) {
+				gCacheExpirationTime_ms.store(0);
+			} else if (gCacheExpirationTime_ms.load() == 0) {
+				gCacheExpirationTime_ms.store(
+				    gOriginalCacheExpirationTime_ms.load());
+			} else if (timesRequestedMemory.load() > 0) {
+				double successRate =
+				    (double)successfulTimesRequestedMemory.load() /
+				    (double)timesRequestedMemory.load();
+				if (successRate < 0.3) {
+					gCacheExpirationTime_ms =
+					    std::max<uint32_t>(gCacheExpirationTime_ms / 3, 1);
+				} else if (successRate > 0.8) {
+					gCacheExpirationTime_ms = std::min<uint32_t>(
+					    gCacheExpirationTime_ms * 2,
+					    gOriginalCacheExpirationTime_ms.load());
+				}
+			}
+			timesRequestedMemory.store(0);
+			successfulTimesRequestedMemory.store(0);
+			lock.unlock();
+			ticksSinceLastUpdate = 0;
+		}
+
 		usleep(USECTICK);
 	}
 }
@@ -532,6 +563,7 @@ void read_data_end(ReadRecord *rrec) {
 
 	std::unique_lock inodeLock(rrec->mutex);
 	rrec->readaheadRequests.discardAllPendingRequests();
+	rrec->stopThread.store(true);
 	inodeLock.unlock();
 }
 
@@ -557,6 +589,7 @@ void read_data_init(uint32_t retries,
 	gChunkserverConnectTimeout_ms = chunkserverConnectTimeout_ms;
 	gChunkserverWaveReadTimeout_ms = chunkServerWaveReadTimeout_ms;
 	gChunkserverTotalReadTimeout_ms = chunkserverTotalReadTimeout_ms;
+	gOriginalCacheExpirationTime_ms = cache_expiration_time_ms;
 	gCacheExpirationTime_ms = cache_expiration_time_ms;
 	gReadaheadMaxWindowSize = readahead_max_window_size_kB * 1024;
 	gReadCacheMaxSize.store((read_chache_max_size_percentage * 0.01) *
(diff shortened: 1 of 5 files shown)


ralcolea (approved):
Great job @rolysr and @dmga44 :+1: :fire: :rocket: 
I shared some minor suggestions for your consideration, but none of them are stoppers to merge!!!

src/mount/readdata.cc
```diff
@@ -410,6 +424,34 @@ void* read_data_delayed_ops(void *arg) {
...
+					gCacheExpirationTime_ms = std::min<uint32_t>(
+					    gCacheExpirationTime_ms * 2,
+					    gOriginalCacheExpirationTime_ms.load());
+				}
+			}
+			timesRequestedMemory.store(0);
+			successfulTimesRequestedMemory.store(0);
+			lock.unlock();
+			ticksSinceLastUpdate = 0;
+		}
```
lgsilva3087, line 453: Consider to move this part to an inline function. Also:
- Create constants for the magic numbers, e.g.: constexpr double kLowerSuccesRate = 0.3.
- Variable timesRequestedMemory is atomic, but is also used always protected by a lock on gReadCacheMemoryMutex. Atomic variables comes with a cost, if they are already protected, please consider using a regular variable.
- I think the same happens with successfulTimesRequestedMemory. Please check.

src/mount/readdata.cc
```diff
@@ -536,11 +579,13 @@ void read_data_init(uint32_t retries,
 	gChunkserverConnectTimeout_ms = chunkserverConnectTimeout_ms;
 	gChunkserverWaveReadTimeout_ms = chunkServerWaveReadTimeout_ms;
 	gChunkserverTotalReadTimeout_ms = chunkserverTotalReadTimeout_ms;
+	gOriginalCacheExpirationTime_ms = cache_expiration_time_ms;
 	gCacheExpirationTime_ms = cache_expiration_time_ms;
 	gReadaheadMaxWindowSize = readahead_max_window_size_kB * 1024;
 	gReadCacheMaxSize.store((read_chache_max_size_percentage * 0.01) *
 	                        gMemoryInfo->getTotalMemory());
 	gReadWorkers = read_workers;
+	kMaxWindowConsideringMaxReadCacheSize = gReadCacheMaxSize.load() / gReadWorkers;
```
lgsilva3087, line 588: From the variable name, look like there is an assignment on a constant.

src/mount/readdata.h
```diff
@@ -211,11 +210,19 @@ struct ReadRecord {
 	std::atomic<uint8_t> refreshCounter = 0;
 	std::atomic<uint16_t> requestsNotDone = 0;
 	bool expired = false; //gMutex
+	std::atomic<bool> stopThread;
+	std::thread garbageCollectorThread;
```
lgsilva3087, line 214: Looks like every ReadRecord now contain a garbage collection thread. I am not familiar with this part, but, does it means that if we are reading from 100 files in parallel, we will have 100 GC threads?
Creating and destroying threads is expensive.

src/mount/readdata.h
```diff
@@ -211,11 +210,19 @@ struct ReadRecord {
 	std::atomic<uint8_t> refreshCounter = 0;
 	std::atomic<uint16_t> requestsNotDone = 0;
 	bool expired = false; //gMutex
+	std::atomic<bool> stopThread;
```
lgsilva3087, line 213: Consider using std::jthread if you need this behavior.

src/mount/readdata.h
```diff
@@ -211,11 +210,19 @@ struct ReadRecord {
...
+	std::thread garbageCollectorThread;
 
 	ReadRecord(uint32_t inode)
 	    : cache(gCacheExpirationTime_ms),
 	      readahead_adviser(gCacheExpirationTime_ms, gReadaheadMaxWindowSize),
-	      inode(inode) {}
+	      inode(inode),
+	      stopThread(false),
```
lgsilva3087, line 220: If it is always initialized with false, please use member initializer instead of constructor parameter.

src/mount/readdata_cache.h
```diff
@@ -412,8 +422,11 @@ class ReadCache {
 			reserved_entries_.push_back(*e);
 		} else {
 			assert(e->refcount == 0);
-			std::unique_lock lock(gUsedReadCacheMemoryMutex);
+			std::unique_lock lock(gReadCacheMemoryMutex);
 			gUsedReadCacheMemory -= e->buffer.size();
+			gReadCacheMemoryAlmostExceeded =
+			    gUsedReadCacheMemory >=
+			    static_cast<uint64_t>(0.8 * gReadCacheMaxSize.load());
```
lgsilva3087, line 429: Use the already mentioned constant here.

src/mount/readdata_cache.h
```diff
@@ -424,13 +437,16 @@ class ReadCache {
...
+		std::unique_lock<std::mutex> lock(gReadCacheMemoryMutex,
 		                                  std::defer_lock);
 		while (!reserved_entries_.empty() && count-- > 0) {
 			Entry *e = std::addressof(reserved_entries_.front());
 			if (e->refcount == 0) {
 				lock.lock();
 				gUsedReadCacheMemory -= e->buffer.size();
+				gReadCacheMemoryAlmostExceeded =
+				    gUsedReadCacheMemory >=
+				    static_cast<uint64_t>(0.8 * gReadCacheMaxSize.load());
```
lgsilva3087, line 449: Same here, or even better, an inline funtion, maybe named updateReadCacheMemoryAlmostExceeded, and reuse it everywhere.

src/mount/readdata.cc
lgsilva3087, line 588: I usually use 256 read workers on powerfull servers and 64 MiB of max readahead window size. Does it means it will reduce the effective window size to 256 KiB?

src/mount/readdata.h
```diff
@@ -230,6 +237,21 @@ struct ReadRecord {
...
 
+	void readCacheGarbageCollectionThread() {
+		while (!stopThread.load()) {
+			std::unique_lock inodeLock(this->mutex);
+			this->cache.collectGarbage();
+			inodeLock.unlock();
+			std::this_thread::sleep_for(std::chrono::milliseconds(300));
+		}
+	}
+	void terminateThread() {
```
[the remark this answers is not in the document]
rolysr: Done

src/mount/readdata.cc
```diff
@@ -422,6 +424,34 @@ void* read_data_delayed_ops(void *arg) {
 			}
 		}
 		lock.unlock();
+
+		ticksSinceLastUpdate++;
+		if (ticksSinceLastUpdate == 180) {
```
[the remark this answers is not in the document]

src/mount/readdata.cc
```diff
@@ -422,6 +424,34 @@ void* read_data_delayed_ops(void *arg) {
...
+					    std::max<uint32_t>(gCacheExpirationTime_ms / 3, 1);
+				} else if (successRate > 0.8) {
+					gCacheExpirationTime_ms = std::min<uint32_t>(
+					    gCacheExpirationTime_ms * 2,
+					    gOriginalCacheExpirationTime_ms.load());
+				}
+			}
+			timesRequestedMemory.store(0);
+			successfulTimesRequestedMemory.store(0);
+			lock.unlock();
```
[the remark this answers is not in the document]

src/mount/readdata.cc
```diff
@@ -59,6 +59,9 @@
 inline std::condition_variable readOperationsAvailable;
 inline std::mutex gReadaheadRequestsContainerMutex;
 inline int kMaxReadCacheRequestRetries = 10;
+inline uint32_t kMaxWindowConsideringMaxReadCacheSize;
```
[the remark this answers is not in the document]

src/mount/readdata.cc
```diff
@@ -59,6 +59,9 @@
 inline std::condition_variable readOperationsAvailable;
 inline std::mutex gReadaheadRequestsContainerMutex;
 inline int kMaxReadCacheRequestRetries = 10;
```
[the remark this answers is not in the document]

src/mount/readdata.cc
```diff
@@ -59,6 +59,9 @@
...
+inline uint32_t kMaxWindowConsideringMaxReadCacheSize;
+inline std::atomic<uint64_t> timesRequestedMemory = 0;
+inline std::atomic<uint64_t> successfulTimesRequestedMemory = 0;
```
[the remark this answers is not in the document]

src/mount/readdata.cc
```diff
@@ -410,6 +424,34 @@ void* read_data_delayed_ops(void *arg) {
 			}
 		}
 		lock.unlock();
+
+		ticksSinceLastUpdate++;
+		if (ticksSinceLastUpdate == 180) {
+			std::unique_lock lock(gReadCacheMemoryMutex);
```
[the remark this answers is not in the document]

src/mount/readdata.cc

src/mount/readdata.h
rolysr: Done

src/mount/readdata_cache.h
rolysr: Done

src/mount/readdata_cache.h
rolysr: Done

src/mount/readdata.cc

src/mount/readdata.cc
```diff
@@ -81,10 +84,16 @@ bool readShouldWaitForSystemMemory(size_t bytesToReadLeft) {
 
 inline void increaseUsedReadCacheMemory(size_t bytesToReadLeft) {
 	gUsedReadCacheMemory += bytesToReadLeft;
+	gReadCacheMemoryAlmostExceeded =
+	    gUsedReadCacheMemory >=
+	    static_cast<uint64_t>(0.8 * gReadCacheMaxSize.load());
```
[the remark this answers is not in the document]

src/mount/readdata.cc
rolysr: In that case that you are describing, you would need that the server you will use have at most 16 GiB. This is because in that case we should have `kMaxWindowConsideringMaxReadCacheSize <= 64 MiB`, and then in this [code](https://github.com/leil-io/saunafs/commit/09cf1ce8a0d7fb2cfd866eb86ee540d2c153b0a2#diff-7464618fee0a32f884909855b3a92331f9ba6649e3ab3fd5a2262b43b837f9e7L212), we should have `recommendedSize = kMaxWindowConsideringMaxReadCacheSize <= 64 MiB`, and then affecting your initial configuration as it was supposed to behave. Then in largers servers with quite more than 16 GiB RAM, there should be no limitation to your initial max readahead window size value.

src/mount/readdata.cc
```diff
@@ -383,9 +407,12 @@ void* read_data_delayed_ops(void *arg) {
 
 	pthread_setname_np(pthread_self(), "readDelayedOps");
 
+	std::unique_lock gMutexLock(gMutex, std::defer_lock),
+	    gUsedReadCacheMemoryLock(gReadCacheMemoryMutex, std::defer_lock);
```
lgsilva3087, line 411: Please use two lines for two declarations.

lgsilva3087 (commented):
Please see my comment.

src/mount/readdata.cc
lgsilva3087: Clear.

src/mount/readdata.h
rolysr: Tested using `jthread` but the current approach was behaving more optimal and stable on tests. Because the `stopThread` variable allows me an easier cleaning control over the thread even before the object destruction at least for the case I needed.

src/mount/readdata.cc

src/mount/readdata.h
lgsilva3087: This was replaced. Now it is done in the read_delayed_ops thread.

lgsilva3087 (dismissed):
LGTM.

src/mount/readdata_cache.h
```diff
@@ -24,22 +24,45 @@
 
 #include "common/small_vector.h"
 #include "common/time_utils.h"
+#include "mount/memory_info.h"
```
ralcolea, line 27: Please, check if this header is really needed.

ralcolea (approved):
Great job @rolysr, @dmga44 and @lgsilva3087 :+1: :fire: :muscle:

src/mount/readdata_cache.h
rolysr: Done

ralcolea (approved):
LGTM :muscle: :fire: :rocket:

lgsilva3087 (approved):
LGTM

PR #243, opened 2024-11-19
doc(master): Add man docs for caseinsensitive

rolysr:
The implemented caseinsensitive option for sfsexports files was not given as part of all man documentation as other options from these configuration files. This change adds a basic doc for it.


uristdwarf (approved):

dmga44 (approved):
LGTM.

PR #244, opened 2024-11-19
feat: Improve Judy related compiler directives

dmga44:
Changes:
- remove the DISABLE_JUDY_FOR_SPARSEBITSET directive: it is extremely
harmful in some cases to put it ON. The load time of free inodes could
increase dramatically.
- split the DISABLE_JUDY_FOR_FSCONTAINERS directive into specific ones
for each container. This change targets higher customizability.
- set DISABLE_JUDY_FOR_ENTRIESCONTAINER to ON by default. Performance
tests show high impact of this option in the unlink speed, and no
downsides for now, thus it would be nice to leave it ON by default.


uristdwarf (approved):
LGTM, but check if `DISABLE_JUDY_FOR_SPARSEBITSET` ifdef is still used, and if you don't want to preserve the option, remove it.

rolysr (approved):
LGTM

uristdwarf (approved):

PR #247, opened 2024-11-22
refactor(common): Replace raw queue with PCQueue

ralcolea:
This commit replaces previous C-based pcqueue with a modern PCQueue class to improve resource management and maintainability. This refactor transitions from C-style manual memory management to smart pointers and modern C++ synchronization primitives like `std::condition_variable` and `std::mutex` to ensure thread-safe concurrent access.

### DIFF
--- a/src/common/pcqueue.cc
+++ b/src/common/pcqueue.cc
@@ -19,276 +19,139 @@
  */
 
 #include "common/platform.h"
-#include "common/pcqueue.h"
+#include "pcqueue.h"
 
-#include <errno.h>
-#include <inttypes.h>
-#include <pthread.h>
-#include <stdlib.h>
-
-#include "common/massert.h"
+#include <cerrno>
+#include <cstdlib>
 #include "devtools/TracePrinter.h"
-#include "errors/sfserr.h"
+#include <pthread.h>
 
-typedef struct _qentry {
-	uint32_t id;
-	uint32_t op;
-	uint8_t *data;
-	uint32_t leng;
-	struct _qentry *next;
-} qentry;
-
-typedef struct _queue {
-	qentry *head,**tail;
-	uint32_t elements;
-	uint32_t size;
-	uint32_t maxsize;
-	uint32_t freewaiting;
-	uint32_t fullwaiting;
-	pthread_cond_t waitfree,waitfull;
-	pthread_mutex_t lock;
-} queue;
-
-void* queue_new(uint32_t size) {
+ProducerConsumerQueue::ProducerConsumerQueue(uint32_t maxSize, Deleter deleter)
+    : maxSize_(maxSize), currentSize_(0), deleter_(deleter) {
 	TRACETHIS();
-	queue *q;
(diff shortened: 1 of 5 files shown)


src/common/pcqueue.h
```diff
@@ -22,24 +22,58 @@
...
 
 template<typename T>
-void queue_deleter_delete(uint8_t *p) {
+void deleterByType(uint8_t *p) {
 	delete ((T*)p);
 }
 
+inline void deleterDummy(uint8_t * /*unused*/) {}
+
+class PCQueue {
```
lgsilva3087, line 38: Consider explicitely adding or better removing the constructors and assignment operators to avoid misuse. From the usage looks like there is no need to allow deep copies.

lgsilva3087, line 38: From the context, looks like PCQueue comes from ProducerConsumerQueue, please consider to rename it to the full name or at least ProdConsQueue.

lgsilva3087, line 38: Consider adding minimum the class description documentation. Also: when is it recomented to use it? Maybe how to use it?

lgsilva3087 (commented):
Great change. Please see my suggestions.

src/common/pcqueue.h
```diff
@@ -22,24 +22,58 @@
...
+	struct QueueEntry {
+		uint32_t jobId;
+		uint32_t jobType;
+		uint8_t *data;
+		uint32_t length;
+
+		QueueEntry(uint32_t jobId, uint32_t jobType, uint8_t *data,
+		           uint32_t length)
+		    : jobId(jobId), jobType(jobType), data(data), length(length) {}
+
```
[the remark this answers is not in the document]
ralcolea: Nice catch. Thanks.
Applied the suggestions :)

src/common/pcqueue.cc
```diff
@@ -19,276 +19,138 @@
...
-	if (data) {
-		*data = qe->data;
-	}
-	if (leng) {
-		*leng = qe->leng;
-	}
-	free(qe);
-	return 0;
+
+	auto entry = queue_.front();
```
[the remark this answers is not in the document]
ralcolea: Done

src/common/pcqueue.h
ralcolea: Done

ralcolea: Done

ralcolea: Done

lgsilva3087 (approved):
LGTM. Great.

rolysr (approved):
LGTM. Probably next steps would be to do the same to `workerthreads` and `jobhash` fields on the `jobpool` struct. This would be perfect batches to continue the improvement of the old C-style code to a modern C++ version in this section.

dmga44 (approved):
Very nice change👍.

lgsilva3087 (approved):
LGTM.

dmga44 (approved):

PR #250, opened 2024-11-28
fix(pcqueue): Remove Helgrind warnings

ralcolea:
The `ProducerConsumerQueue` class previously triggered Helgrind warnings due to condition variables being notified (`notify_one()`) without holding the associated mutex. This behavior could result in race conditions and undefined behavior.

This commit ensures that `notEmpty` and `notFull` condition variables are always notified while holding the mutex, guaranteeing thread safety and proper synchronization.


lgsilva3087 (approved):
LGTM

rolysr (approved):
LGTM

PR #178, opened 2024-09-02
fix(mount): Fix client race conditions

dmga44:
This PR targets getting rid of race conditions in the client side
with the help of helgring (valgrind suite). Though it does not remove
all of the issues pointed out by helgrind, it does fixes many of them.

This PR also fixes the issue with the many parallel readings, thus 
bringing back the default values of the cacheexpirationtime parameter.

List of changes:
- check the status of the entries in result (ReadCache::Result) before returning
it in the request function.
- rename operations manager mutex to gReadaheadRequestsContainerMutex
and restrict its use to the cases when the container is used.
- add mutex per ReadRecord to protect its readaheadRequests,
readahead_adviser and cache.
- add mutex per read cache's Entry to protect its members, specially
the buffer and the timer.
- fix issue that may cause crash in the client related with the proper release
of the entries.
- remove ReadRecord's locator and replace it by creating locators per
ChunkReader.
- protect terminate variable from masterproxy.
- protect inodedata's maxfleng access in writedata.

Add some extra test?: helgrind seems to keep pointing at issues that we can't solve
Performance tests: performance did not show any difference.

PR #251, opened 2024-11-28
test(all): Add new basic helgrind test

lgsilva3087:
The threading safety was not checked explicitely in our tests. Threading issues can cause non-deterministic runtime errors that are hard to reproduce and debug. We use multiple threads so far in the chunkservers and clients.

This change introduces a basic test using helgrind (from valgrind). The idea is to gradually add complexity to this test and create others (more specific) when needed.

Some extra commits were added for already detected warnings.

### DIFF
--- a/src/mount/readdata.cc
+++ b/src/mount/readdata.cc
@@ -328,7 +328,6 @@ inline ConnectionPool gReadConnectionPool;
 inline ChunkConnectorUsingPool gChunkConnector(gReadConnectionPool);
 inline ReadaheadOperationsManager gReadaheadOperationsManager;
 inline std::mutex gMutex;
-inline std::mutex gReadCacheMemoryMutex;
 
 inline ReadRecords gActiveReadRecords;
 inline pthread_t delayedOpsThread;
@@ -369,7 +368,7 @@ bool read_data_get_prefetchxorstripes() {
 }
 
 inline void clear_active_read_records() {
-	std::unique_lock lock(gMutex);
+	std::unique_lock gMutexLock(gMutex);
 
 	for (ReadRecords::value_type& readRecord : gActiveReadRecords) {
 		delete readRecord.second;
@@ -385,7 +384,7 @@ void* read_data_delayed_ops(void *arg) {
 
 	for (;;) {
 		gReadConnectionPool.cleanup();
-		std::unique_lock lock(gMutex);
+		std::unique_lock gMutexLock(gMutex);
 		if (readDataTerminate) {
 			return EMPTY_REQUEST;
 		}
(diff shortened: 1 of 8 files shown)


uristdwarf (approved):
So far looks good to me. Even if all the scenarios are not fully implemented, I think it's good to have this in dev ASAP for other developers

aNeutrino:
@lgsilva3087 consider add simple explanation/rationale for this option in your commit description
`--history-level=approx`

lgsilva3087:
> @lgsilva3087 consider add simple explanation/rationale for this option in your commit description `--history-level=approx`

Done.

ralcolea (approved):
Great job @lgsilva3087 and @dmga44 :+1: :fire: :muscle:

src/mount/readdata.cc
```diff
@@ -441,15 +443,20 @@ void* read_worker(void *arg) {
...
 		if (request->error_code != SAUNAFS_STATUS_OK
 		    || request->state == ReadaheadRequestState::kDiscarded) {
-			readRecord->requestsNotDone--;
-			std::unique_lock entryLock(*entry->mutexPtr);
+			{
+				std::unique_lock inodeLock(readRecord->mutex);
+				request.reset();
+			}
+
+			std::unique_lock entryLock(entry->mutex);
```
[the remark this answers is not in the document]
dmga44: You are very much right :)

The point of the entryLock there is to make helgrind happy. It came from another batch of changes ( #178 ) that reduced the number of helgrind warnings, but not removed at all the false positives. In this case, I think it mitigated a false positive: release of the entry versus previous read or write of some of these parameters.

I ran the test that we added on the branch with the current changes after removing this line and it passed. But, consider the following, this lock works very similar to the lock added on ```ReadCache::Result::release()``` which highly impacts the false positives on the test, maybe the test is not going through the lock you point out.

Thanks

ralcolea (approved):
Great job!!! :clap: :fire: :rocket:

src/mount/readdata.cc
```diff
@@ -663,15 +672,15 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
 			uint32_t read_cache_bytes_to_reserve =
 				getBytesToBeReadFromCS(reader.index(), offset_in_chunk,
 										size_in_chunk, reader.fileLength());
-			std::unique_lock lock(gReadCacheMemoryMutex);
+			std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
```
rolysr, line 675: Consider if it is optimal creating a deferred version of this lock outside the container `while` and then lock and unlock as needed. Consider extending the change to all `usedMemoryLock` uses inside the upper `while`.

dmga44: Let's add this change in further PRs, that we are now currently working on. Thank you.

PR #256, opened 2024-12-12
test: Move helgrind test to SingleMachineTests

uristdwarf:
This test was not working good with the limited resources and other tests running. This moves the test to another suite which is run consecutively and with full machine resources.

* Rename MachineTests to a better and more clearer name


lgsilva3087 (approved):
LGTM

rolysr (approved):
LGTM

PR #257, opened 2024-12-13
fix(mount): Fix case-insensitive test

rolysr:
On the Windows client, the case-insensitive test was failing because the client was not correctly updating the case insensitive flag for a given directory. Also, the test had a line with an incorrect path when checking the proper deletion of files. This commit fixes those issues.

### DIFF
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -447,9 +447,7 @@ uint8_t fs_lookup(const FsContext &context, uint32_t parent, const HString &name
 		return SAUNAFS_ERROR_EINVAL;
 	}
 
-	auto parentNode = static_cast<FSNodeDirectory *>(wd);
-	parentNode->case_insensitive = context.sesflags() & SESFLAG_CASEINSENSITIVE;
-	FSNode *child = fsnodes_lookup(parentNode, name);
+	FSNode *child = fsnodes_lookup(static_cast<FSNodeDirectory*>(wd), name);
 	if (!child) {
 		return SAUNAFS_ERROR_ENOENT;
 	}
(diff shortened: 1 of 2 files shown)


src/master/filesystem_operations.cc
```diff
@@ -960,6 +960,8 @@ uint8_t fs_mknod(const FsContext &context, uint32_t parent, const HString &name,
 	    fsnodes_quota_exceeded_dir(wd, {{QuotaResource::kInodes, 1}})) {
 		return SAUNAFS_ERROR_QUOTA;
 	}
+	static_cast<FSNodeDirectory *>(wd)->case_insensitive =
```
ralcolea, line 963: Did you consider to move the responsibility of updating `case_insensitive` attribute to other function, like `fsnodes_fill_attr()`?

src/master/filesystem_operations.cc
```diff
@@ -1005,6 +1007,8 @@ uint8_t fs_mkdir(const FsContext &context, uint32_t parent, const HString &name,
...
```
ralcolea, line 1010: Same comment here.

ralcolea (approved):
LGTM :+1:

src/master/filesystem_operations.cc
rolysr: Considered. But in this case it is useful to keep the implementation on `mkdir `and `mknod ` only in order to specify that these are the places I want to make the set. I consider that `fsnodes_fill_attr` is mainly focused on the attributes than other specific data from nodes and also I consider it won't be necessary to execute this specific code in the other places where `fsnodes_fill_attr` is called. However, thanks to your suggestion I notice an extra place that I was adding this logic and not being completely necessary.

src/master/filesystem_operations.cc
rolysr: Answered [here](https://github.com/leil-io/saunafs/pull/257#discussion_r1895859494).

ralcolea (approved):
LGTM :+1:

dmga44 (approved):
LGTM

PR #262, opened 2024-12-22
fix: Fix test_auto_recovery_incversion (Windows client tests)

dmga44:
This commit adds a small sleep right before printing the metadata to
allow the master to dump metadata. This process seems to be slower on
Windows (WSL1) testing, which is why it is needed.


ralcolea (approved):
LGTM :+1: 
Please, consider changing the commit tag from `fix` to `test`.
IMHO, this tag is more consistent with the change.

rolysr (approved):
LGTM.

PR #263, opened 2024-12-23
refactor: Minor logging changes

dmga44:
This PR contains two relativetly unrelated changes:
- remove extra endline in some mount debug in some log_debug calls in the mount code. These were the only ones doing it and the log_debug already added a newline after the message, so an empty line was generated.
- support new log options in Windows Client. This change includes removing the automatic flush level at err and defining the default value of the new added options (sfsloglevel and sfslogflushlevel) in the default place. These changes only affect Windows Client code.


ralcolea (approved):
LGTM :+1: 
Please, consider merging both commits in a single one, as one of them only removes endline symbols and do not change the logic.

rolysr (approved):
LGTM

lgsilva3087 (approved):

PR #266, opened 2024-12-27
feat: Reduce malloc virtual memory usage

lgsilva3087:
The new option LIMIT_GLIBC_MALLOC_ARENAS allows to tune (from configuration) the number of arenas used by malloc in environments with glibc. Setting it to a small positive number like 4 or 8, reduce the virtual memory used by the application, at the cost of possible performance drop for multi-threading (due to contention).

The default is 0, which means disabled (lets glibc decide the values). The options is meant to be used in environments with limited RAM.

So far, the new variable is only used in the daemons: metadata, metalogger and chunkservers.

### DIFF
--- a/tests/test_suites/ShortSystemTests/test_limit_glibc_malloc_arenas.sh
+++ b/tests/test_suites/ShortSystemTests/test_limit_glibc_malloc_arenas.sh
@@ -1,0 +1,88 @@
+timeout_set "2 minutes"
+
+CHUNKSERVERS=3 \
+	MASTER_CUSTOM_GOALS="1 ec21: \$ec(2,1)" \
+	AUTO_SHADOW_MASTER="NO" \
+	MOUNT_EXTRA_CONFIG="sfscachemode=NEVER" \
+	CHUNKSERVER_EXTRA_CONFIG="MAGIC_DEBUG_LOG = $TEMP_DIR/log|LOG_FLUSH_ON=DEBUG" \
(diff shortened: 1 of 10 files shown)


rolysr (dismissed):
LGTM

tests/test_suites/ShortSystemTests/test_limit_glibc_malloc_arenas.sh
```diff
@@ -0,0 +1,88 @@
...
+# This restart makes the later comparison in better conditions.
+saunafs_master_daemon restart
+saunafs_wait_for_all_ready_chunkservers
+
+arenas=2
+
+### Test the limit in the first chunkserver ###
+
+# Get initial values
+cs0Pid=$(saunafs_chunkserver_daemon 0 test 2>&1 | sed 's/.*pid: //')
```
ralcolea, line 47: Consider adding `tr -d '\0'` to suppress the warning from `sed` ignoring null byte in input. 
```bash
cs0Pid=$(saunafs_chunkserver_daemon 0 test 2>&1 | tr -d '\0' | sed 's/.*pid: //')
```

ralcolea (dismissed):

tests/test_suites/ShortSystemTests/test_limit_glibc_malloc_arenas.sh
lgsilva3087: Done.

rolysr (approved):
LGTM

ralcolea (approved):

PR #268, opened 2024-12-30
feat: Speedup unlink operation for big directories

dmga44:
This PR mainly targets the unlink operation performance on big directories. The key change was to replace the O(n) [n = number of entries in the directory] previous implementation of getChildName to other one that behaves O(1) for non-hardlinked files.

There are some other side changes:
- modify the big session metadata benchmark to show results per every chunk of files (not yet configurable).
- speedup the dumping of metadata due to now creating the buffer to store nodes and edges every time. These buffers were not small.
- increase the size of the chunk hash table used in master to make their linked lists shorter. This change targets speeding up the traversing of those linked lists.

### DIFF
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -71,6 +71,9 @@ FSNode *FSNode::create(uint8_t type) {
 }
 
 void FSNode::destroy(FSNode *node) {
+	for (auto const &[_, handlePtr] : node->parent) {
+		delete handlePtr;
+	}
 	switch (node->type) {
 	case kFile:
 	case kTrash:
@@ -224,8 +227,9 @@ int fsnodes_nameisused(FSNodeDirectory *node, const HString &name) {
 
 /*! \brief Returns true iff \param ancestor is ancestor of \param node. */
 bool fsnodes_isancestor(FSNodeDirectory *ancestor, FSNode *node) {
-	for(const auto &parent_inode : node->parent) {
-		FSNodeDirectory *dir_node = fsnodes_id_to_node_verify<FSNodeDirectory>(parent_inode);
+	for (const auto &[parentId, _] : node->parent) {
+		FSNodeDirectory *dir_node =
+		    fsnodes_id_to_node_verify<FSNodeDirectory>(parentId);
 
 		while(dir_node) {
 			if (ancestor == dir_node) {
@@ -235,7 +239,8 @@ bool fsnodes_isancestor(FSNodeDirectory *ancestor, FSNode *node) {
 			assert(dir_node->parent.size() <= 1);
 
 			if (!dir_node->parent.empty()) {
-				dir_node = fsnodes_id_to_node_verify<FSNodeDirectory>(dir_node->parent[0]);
+				dir_node = fsnodes_id_to_node_verify<FSNodeDirectory>(
+				    dir_node->parent[0].first);
 			} else {
 				dir_node = nullptr;
 			}
@@ -310,7 +315,8 @@ FSNodeDirectory *fsnodes_get_first_parent(FSNode *node) {
 	assert(node);
 	FSNodeDirectory *parent;
 	if (!node->parent.empty()) {
-		parent = fsnodes_id_to_node_verify<FSNodeDirectory>(node->parent[0]);
+		parent =
+		    fsnodes_id_to_node_verify<FSNodeDirectory>(node->parent[0].first);
 	} else {
 		parent = gMetadata->root;
 	}
@@ -330,8 +336,9 @@ static inline void fsnodes_sub_stats(FSNodeDirectory *parent, statsrecord *sr) {
 		psr->size -= sr->size;
 		psr->realsize -= sr->realsize;
 		if (parent != gMetadata->root) {
-			for (auto inode : parent->parent) {
-				FSNodeDirectory *node = fsnodes_id_to_node_verify<FSNodeDirectory>(inode);
+			for (auto const &[parentId, _] : parent->parent) {
+				FSNodeDirectory *node =
+				    fsnodes_id_to_node_verify<FSNodeDirectory>(parentId);
 				fsnodes_sub_stats(node, sr);
 			}
 		}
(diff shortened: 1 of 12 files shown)


antuan96314 (dismissed):
LGTM

src/master/filesystem_node_types.h
```diff
@@ -203,17 +206,19 @@ struct FSNodeDevice : public FSNode {
...
 	};
-	typedef flat_map<hstorage::Handle, FSNode *, std::vector<std::pair<hstorage::Handle, FSNode *>>,
-	HandleCompare> EntriesContainer;
-#endif
+
+	typedef __gnu_pbds::tree<std::pair<hstorage::Handle, FSNode *>,
+	                         __gnu_pbds::null_type, HandleCompare,
+	                         __gnu_pbds::rb_tree_tag,
+	                         __gnu_pbds::tree_order_statistics_node_update>
+	    EntriesContainer;
```
lgsilva3087, line 221: How is the portability of this extensions? Are they available only for GCC or also in Clang (for instance)?

lgsilva3087 (commented):
Great job. Please see my question.

src/master/filesystem_node_types.h
dmga44: This change was moved to PR #272.

rolysr (approved):
LGTM

src/master/filesystem_node.cc
```diff
@@ -71,6 +71,9 @@ FSNode *FSNode::create(uint8_t type) {
 }
 
 void FSNode::destroy(FSNode *node) {
+	for (auto const &[_ /*: parentId*/, handlePtr] : node->parent) {
```
ralcolea: I like the idea of using the placeholder (`_`) and remove `/*: parentId/*` to be consistent with other parts of the code.

src/master/filesystem_node.cc
```diff
@@ -509,29 +519,55 @@ void fsnodes_remove_edge(uint32_t ts, FSNodeDirectory *parent, const HString &na
 	}
 
 	fsnodes_update_checksum(parent);
-
-	auto it = std::find(node->parent.begin(), node->parent.end(), parent->id);
-	if (it != node->parent.end()) {
-		node->parent.erase(it);
-	}
+	if (!parent->case_insensitive) {
```
ralcolea: Consider refactoring this code to avoid duplicated scenarios with very similar logic:
```cpp
HString currentName = name;
if (parent->case_insensitive) {
	currentName = HString::hstringToLowerCase(name);
}

auto it = std::find_if(
    node->parent.begin(), node->parent.end(),
    [parent,
     currentName](const std::pair<uint32_t, const hstorage::Handle *> &p) {
	    return p.first == parent->id && p.second->get() == currentName;
    });

if (it != node->parent.end()) { 
      node->parent.erase(it);
}
```

src/master/filesystem_node_types.h
```diff
@@ -123,8 +123,10 @@ struct FSNode {
 	uint32_t gid; /*!< Group id. */
 	uint32_t trashtime; /*!< Trash time. */
 
-	compact_vector<uint32_t, uint32_t> parent; /*!< Parent nodes ids. To reduce memory usage ids
-	                                                are stored instead of pointers to FSNode. */
+	compact_vector<std::pair<uint32_t, const hstorage::Handle *>, uint32_t>
+	    parent; /*!< Parent nodes ids + name of the of entries in those parents.
```
ralcolea: Please, fix some typos in the description of `parent` attribute.
I think the description should reference `handle` instead of `names`, even if in practice, `Handle` behaves like a name, but I think `Handle` it's more general and contains more information.

src/master/filesystem_node_types.h
```diff
@@ -203,17 +205,17 @@ struct FSNodeDevice : public FSNode {
  * Avg size (10 files) ~ 280B (28B per file)
  */
 struct FSNodeDirectory : public FSNode {
-#if defined(SAUNAFS_HAVE_64BIT_JUDY) && !defined(DISABLE_JUDY_FOR_ENTRIESCONTAINER)
-	typedef judy_map<hstorage::Handle, FSNode *> EntriesContainer;
```
ralcolea, line 207: Why is it removed `judy_map` from `FSNodeDirectory`?
What consequences can have this change in the performance of operations like insertion, deletion and lookup?
Why is it removed only from `FSNodeDirectory` and not from other data types?

src/master/filesystem_periodic.cc
```diff
@@ -417,8 +418,12 @@ void fs_process_file_test() {
 					FSNode *node = entry.second;
 
 					if (!node ||
-					    std::find(node->parent.begin(), node->parent.end(),
-					              f->id) == node->parent.end()) {
+					    std::find_if(
```
ralcolea: Consider storing the result of `std::find_if()` function in a different variable.
The mixture of `std::find_if` with the lambda and the comparison with `node->parent.end()`
is less readable and more difficult to read.

ralcolea (commented):
Very nice changes @dmga44 :+1: :fire: :muscle: 
Please, see my suggestions and questions.

src/master/filesystem_node.cc
dmga44: Please notice there is another change which you don't consider when case insensitive:
```return p.first == parent->id && HString::hstringToLowerCase(p.second->get()) == lowerCaseName;```
the ```hstringToLowerCase``` function is not applied in your proposal, and it is needed.

I think it can be applied, more or less.

src/master/filesystem_node_types.h
dmga44: Let's go one by one:
- the ```judy_map``` was removed mainly because the type of the keys in the map changed from ```Handle``` to ```Handle *```, and having it would imply coding more things. Also, the ```judy_map``` was not used in the last releases, and was substituted by the ```flat_map``` which is used here.
- the motivation of the substitution of ```judy_map``` by ```flat_map``` was increasing the performace of insertion, deletion and lookup. Also, the PR #272 replaces this data structure again.
- in the other places where the ```judy_map``` is used, I have not seen any reason to remove it or not use it. Substituting it in some places, like in the ```SparseBitset``` structure may have terrible consecuences.

src/master/filesystem_node.cc
dmga44: Applied.

src/master/filesystem_node_types.h
dmga44: Applied.

src/master/filesystem_periodic.cc
dmga44: Notice the first part of the or operator, ```!node``` ensures that when the first part is not met, and the second is ran, the node always exist. That's why it was not looking that bad before, I've uploaded a change that maybe looks better, comments added :)

src/master/filesystem_node_types.h
ralcolea: I remembered it was the other way around, I meant, judy_map was faster than flat_map, that's why my question. Your third idea is closer to what I remembered :)

ralcolea (approved):

PR #272, opened 2025-01-06
feat: Improve metadata speed for big directories

dmga44:
[Must be merged after #268 ]

This commit changes the old flat_map that was advised to use for an
ordered_statistics structure that ensures good performance in every
case. The custom flat_map structure previously used had O(n) insertion
and elimination costs, the new one has O(logn) cost for both of those
operations. The goal of this change is to speedup the metadata
operations for directories with lots of files.


src/master/filesystem_store.cc
```diff
@@ -248,20 +267,21 @@ void fs_storeedge(FSNodeDirectory *parent, FSNode *child,
 	uint8_t uedgebuff[4 + 4 + 2 + 65535];
```
uristdwarf: Check if this variable is still used

src/master/filesystem_store.cc
```diff
@@ -248,20 +267,21 @@ void fs_storeedge(FSNodeDirectory *parent, FSNode *child,
...
 	uint8_t *ptr;
 	if (child == nullptr) {  // last edge
-		memset(uedgebuff, 0, 4 + 4 + 2);
-		if (fwrite(uedgebuff, 1, 4 + 4 + 2, fd) != (size_t)(4 + 4 + 2)) {
+		memset(gEdgeStoreBuffer, 0, kEdgeHeaderSize);
+		if (fwrite(gEdgeStoreBuffer, 1, kEdgeHeaderSize, fd) !=
+		    (size_t)(kEdgeHeaderSize)) {
 			safs_pretty_syslog(LOG_NOTICE, "fwrite error");
 			return;
```
uristdwarf, line 273: It would be nice to know what kind of fwrite error, instead of just error

src/master/filesystem_store.cc
```diff
@@ -248,20 +267,21 @@ void fs_storeedge(FSNodeDirectory *parent, FSNode *child,
...
 	put32bit(&ptr, child->id);
 	put16bit(&ptr, name.length());
 	memcpy(ptr, name.c_str(), name.length());
-	if (fwrite(uedgebuff, 1, 4 + 4 + 2 + name.length(), fd) !=
-	    (size_t)(4 + 4 + 2 + name.length())) {
+	if (fwrite(gEdgeStoreBuffer, 1, kEdgeHeaderSize + name.length(), fd) !=
+	    (size_t)(kEdgeHeaderSize + name.length())) {
 		safs_pretty_syslog(LOG_NOTICE, "fwrite error");
 		return;
 	}
```
uristdwarf, line 286: Again, what kind of fwrite error?

utils/big_session_metadata_benchmark.cc
```diff
@@ -88,7 +90,7 @@ int main(int argc, char **argv) {
 	std::vector<int> testFilesFds(numberOfFiles);
 	std::vector<int> filesOrder(numberOfFiles);
 	auto start = std::chrono::high_resolution_clock::now();
-
+	auto chunkStart = start;
```
rolysr, line 93: Why not directly set the `std::chrono::high_resolution_clock::now()` to `chunkStart`?

src/master/filesystem_node.cc
```diff
@@ -1012,60 +1050,74 @@ void fsnodes_getdir(uint32_t rootinode, uint32_t uid, uint32_t gid, uint32_t aui
...
+			next_index = (*first_dirent_it).first->data() & ~SIGN_BIT_64;
 		}
 		dir_entries.emplace_back(kDotDotEntryIndex, next_index, std::move(inode), std::string(".."), std::move(attr));
 
 		first_entry = next_index;
 		--number_of_entries;
 	}
 
-	if (number_of_entries == 0) {
+	if (number_of_entries == 0 || p->entries.empty()) {
```
rolysr, line 1083: Why was this second condition (`p->entries.empty()`) added here?

src/master/filesystem_checksum.cc
```diff
@@ -41,8 +41,9 @@ static uint64_t fsnodes_checksum(FSNode *node, bool full_update = false) {
 	case FSNode::kDirectory:
 		if (full_update) {
 			static_cast<FSNodeDirectory*>(node)->entries_hash = 0;
-			for(const auto &entry : *static_cast<FSNodeDirectory*>(node)) {
-				static_cast<FSNodeDirectory*>(node)->entries_hash ^= entry.first.hash();
+			for (const auto &entry : *static_cast<FSNodeDirectory *>(node)) {
+				static_cast<FSNodeDirectory *>(node)->entries_hash ^=
+				    entry.first->hash();
```
rolysr, line 46: This change seems to be only about formatting, please consider ignoring it.

src/master/filesystem_store.cc
```diff
@@ -495,28 +515,30 @@ void fs_storenode(FSNode *f, FILE *fd) {
...
+			for (i = 0; i < kChunksBucketSize; i++) {
 				put64bit(&chptr, node_file->chunks[indx]);
 				indx++;
 			}
-			if (fwrite(ptr, 1, 8 * 65536, fd) != (size_t)(8 * 65536)) {
+			if (fwrite(ptr, 1, 8 * kChunksBucketSize, fd) !=
+			    (size_t)(8 * kChunksBucketSize)) {
 				safs_pretty_syslog(LOG_NOTICE, "fwrite error");
 				return;
 			}
```
uristdwarf, line 541: What does the 8 stand for?

rolysr (dismissed):
Very good improvement !! Just few comments/doubts below.

src/master/filesystem_store.cc
```diff
@@ -430,16 +429,20 @@ int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile, size_
 			currentParentId = parentId;
 		}
 
-		auto it = parent->entries.insert({hstorage::Handle(name), child}).first;
-		parent->entries_hash ^= (*it).first.hash();
+		hstorage::Handle *handlePtr = new hstorage::Handle(name);
+		auto it = parent->entries.insert({handlePtr, child}).first;
+		parent->entries_hash ^= (*it).first->hash();
```
uristdwarf, line 434: Why new? Why not use a a smart pointer?

uristdwarf (commented):
WIP review, some of my comments apply to #268

src/master/filesystem_store.cc
uristdwarf: In fact, it should not be LOG_NOTICE, it should be critical because fwrite failing is a system call failing, which can be very bad

utils/big_session_metadata_benchmark.cc
dmga44: Well, because ```start``` already contains that value, which is the desired one.

src/master/filesystem_node.cc
dmga44: p is the directory we're traversing to get the entries from.
p->entries is the container of pairs <name (actually ptr to handle of that name), node (actually pointer to inner representation of the node)> the directory p contains.

At that point of the execution, we have already check the trivial sons ```'.'``` and ```'..'``` that are not included in the entries container, and only the ones coming from that container should replied from that point on. So, if it is empty, then nothing else should be replied in that function.

src/master/filesystem_checksum.cc
dmga44: Please note there is a change in the body of the for loop. Though you're right about the for statement itself.

src/master/filesystem_store.cc
dmga44: It was not. It is removed.

src/master/filesystem_store.cc
dmga44: It would be nice, yeah. But IMO the change you're proposing goes beyond the scope of this PR.

src/master/filesystem_store.cc
dmga44: Same as above.

src/master/filesystem_store.cc
dmga44: It stands for the size of the data type used to store the chunk ID. In this case, it is ```uint64_t``` which is 8 bytes long.

src/master/filesystem_store.cc
dmga44: I agree it would be nice to have a shared_ptr there. But we are going to have several millions of them, at least two per inode. This makes me try to keep the overhead of those pointers as low as possible.

src/master/filesystem_node.cc
```diff
@@ -509,29 +519,55 @@ void fsnodes_remove_edge(uint32_t ts, FSNodeDirectory *parent, const HString &na
...
+			           HString::hstringToLowerCase(p.second->get()) ==
+			               lowerCaseName;
+		    });
+		if (it != node->parent.end()) {
+			node->parent.erase(it);
+		}
+	}
+	// Delete the handle after the check in the parent vector in the son is
+	// done.
+	delete handlePtrToErase;
```
uristdwarf, line 548: Hmm....This is practically the same code but for different case settings. I'm wondering if there's some better way to do this. Not in any way a deal breaker (in fact, if you can't find a way that's better, better keep it as it is rather than trying to force code deduplication), but wanted to note it down.

src/master/filesystem_node_types.h
```diff
@@ -44,10 +44,9 @@
     defined(DISABLE_JUDY_FOR_RESERVEDPATHCONTAINER)
 #include <map>
 #endif
-#if !defined(SAUNAFS_HAVE_64BIT_JUDY) || \
-    defined(DISABLE_JUDY_FOR_ENTRIESCONTAINER)
-#include "common/flat_map.h"
-#endif
+
+#include <ext/pb_ds/assoc_container.hpp>
+#include <ext/pb_ds/tree_policy.hpp>
```
uristdwarf, line 49: Could cause issues if not using libstdc++. Not a problem right now but could be later

uristdwarf (commented):
WIP review 2/3

src/master/filesystem_store.cc
uristdwarf: Makes sense, but it wasn't obvious at first sight. Perhaps like the others, it could be a named constant?

src/master/filesystem_store.cc
uristdwarf: Could a `unique_ptr` work? Pretty much the same resource/time usage as a raw pointer.

Also, it may be premature optimization, you should definitely benchmark using a raw pointer vs using a shared_ptr/unique_ptr

src/master/filesystem_store.cc
uristdwarf: I think cleaning up our code is something we should do regularly in our PR's, even if it isn't 100% related to the PR. Of course this depends on the cleanup size and how related to the PR it is. My opinion is that in this case it's very close to your other changes (which again are cleaning up the code by making it more readable) and the changes would be relatively small.

I however will leave this up to your personal judgment, it is in no way a blocker.

src/master/filesystem_store.cc
dmga44: In this case, a ```unique_ptr``` won't work. Notice the change introduces a copy of the pointer in the entries structure and in the parent structure of the son (some lines below). That's why I was mentioning shared_ptr, which should have higher overhead.

src/master/filesystem_node_types.h
dmga44: True. My idea about this is to fix it by bringing back the ```flat_map``` in those cases, but it could be done during those non-libstdc++ changes.

src/master/filesystem_node.cc
dmga44: I agree with you, it does not look very good. When writing this piece of code I thought of the lambda in the ```find_if``` function and it appears to be necessary different, unless some very refactor is applied here. The other part if the last if to remove the element pointed by ```it```, and the idea would be to declare ```it``` before the exterior ```if``` but the real type of ```it``` is quite horrible, so it does not look very good either way :(

src/master/filesystem_store.cc
dmga44: Yes, but I did not wanted to go that way because this file has this sizes hardcoded everywhere. Please note that the change I was aiming to do was to reuse the buffers, I also named the long constant sizes to be more readable because it seemed related and that's why this line got changed. The change you are proposing here drifts even further from the initial point into a general refactor.

src/master/chunks.cc
```diff
@@ -79,8 +79,10 @@
 #define MINCHUNKSLOOPCPU    10
 #define MAXCHUNKSLOOPCPU    90
 
-#define HASHSIZE 0x100000
-#define HASHPOS(chunkid) (((uint32_t)chunkid)&0xFFFFF)
+constexpr uint8_t kChunkHashBits = 22;
+constexpr int32_t kChunkHashSize = (1 << kChunkHashBits);
+constexpr uint32_t kChunkHashMask = kChunkHashSize - 1;
+#define HASHPOS(chunkid) (((uint32_t)chunkid) & kChunkHashMask)
```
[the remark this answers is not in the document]
dmga44: Indeed, applied 👍

uristdwarf (approved):
The rest looks good to me.

src/master/filesystem_store.cc
uristdwarf: >  The change you are proposing here drifts even further from the initial point into a general refactor.
Fair enough, though it's definitely needed

src/master/filesystem_node_types.h
```diff
@@ -206,15 +207,17 @@ struct FSNodeDevice : public FSNode {
...
+		                const std::pair<hstorage::Handle *, FSNode *> &b) const {
+			return std::make_pair(a.first->data(), a.second) <
+			       std::make_pair(b.first->data(), b.second);
 		}
 	};
 
-	typedef flat_map<hstorage::Handle *, FSNode *,
-	                 std::vector<std::pair<hstorage::Handle *, FSNode *>>,
-	                 HandleCompare>
+	typedef __gnu_pbds::tree<std::pair<hstorage::Handle *, FSNode *>,
```
ralcolea: Consider to use modern C++ `using` instruction instead of `typedef`  in the definition of `EntriesContainer`. 

Consider also to apply the same change in the current definition of `iterator` and `const_iterator` that also use `typedef`.

ralcolea (approved):

src/master/filesystem_node_types.h
dmga44: Applied within #291 .

ralcolea (approved):
LGTM :+1:

PR #273, opened 2025-01-06
chore: Remove unused protocol packets references

lgsilva3087:
MATOCS_CHUNKOP and CSTOMA_CHUNKOP are not longer used in the code. This commit removes their remaining references, except for the definitions.


ralcolea (approved):
LGTM :+1:

uristdwarf:
Are we sure this won't cause problems with LizardFS? Since there are still migrations going on from SaunaFS to LizardFS

rolysr (approved):
LGTM

PR #274, opened 2025-01-07
feat(mount): Add client limit glibc malloc arenas

rolysr:
This change adds the `limitglibcmallocarenas` option to the client. This options allows to reduce the virtual memory used by the client, at the cost of possible performance drop for multi-threading (due to contention).

The option is meant to be used in environments with limited RAM.

### DIFF
--- a/src/mount/sauna_client.h
+++ b/src/mount/sauna_client.h
@@ -126,7 +126,7 @@ struct FsInitParams {
 	static constexpr unsigned kDefaultAclCacheSize = 1000;
 	static constexpr bool     kDefaultVerbose = false;
 	static constexpr bool     kDirectIO = false;
-
+	static constexpr unsigned kDefaultLimitGlibcMallocArenas = 0;
 	// Thank you, GCC 4.6, for no delegating constructors
 	FsInitParams()
 	             : bind_host(), host(), port(), meta(false), mountpoint(), subfolder(kDefaultSubfolder),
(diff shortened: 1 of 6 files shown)


tests/setup_machine.sh
```diff
@@ -133,6 +133,7 @@ apt_packages=(
 	libyaml-cpp-dev
 	netcat-openbsd
 	python3-venv
+	sysstat
```
dmga44: Consider not adding this package into the required ones because we are not using it.

rolysr: Done.

src/mount/sauna_client.h
```diff
@@ -122,7 +122,9 @@ struct FsInitParams {
 	static constexpr unsigned kDefaultAclCacheSize = 1000;
 	static constexpr bool     kDefaultVerbose = false;
 	static constexpr bool     kDirectIO = false;
-
+#ifndef _WIN32
```
ralcolea: Why is it different the validation in this case? Is it really needed? Why not having for all platforms and just ignore it in platforms not supporting glibc?

tests/test_suites/ShortSystemTests/test_limit_glibc_malloc_arenas_client.sh
```diff
@@ -0,0 +1,82 @@
...
+echo "Running fio commands and measuring VSZ for both mount points"
+avg_vsz_mount0=$(run_fio_and_measure_vsz "${info[mount0]}" $pid1)
+avg_vsz_mount1=$(run_fio_and_measure_vsz "${info[mount1]}" $pid2)
+
+# Print the average VSZ values
+echo "Average VSZ for ${info[mount0]}: ${avg_vsz_mount0}"
+echo "Average VSZ for ${info[mount1]}: ${avg_vsz_mount1}"
+
+# Compare the average VSZ values in one line
+result=0
```
ralcolea: Consider this alternative for your validation:
```bash
assert_less_than "${avg_vsz_mount1}" "${avg_vsz_mount0}"
```

ralcolea (approved):
LGTM. Please, see my suggestions.

src/mount/sauna_client.h
rolysr: Done

tests/test_suites/ShortSystemTests/test_limit_glibc_malloc_arenas_client.sh
rolysr: Done

lgsilva3087 (approved):
LGTM

PR #276, opened 2025-01-07
feat(mount): Enable status updater thread option

rolysr:
This commit adds an option to enable the status updater thread in the Windows client side in order to manage when to use it or not. The change only affects the Windows client side.


lgsilva3087 (approved):

dmga44 (approved):

ralcolea (approved):
LGTM

PR #279, opened 2025-01-10
test: Use clang19 for running test_build_saunafs_using_clang

uristdwarf:
Earlier versions of clang had trouble compiling due to the __cp_concepts
value being lower 202002L. This was updated in clang19, however most
major distributions do not ship it yet.

The commit uses a script to setup repositories for llvm toolchains and
installs the latest version from clang.

See here for more information: https://github.com/clangd/clangd/issues/2200


dmga44 (approved):

rolysr (approved):
LGTM

PR #280, opened 2025-01-12
feat(mount): add log on Windows notification area

rolysr:
This change adds a log on Windows notification area, specifically for the error associated with the writing process, which allows the user to have a proper feedback on the error on the Windows environment. This is a Windows-specific feature.

### DIFF
--- a/src/mount/fuse/main.cc
+++ b/src/mount/fuse/main.cc
@@ -261,6 +261,8 @@ static int mainloop(struct fuse_args *args, struct fuse_cmdline_opts *fuse_opts,
 	params.debug_mode = gMountOptions.debug;
 	params.direct_io = gMountOptions.directio;
 	params.ignore_flush = gMountOptions.ignoreflush;
+	params.log_notifications_area = gMountOptions.lognotificationarea;
+	params.message_suppression_period = gMountOptions.messagesuppressionperiod;
 
 	if (!gMountOptions.meta) {
 		SaunaClient::fs_init(params);
--- a/src/mount/fuse/mount_config.cc
+++ b/src/mount/fuse/mount_config.cc
@@ -94,6 +94,8 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("nostdmountoptions", nostdmountoptions, 1),
 	SFS_OPT("sfsignoreflush=%d", ignoreflush, 0),
 	SFS_OPT("limitglibcmallocarenas=%d", limitglibcmallocarenas, 0),
+	SFS_OPT("sfslognotificationarea=%d", lognotificationarea, 0),
+	SFS_OPT("sfsmessagesuppressionperiod=%u", messagesuppressionperiod, 0),
 
 	SFS_OPT("enablefilelocks=%u", filelocks, 0),
 	SFS_OPT("nonempty", nonemptymount, 1),
@@ -239,6 +241,9 @@ void usage(const char *progname) {
 "    -o limitglibcmallocarenas=N  limit glibc malloc arenas to given value - prevents "
 				"from using huge amount of virtual memory. Use it in constrained memory "
 				"environments (default: %u)\n"
+"    -o sfslognotificationarea=0|1  enable/disable logging to Linux notification area (default: %d)\n"
+"    -o sfsmessagesuppressionperiod=N  set period of message suppression in seconds for logging on "
+				"notification area (default: %u)\n"
 "\n",
 		SaunaClient::FsInitParams::kDefaultCacheExpirationTime,
 		SaunaClient::FsInitParams::kDefaultReadaheadMaxWindowSize,
@@ -269,7 +274,9 @@ void usage(const char *progname) {
 		SaunaClient::FsInitParams::kDefaultIoRetries,
 		SaunaClient::FsInitParams::kDefaultSymlinkCacheTimeout,
 		SaunaClient::FsInitParams::kDefaultSubfolder,
-		SaunaClient::FsInitParams::kDefaultLimitGlibcMallocArenas
+		SaunaClient::FsInitParams::kDefaultLimitGlibcMallocArenas,
+		SaunaClient::FsInitParams::kDefaultLogNotificationArea,
+		SaunaClient::FsInitParams::kDefaultMessageSuppressionPeriod
 	);
 	printf(
 "CMODE can be set to:\n"
--- a/src/mount/notification_area_logging.h
+++ b/src/mount/notification_area_logging.h
@@ -1,0 +1,189 @@
+/*
+ Copyright 2023      Leil Storage OÜ
+
+ SaunaFS is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, version 3.
+
+ SaunaFS is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with SaunaFS  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "common/platform.h"
+
+#ifdef _WIN32
+#include <shellapi.h>
+#include <windows.h>
+#include <string>
+#else
+#include <cstdlib>
+#endif
+
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+#include <string>
+#include <thread>
+#include <unordered_map>
+
+inline bool gShowMessagesOnNotificationArea = false;
+
+struct MessageCache {
+	std::string message;
+	std::chrono::steady_clock::time_point timestamp;
+};
+
+inline std::jthread notificationThread;
+
+inline std::unordered_map<std::string, MessageCache> messageCache;
+inline std::map<int, MessageCache> fullPathFromInodeCache;
+inline std::chrono::seconds gMessageSuppressionPeriodSeconds;
+inline std::mutex messageCacheMutex;
+inline std::condition_variable cv;
+inline std::atomic<bool> stopThread = false;
+
+#ifdef _WIN32
+inline void ShowWindowsNotification(const std::string &message) {
+	// Create a hidden window
+	HWND hWnd = CreateWindowEx(0, "STATIC", "HiddenWindow", 0, 0, 0, 0, 0,
+	                           HWND_MESSAGE, NULL, NULL, NULL);
+
+	// Define the NOTIFYICONDATA structure
+	NOTIFYICONDATA nid = {};
+	nid.cbSize = sizeof(NOTIFYICONDATA);
+	nid.hWnd = hWnd;                   // Handle to the hidden window
+	nid.uID = 1001;                    // Unique ID for the notification icon
+	nid.uFlags = NIF_INFO | NIF_ICON;  // Display an info balloon and an icon
+	nid.dwInfoFlags = NIIF_INFO;       // Info icon
+	strcpy_s(nid.szInfo, message.c_str());                // Set the message
+	strcpy_s(nid.szInfoTitle, "SaunaFS Windows Client");  // Set the title
+	nid.hIcon = LoadIcon(NULL, IDI_INFORMATION);  // Load a standard icon
+
+	// Display the notification
+	if (!Shell_NotifyIcon(NIM_ADD, &nid)) {
+		// Handle error
+		safs::log_debug(
+		    "ShowMessageOnWindowsNotificationArea: Failed to "
+		    "display notification icon");
+	}
+
+	// Clean up
+	DestroyWindow(hWnd);
+}
+#else
+inline void ShowLinuxNotification(const std::string &message) {
+	std::string command =
+	    R"(notify-send "SaunaFS Linux Client" ")" + message + "\"";
+
+	if (std::system(command.c_str()) != 0) {
+		safs::log_debug(
+		    "addNotificationMessage: Failed to show notification using "
+		    "notify-send");
+	}
+}
+#endif
+
+inline void NotificationThread() {
+	while (!stopThread) {
+		std::unique_lock<std::mutex> lock(messageCacheMutex);
+		cv.wait(lock, [] { return !messageCache.empty() || stopThread; });
+
+		if (stopThread) { break; }
+
+		auto now = std::chrono::steady_clock::now();
+		for (auto it = messageCache.begin(); it != messageCache.end();) {
+			if ((now - it->second.timestamp) >=
+			    gMessageSuppressionPeriodSeconds) {
+				// Show the message
+#ifdef _WIN32
+				ShowWindowsNotification(it->second.message);
+#else
+				ShowLinuxNotification(it->second.message);
+#endif
+				it = messageCache.erase(it);
+			} else {
+				++it;
+			}
+		}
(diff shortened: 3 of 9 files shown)


src/mount/fuse/main.cc
```diff
@@ -175,6 +176,12 @@ static int mainloop(struct fuse_args *args, struct fuse_cmdline_opts *fuse_opts,
 	const char *mountpoint = fuse_opts->mountpoint;
 	bool multithread = !fuse_opts->singlethread;
 	bool foreground = fuse_opts->foreground;
+	gShowMessagesOnNotificationArea = gMountOptions.lognotificationarea;
+	messageSuppressionPeriod = gMountOptions.messagesuppressionperiod;
+
+	if (gShowMessagesOnNotificationArea && messageSuppressionPeriod > 0) {
+		StartNotificationThread();
+	}
```
dmga44: Please notice this code could be moved to notification_area_logging file, removing the dependancy (the include) and simplifying the platform specific code.

The two parameters could be passed in the params, and inside the fs_init call something like notifications_init or something like that.

src/mount/readdata.cc
```diff
@@ -732,6 +733,14 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
 				throw RecoverableReadException(
 				    "Not enough read cache memory available for reading");
+				if (gShowMessagesOnNotificationArea &&
+				    messageSuppressionPeriod > 0) {
+					std::string errorInodeFullPath = "";
+					fs_fullpath(rrec->inode, 0, 0, errorInodeFullPath);
+					ShowMessageOnNotificationArea(
+					    "Not enough read cache memory available for reading " +
+					    errorInodeFullPath);
+				}
```
dmga44: A small comment here, this code is after a ```throw``` statement, so it should be inaccesible.

Anyway, I'd not log any errors in this point.

src/mount/readdata.cc
```diff
@@ -754,6 +763,14 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
 			print_error_msg(reader, try_counter, ex);
 			std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
+			if (gShowMessagesOnNotificationArea &&
+			    messageSuppressionPeriod > 0) {
+				std::string errorInodeFullPath = "";
+				fs_fullpath(rrec->inode, 0, 0, errorInodeFullPath);
+				ShowMessageOnNotificationArea(
+				    "Unrecoverable read error: " + std::string(ex.what()) +
+				    " " + errorInodeFullPath.c_str());
+			}
```
dmga44: A general comment on the read errors logging: I think a function ```ShowReadErrorMessageOnNotificationArea``` could be in line to reduce the analogous code.

Also, considering the parameters passed to that function, you could save some calls to ```fs_fullpath``` which is an expensive funcion.

src/mount/readdata.cc
```diff
@@ -768,11 +785,27 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
 			if (try_counter > maxRetries) {
 				std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
+				if (gShowMessagesOnNotificationArea &&
+				    messageSuppressionPeriod > 0) {
+					std::string errorInodeFullPath = "";
+					fs_fullpath(rrec->inode, 0, 0, errorInodeFullPath);
+					ShowMessageOnNotificationArea(
+					    "Read error: " + std::string(ex.what()) + " " +
+					    errorInodeFullPath.c_str());
+				}
```
dmga44: Another general and very important comment is that all these logging must not be inside the ```usedMemoryLock```, it could kill performance on stressfull scenarios..

src/mount/readdata.cc
```diff
@@ -768,11 +785,27 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
 			} else {
 				std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
+				if (gShowMessagesOnNotificationArea &&
+				    messageSuppressionPeriod > 0) {
+					std::string errorInodeFullPath = "";
+					fs_fullpath(rrec->inode, 0, 0, errorInodeFullPath);
+					ShowMessageOnNotificationArea(
+					    "Read error: " + std::string(ex.what()) + " " +
+					    errorInodeFullPath.c_str());
+				}
```
dmga44, line 787: I'd also add another condition to log this error here, something like ```try_counter > 9```. This way the errors from a single file would only appear one second from each other.

Recall the first few retries happen a few milliseconds from each other.

src/mount/writedata.cc
```diff
@@ -486,6 +487,16 @@ void InodeChunkWriter::processJob(inodedata* inodeData) {
...
+
+			if (gShowMessagesOnNotificationArea &&
+			    messageSuppressionPeriod > 0) {
+				std::string errorInodeFullPath = "";
+				fs_fullpath(inodeData_->inode, 0, 0, errorInodeFullPath);
+				ShowMessageOnNotificationArea(
+				    std::string(e.what()) +
+				    " from location: " + errorInodeFullPath);
+			}
+
```
dmga44: Like the read errors, you could use a very similar funcion to reduce the analogous code.

Also, you could add another condition like, ```trycnt > 4``` to don't bother for the first few tries, and also remove the locked chunk case.

dmga44: It would be also nice to add another notification at the start of the catch statement of the ```UnrecoverableWriteException``` (line 509).

src/mount/fuse/main.cc
rolysr: Done

src/mount/readdata.cc

src/mount/readdata.cc

src/mount/writedata.cc
rolysr: Done

src/mount/readdata.cc
rolysr: Done

src/mount/readdata.cc

src/mount/notification_area_logging.h
```diff
@@ -0,0 +1,170 @@
...
+	while (!stopThread) {
+		std::unique_lock<std::mutex> lock(messageCacheMutex);
+		cv.wait(lock, [] { return !messageCache.empty() || stopThread; });
+
+		if (stopThread) break;
+
+		auto now = std::chrono::steady_clock::now();
+		for (auto it = messageCache.begin(); it != messageCache.end();) {
+			if ((now - it->second.timestamp) >=
+			    std::chrono::seconds(gMessageSuppressionPeriod)) {
```
ralcolea: The value `std::chrono::seconds(gMessageSuppressionPeriod)` is constant, so it can be cached after every update of `gMessageSuppressionPeriod` and avoid unneeded call to `std::chrono::seconds()`.

src/mount/notification_area_logging.h
```diff
@@ -0,0 +1,170 @@
...
+#endif
+				it = messageCache.erase(it);
+			} else {
+				++it;
+			}
+		}
+	}
+}
+
+inline void ShowMessageOnNotificationArea(const std::string &message) {
```
ralcolea: This function does not show the message, it only add/updates the message in the cache.
Maybe the function can be renamed to `addNotificationMessage()` or similar.

src/mount/notification_area_logging.h
```diff
@@ -0,0 +1,170 @@
...
+	if (stopThread) {
+		return;
+	}
+	auto now = std::chrono::steady_clock::now();
+
+	std::unique_lock<std::mutex> lock(messageCacheMutex);
+	auto it = messageCache.find(message);
+	if (it != messageCache.end() &&
+	    (now - it->second.timestamp) <
+	        std::chrono::seconds(gMessageSuppressionPeriod)) {
```
ralcolea: Consider use the cached value to determine the number of seconds.

src/mount/notification_area_logging.h
```diff
@@ -0,0 +1,170 @@
...
+		    "modify notification icon");
+	}
+
+	// Clean up
+	DestroyWindow(hWnd);
+}
+#else
+inline void ShowLinuxNotification(const std::string &message) {
+	std::string command =
+	    "notify-send \"SaunaFS Linux Client\" \"" + message + "\"";
```
ralcolea: Consider using a Raw String Literals to avoid escaping special characters and improve the readability:
```suggestion
	    R"(notify-send "SaunaFS Linux Client" ")" + message + "\"";
```

src/mount/readdata.cc
```diff
@@ -753,6 +756,14 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
 			try_counter = 0;
 		} catch (UnrecoverableReadException &ex) {
 			print_error_msg(reader, try_counter, ex);
+			if (gShowMessagesOnNotificationArea) {
+				if (errorInodeFullPath.empty()) {
+					fs_fullpath(rrec->inode, 0, 0, errorInodeFullPath);
+				}
+				ShowMessageOnNotificationArea(
+				    "Unrecoverable read error: " + std::string(ex.what()) +
+				    " " + errorInodeFullPath.c_str());
```
ralcolea: The function `c_str()` is redundant in this case:
```suggestion
				    " " + errorInodeFullPath);
```

src/mount/readdata.cc
```diff
@@ -762,16 +773,31 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
 			if (try_counter > 0) {
 				print_error_msg(reader, try_counter, ex);
 			}
 			force_prepare = true;
 			if (try_counter > maxRetries) {
+				if (gShowMessagesOnNotificationArea) {
+					ShowMessageOnNotificationArea(
+					    "Read error: Exceeded max retries:" +
+					    std::string(ex.what()) + " " +
+					    errorInodeFullPath.c_str());
```
ralcolea: The function c_str() is redundant in this case:
```suggestion
					    errorInodeFullPath);
```

src/mount/readdata.cc
```diff
@@ -762,16 +773,31 @@ int read_to_buffer(ReadRecord *rrec, uint64_t current_offset,
...
 				std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
 				decreaseUsedReadCacheMemory(total_read_cache_bytes_to_reserve);
 				usedMemoryLock.unlock();
 				return SAUNAFS_ERROR_IO;
 			} else {
+				if (gShowMessagesOnNotificationArea &&
+				    try_counter > kMinTryCounterToShowReadErrorMessage) {
+					ShowMessageOnNotificationArea(
+					    "Read error: " + std::string(ex.what()) + " " +
+					    errorInodeFullPath.c_str());
```
ralcolea: The function c_str() is redundant in this case:
```suggestion
					    errorInodeFullPath);
```

src/mount/sauna_client.h
```diff
@@ -162,9 +164,10 @@ struct FsInitParams {
 	             enable_status_updater_thread(kDefaultEnableStatusUpdaterThread),
 	             ignore_utimens_update(kDefaultIgnoreUtimensUpdate),
 #endif
-	             ignore_flush(kDefaultIgnoreFlush), verbose(kDefaultVerbose), direct_io(kDirectIO)
-	             ,limit_glibc_malloc_arenas(kDefaultLimitGlibcMallocArenas) 
-				 {	
+	             ignore_flush(kDefaultIgnoreFlush), verbose(kDefaultVerbose), direct_io(kDirectIO),
+	             limit_glibc_malloc_arenas(kDefaultLimitGlibcMallocArenas),
+	             log_notifications_area(kDefaultLogNotificationArea), message_suppression_period(kDefaultMessageSuppressionPeriod) {
```
ralcolea: This line is too long. Consider using one line for each variable initialization.

src/mount/sauna_client.h
```diff
@@ -201,11 +204,10 @@ struct FsInitParams {
...
 	             ignore_utimens_update(kDefaultIgnoreUtimensUpdate),
 #endif
-	             ignore_flush(kDefaultIgnoreFlush), verbose(kDefaultVerbose), direct_io(kDirectIO) 
-#ifndef _WIN32
-	             ,limit_glibc_malloc_arenas(kDefaultLimitGlibcMallocArenas) 
-#endif 
-				 {
+	             ignore_flush(kDefaultIgnoreFlush), verbose(kDefaultVerbose), direct_io(kDirectIO),
+	             limit_glibc_malloc_arenas(kDefaultLimitGlibcMallocArenas),
+	             log_notifications_area(kDefaultLogNotificationArea), message_suppression_period(kDefaultMessageSuppressionPeriod) {
```
ralcolea: This line is too long. Consider using one line for each variable initialization.

src/mount/writedata.cc
```diff
@@ -391,11 +392,15 @@ class InodeChunkWriter {
...
 	static const uint32_t kTimeToFinishOperations = 5;
+	// Minimum tries counter value before showing write error message on
+	// notifications area
+	static const uint32_t kMinTryCounterToShowWriteErrorMessage = 4;
 };
 
 void InodeChunkWriter::processJob(inodedata* inodeData) {
 	LOG_AVG_TILL_END_OF_SCOPE0("InodeChunkWriter::processJob");
 	inodeData_ = inodeData;
+	std::string errorInodeFullPath = "";
```
ralcolea: Initializing a string variable to the empty string is redundant.

ralcolea (commented):
Good job @rolysr. Please, see my suggestions.

src/mount/notification_area_logging.h
rolysr: Done

src/mount/notification_area_logging.h
rolysr: Done

src/mount/notification_area_logging.h
rolysr: Done

src/mount/notification_area_logging.h
rolysr: Done

src/mount/readdata.cc

src/mount/readdata.cc

src/mount/sauna_client.h
rolysr: Done

src/mount/writedata.cc
rolysr: Done

src/mount/sauna_client.h
rolysr: Done

ralcolea (approved):
LGTM :+1:

src/mount/fuse/mount_config.cc
```diff
@@ -94,6 +94,8 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("nostdmountoptions", nostdmountoptions, 1),
 	SFS_OPT("sfsignoreflush=%d", ignoreflush, 0),
 	SFS_OPT("limitglibcmallocarenas=%d", limitglibcmallocarenas, 0),
+	SFS_OPT("sfslognotificationarea=%s", lognotificationarea, 0),
```
dmga44: Is it not suppose to be ```sfslognotificationarea=%d```(like ```sfsignoreflush``` option, for instance) ? I think it means you are parsing a string.

dmga44 (approved):
Nice job 👍

src/mount/fuse/mount_config.cc

ralcolea (approved):
Great job @rolysr :+1: :muscle: :fire:

PR #286, opened 2025-01-22
fix: Fix CLang compilation

dmga44:
CLang compiler was detecting one of the ```find``` functions in the
```FSNodeDirectory``` as recursive, which was causing the CLang build
to fail. This commit removes that function, which was not used in
the rest of the code and it is also simple.


lgsilva3087 (approved):

ralcolea (approved):
LGTM :+1:

PR #289, opened 2025-01-23
test: Fix glibc malloc arenas option on client

rolysr:
This commit moves the test for checking the tuning of the glibc malloc arenas option on the client to the SingleMachineTests suite to avoid the nondeterministic behavior of the test in the continuous integration for the ShortSystemTests.


uristdwarf (approved):

lgsilva3087 (approved):
LGTM

PR #290, opened 2025-01-24
feat(chunkserver): Reduce strings RAM usage

lgsilva3087:
The previous version was caching the complete paths to the meta and data parts for each chunk, consuming a considerable amount of RAM.

This change only caches the basenames and uses the owner's meta and data paths to construct the final path.


ralcolea (approved):

dmga44 (approved):
LGTM.

PR #292, opened 2025-01-28
chore(master): Update includes

lgsilva3087:
Previously, there were too many unused includes.

This commit removes the unneeded includes in the master files.
The includes were also sorted according clang-format, except for:

- common/platform.h is the first and should be in a separate group.
- If it is a .cc with associated .h, then the .h follows
  common/platform.h. Also in a separate group.
- Then the system libraries/dependencies (sorted by clang-format).
- Then the saunafs includes.

### DIFF
--- a/src/main/main.cc
+++ b/src/main/main.cc
@@ -19,18 +19,13 @@
  */
 
 #include "common/platform.h"
+
 #include "init.h"
 
-#include <errno.h>
 #include <fcntl.h>
 #include <grp.h>
-#include <ios>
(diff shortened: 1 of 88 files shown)


rolysr (approved):
Good change. Please consider on next steps using the `include-what-you-use` from `vcpkg` tool so that we can integrate it on our compilation process similar to this: `set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE include-what-you-use)`.

ralcolea (approved):

PR #293, opened 2025-01-29
fix: Fixes on sfsappendchunks + test

dmga44:
When appending chunks of relatively big files to some destination file
the resulting size could be wrong. This issue was caused by an overflow
when calculating new size of the destination file.

The previous version of the code was using a legacy (C-like) approach
to enlarge the chunk list of the resulting file of appending chunks. It
was increasing one by one when too small, then eight by eight then 64
at a time, and having at as many as needed.

This approach had a downside, which is that the real number of chunks
of that file was not shown correctly when using fileinfo tool, and
extra chunks were appearing without data, which might be confusing. It
was also not fulfilling any purpose, motivation for that change was
reducing the malloc calls when requesting memory for new chunks, but
the vector's resize manages to perform this already quite efficient.

This PR also improves sfstools test to check the prevous two issues.


rolysr (approved):
LGTM

lgsilva3087 (approved):
LGTM

PR #295, opened 2025-02-05
feat: Use quota in volume size option

dmga44:
The following changes are proposed:
- introduce the option ```usequotainvolumesize``` to display the user specific quota instead of the storage total size. This option is also added to tweaks hidden file.
- introduce the option ```statfscachetimeout``` to tune the statfs output caching, it is disabled by default. This option is also added to tweaks hidden file.
- add a test to check the behavior of the option.
- a small refactor in sauna_client.h (a new option is added and it would be nice to pack it together).

### DIFF
--- a/src/mount/fuse/mount_config.cc
+++ b/src/mount/fuse/mount_config.cc
@@ -96,6 +96,7 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("limitglibcmallocarenas=%d", limitglibcmallocarenas, 0),
 	SFS_OPT("sfslognotificationarea=%d", lognotificationarea, 0),
 	SFS_OPT("sfsmessagesuppressionperiod=%u", messagesuppressionperiod, 0),
+	SFS_OPT("usequotainvolumesize=%d", usequotainvolumesize, 0),
 
 	SFS_OPT("enablefilelocks=%u", filelocks, 0),
 	SFS_OPT("nonempty", nonemptymount, 1),
@@ -244,6 +245,8 @@ void usage(const char *progname) {
 "    -o sfslognotificationarea=0|1  enable/disable logging to Linux notification area (default: %d)\n"
 "    -o sfsmessagesuppressionperiod=N  set period of message suppression in seconds for logging on "
 				"notification area (default: %u)\n"
+"    -o usequotainvolumesize=0|1  when set to 1, use the user and group specific quota hard "
+				"limit to calculate the volume size (default: %u)\n"
 "\n",
 		SaunaClient::FsInitParams::kDefaultCacheExpirationTime,
 		SaunaClient::FsInitParams::kDefaultReadaheadMaxWindowSize,
@@ -276,7 +279,8 @@ void usage(const char *progname) {
 		SaunaClient::FsInitParams::kDefaultSubfolder,
 		SaunaClient::FsInitParams::kDefaultLimitGlibcMallocArenas,
 		SaunaClient::FsInitParams::kDefaultLogNotificationArea,
-		SaunaClient::FsInitParams::kDefaultMessageSuppressionPeriod
+		SaunaClient::FsInitParams::kDefaultMessageSuppressionPeriod,
+		SaunaClient::FsInitParams::kDefaultUseQuotaInVolumeSize
 	);
 	printf(
 "CMODE can be set to:\n"
--- a/src/mount/fuse/mount_config.cc
+++ b/src/mount/fuse/mount_config.cc
@@ -96,6 +96,7 @@ struct fuse_opt gSfsOptsStage2[] = {
 	SFS_OPT("limitglibcmallocarenas=%d", limitglibcmallocarenas, 0),
 	SFS_OPT("sfslognotificationarea=%d", lognotificationarea, 0),
 	SFS_OPT("sfsmessagesuppressionperiod=%u", messagesuppressionperiod, 0),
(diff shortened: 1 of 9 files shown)


tests/test_suites/ShortSystemTests/test_use_quota_in_volume_size.sh
```diff
@@ -0,0 +1,35 @@
+# Test will only succeed if the available space is more than 1GiB
+#
+# To run this test you need to add the following lines to /etc/sudoers.d/saunafstest:
+#
+# saunafstest ALL = NOPASSWD: /usr/bin/tee /tmp/SaunaFS-autotests/mnt/sfs*
+#
+#
```
lgsilva3087: Consider adding this to the setup machine script.

src/mount/fuse/mount_config.cc
```diff
@@ -239,6 +240,8 @@ void usage(const char *progname) {
 "    -o limitglibcmallocarenas=N  limit glibc malloc arenas to given value - prevents "
 				"from using huge amount of virtual memory. Use it in constrained memory "
 				"environments (default: %u)\n"
+"    -o usequotainvolumesize=0|1  replace normal volume size with the space hard limit "
+				"of the user specific quota (if assigned) when set to 1 (default: %u)\n"
```
lgsilva3087: Consider:
when set to 1, use the user-specific quota hard limit as the volume size.

lgsilva3087: Does it consider quota for groups?

src/mount/sauna_client.cc
```diff
@@ -687,6 +688,34 @@ struct statvfs statfs(Context &ctx, Inode ino) {
 	}
 	(void)ino;
 	fs_statfs(&totalspace,&availspace,&trashspace,&reservedspace,&inodes);
+	if (gUseQuotaInVolumeSize) {
```
lgsilva3087, line 712: Applications can trigger a surprisingly high number of statfs (even bigger in Windows). Consider to cache the final quota value for a short time (1-5 seconds).

lgsilva3087 (approved):
Looks good, but please see my comments.

tests/test_suites/ShortSystemTests/test_use_quota_in_volume_size.sh
```diff
@@ -0,0 +1,35 @@
...
+CHUNKSERVERS=1 \
+	USE_RAMDISK=YES \
+	SFSEXPORTS_EXTRA_OPTIONS="allcanchangequota" \
+	MOUNT_EXTRA_CONFIG="usequotainvolumesize=1" \
+	setup_local_empty_saunafs info
+
+cd "${info[mount0]}"
+hard_limit=$((1024*1024*1024))
+initial_size=$(df --block-size=1 | grep "${info[mount0]}" | awk '{print $2}')
+[[ ${initial_size} -ne ${hard_limit} ]] && echo "OK: the available space is different than 1GiB"
```
rolysr: As this test is supposed to pass when available space is greater than 1GiB, then you can consider using `assert_greater_than` for better readability

rolysr (approved):
LGTM

tests/test_suites/ShortSystemTests/test_use_quota_in_volume_size.sh
dmga44: This was copied from another test, so it should be there (the ```tee``` line in sudoers.d).

src/mount/fuse/mount_config.cc
dmga44: Applied.

dmga44: I considered it, even coded the idea but the context coming from fuse layer passes a wrong ```gid```. Have no idea why.

src/mount/sauna_client.cc
dmga44: Well, I gave it some thought and the idea could be extended to the whole statfs output because the quota result (specially user's used space) could change at the same pace of the whole storage used space.

I have the cache for the whole statfs and an option to tune it, and also added it to the tweaks, so it could be activated in runtime.

tests/test_suites/ShortSystemTests/test_use_quota_in_volume_size.sh
dmga44: Applied.

tests/test_suites/ShortSystemTests/test_use_quota_in_volume_size.sh
lgsilva3087: In this case, we can remove the comments from the test.

src/mount/fuse/mount_config.cc
lgsilva3087: Please check about secondary groups. Specifically this constant kSecondaryGroupsBit.

src/mount/sauna_client.cc
lgsilva3087: Great.

tests/test_suites/ShortSystemTests/test_use_quota_in_volume_size.sh
dmga44: Removed.

src/mount/fuse/mount_config.cc
dmga44: Thanks.

I checked that only the main group of the user was the one affected, so I only asked for the main group.

PR #296, opened 2025-02-09
feat(mds,common): Start decoupling metadata backend

lgsilva3087:
The metadata persistency layer (or backend) is hard-coded to use only one way, mostly based on the metadata.sfs file. Implementing other backends (like a database) is extremely complex, due to the internal dependencies on this fixed format.

This commit starts a series of commits to decouple the metadata persistency layer from the unique current implementation. The goal is to make easy the implementation of new approaches.

The main changes of this specific commit are:
- Create the IMetadataBackend interface to provide the extension point.
- Create the MetadataBackendFile as concrete implementation, containing the current behavior.
- Remove the files filesystem_store.h and filesystem_store.cc from master (mostly moved to the new files).
- Remove the metadata.h and metadata.cc files from common, they were doing a job very specific to metadata servers.
- Update all modules to use the new gMetadataBackend instance (this will hold the concrete backend when available).

This change involves modifications in master, shadows, metaloggers and metarestore. Other commits will complement this one.


uristdwarf:
My current comment about this is that we should _probably_ get rid of the metalogger ifdefines, and instead have the metalogger code completely separate and not related to master at all.

I'm not sure exactly how to do this, but one way is to copy changelog.cc and masterconn.cc to metalogger and remove all the ifdefines in those files. This will result in duplicate code, but we can then try to de-duplicate this code into a library which both master and metalogger can be built with. The other idea is to rewrite metalogger.

Same thing with metarestore, we should strive to separate code that both master and metarestore use into specific libraries that both can use. This makes it easier to maintain both metarestore, metalogger and master all separately .

So IMO, the metalogger and metarestore preferably should not be part of the interface at all, unless we will have different changelogs for different metadata.

lgsilva3087:
> My current comment about this is that we should _probably_ get rid of the metalogger ifdefines, and instead have the metalogger code completely separate and not related to master at all.
> 
> I'm not sure exactly how to do this, but one way is to copy changelog.cc and masterconn.cc to metalogger and remove all the ifdefines in those files. This will result in duplicate code, but we can then try to de-duplicate this code into a library which both master and metalogger can be built with. The other idea is to rewrite metalogger.
> 
> Same thing with metarestore, we should strive to separate code that both master and metarestore use into specific libraries that both can use. This makes it easier to maintain both metarestore, metalogger and master all separately .
> 
> So IMO, the metalogger and metarestore preferably should not be part of the interface at all, unless we will have different changelogs for different metadata.

Hi Urmas, I was thinking the same, but it will be a step-by-step process. I am still identifying the pieces.

src/metalogger/init.h
```diff
@@ -24,13 +24,28 @@
 
 #include "common/run_tab.h"
 #include "master/masterconn.h"
+#include <master/metadata_backend_file.h>
```
dmga44: Though I think it looks better with the angulars, I thought the rule for this commit was to keep the quotes if were used before.

dmga44 (commented):
Great work! 👍 

I added a small comment.

antuan96314 (approved):
LGTM

src/metalogger/init.h
lgsilva3087: Thanks, done.

antuan96314 (approved):

dmga44 (approved):
LGTM.

PR #297, opened 2025-02-10
perf(uraft): Reduce CPU usage on non-leader nodes

ralcolea:
This commit ensures that periodic validation of floating IP status is performed only on the leader node, reducing CPU usage on non-leader nodes.


uristdwarf (approved):

lgsilva3087 (approved):
LGTM. Consider reducing the frequency for checking the floating IP.

PR #298, opened 2025-02-12
Add saunafs command on Windows

rolysr:
This PR is related to the corresponding integration of the saunafs tool on a Windows environment. The main changes were:

- Fix related dependencies and added Windows specific code to make the tools submodule to be compiled on Windows.
- Fix the `open_master_conn` function in order to correctly get the corresponding inode from a given Windows file system path.
- Add Windows specific code for making the commands `makesnapshot` and `rremove` to work on Windows.
- Fix saunafs command for `sfsmountingsubfolder` on the Windows client.
- Integrate saunafs.exe on Windows client related tests.


src/common/quota_database.h
dmga44, line 1: I think you should also move the file ```quota_database.cc``` file to make the change more consistent.

src/common/signal_handling.h
```diff
@@ -0,0 +1,84 @@
...
+
+// Define signal constants for Windows
+#define SIGINT 2
+#define SIGTERM 15
+#define SIGHUP 1
+#define SIGUSR1 10
+
+#else
+#include <signal.h>
+#endif
```
dmga44: Please notice the missing endline.

src/master/filesystem_quota.h
```diff
@@ -25,6 +25,8 @@
 #include <utility>
 #include "protocol/quota.h"
 #include <master/filesystem_node_types.h>
+#include "common/quota_database.h"
+#include "master/filesystem_freenode.h"
```
dmga44, line 28: I think we should come to an agreement on whether to use quotes or angular in these kind of includes.

src/master/filesystem_operations.cc
```diff
@@ -507,6 +507,34 @@ uint8_t fs_full_path_by_inode(const FsContext &context, uint32_t initial_inode,
...
+		} else {
+			current_name += path[name_position];
+		}
+		name_position++;
+	}
+
+	inode = current_inode;
+	return SAUNAFS_STATUS_OK;
+}
+
```
dmga44: I think there are some permission checks missing here.

You could think of this function as several ```fs_lookup``` calls one on top of the other. There are things which are there that are not  present here, to mention some:
- call to ```verify_session```
- call to ```fsnodes_get_node_for_operation``` (I think this is the one doing the permission check)
- checks on the implicit entries (```.``` and ```..```)
- call to ```fsnodes_namecheck```
- entry in ```gFsStatsArray``` and increasing its count.

Consider the approach to partition the path and call severall ```fs_lookup``` (ignoring the ```gFsStatsArray``` update).

Also notice that some of these add-ups could be applied in ```fs_full_path_by_inode```.

src/mount/client/saunafs_c_api.cc
```diff
@@ -146,7 +146,14 @@ const char *sau_error_string(sau_err_t saunafs_error_code) {
...
 	try {
-		Client::Context *ret = new Client::Context(getuid(), getgid(),
+#ifdef _WIN32
+		uint32_t uid = 0;
+		uint32_t gid = 0;
+#else
+		uint32_t uid = getuid();
+		uint32_t gid = getgid();
+#endif
+		Client::Context *ret = new Client::Context(uid, gid,
```
dmga44: Is this change required? I don't think we are building the this client in Windows.

src/tools/CMakeLists.txt
```diff
@@ -30,8 +30,16 @@ install(FILES sfstools.sh DESTINATION ${BIN_SUBDIR} PERMISSIONS OWNER_EXECUTE OW
...
+        COMMAND ${CMAKE_COMMAND} -E create_symlink sfstools.sh ${LINK}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+    else()
+        # Unix-like systems command to create a symbolic link
+        add_custom_target(${LINK} ALL
+            COMMAND ln -sf sfstools.sh ${LINK}
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+    endif()
+    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${LINK} DESTINATION ${BIN_SUBDIR})
+endforeach()
```
dmga44: Please notice the missing endline.

src/tools/CMakeLists.txt
```diff
@@ -30,8 +30,16 @@ install(FILES sfstools.sh DESTINATION ${BIN_SUBDIR} PERMISSIONS OWNER_EXECUTE OW
...
 foreach(LINK ${SFSTOOL_LINKS})
-  add_custom_target(${LINK} ALL
-      COMMAND ln -sf sfstools.sh ${LINK}
-      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${LINK} DESTINATION ${BIN_SUBDIR})
-endforeach()
+    if(WIN32)
+        # Windows-specific command to create a symbolic link
+        add_custom_target(${LINK} ALL
+        COMMAND ${CMAKE_COMMAND} -E create_symlink sfstools.sh ${LINK}
```
dmga44: Consider not even creating these links because it is causing only Admin users could build in Windows, and we are not packing the symlinks right now.

src/tools/master_functions.cc
```diff
@@ -181,7 +249,14 @@ int open_master_conn(const char *name, uint32_t *inode, mode_t *mode, bool needr
 			printf("%s: path too long\n", name);
 			return -1;
 		}
-		strcpy(rpath + rpath_len, "/" SPECIAL_FILE_NAME_MASTERINFO);
+
+		if (rpath_len == 4 && rpath[2] == '\\' && rpath[3] == '.') {
+			strcpy(rpath + rpath_len - 1, SPECIAL_FILE_NAME_MASTERINFO);
+		} else if (rpath_len == 3 && rpath[2] == '\\') {
+			strcpy(rpath + rpath_len, SPECIAL_FILE_NAME_MASTERINFO);
```
dmga44, line 336: I guess this part of the code fixes the ```.masterinfo``` hidden file path for Windows drives paths like ```S:\```. Do you know if it works for any kind of mountpoints in Windows?, please consider we could also mount in an specific folder.

src/tools/recursive_remove.cc
```diff
@@ -58,23 +59,57 @@ static int recursive_remove(const char *file_name, int long_wait) {
 	sigaddset(&set, SIGUSR1);
 	sigprocmask(SIG_BLOCK, &set, NULL);
 
+	auto find_last_delimiter_pos = [](const std::string &parent_path) {
+		std::size_t last_pos_delimiter1 = parent_path.find_last_of("/");
+		std::size_t last_pos_delimiter2 = parent_path.find_last_of("\\");
```
dmga44: A minor suggestion, wouldn't look better ```last_pos_delimiter_unix``` and ```last_pos_delimiter_win```?

src/tools/recursive_remove.cc
```diff
@@ -58,23 +59,57 @@ static int recursive_remove(const char *file_name, int long_wait) {
...
+			last_pos_delimiter =
+			    std::max(last_pos_delimiter1, last_pos_delimiter2);
+		} else {
+			last_pos_delimiter = (last_pos_delimiter1 == std::string::npos)
+			                         ? last_pos_delimiter2
+			                         : last_pos_delimiter1;
+		}
+
+		return last_pos_delimiter;
+	};
```
dmga44, line 78: The body of the function could be the following, (only difference is that instead of returning ```std::string::npos``` when none found, it returns 0):
```
		std::size_t last_pos_delimiter1 = parent_path.find_last_of("/");
		std::size_t last_pos_delimiter2 = parent_path.find_last_of("\\");
		last_pos_delimiter1 = (last_pos_delimiter1 == std::string::npos)
		                          ? 0
		                          : last_pos_delimiter1;
		last_pos_delimiter2 = (last_pos_delimiter2 == std::string::npos)
		                          ? 0
		                          : last_pos_delimiter2;
		return std::max(last_pos_delimiter1, last_pos_delimiter2);
```

This one is a bit more compact way of doing the same, left as a minor suggestion.

src/tools/snapshot.cc
```diff
@@ -121,10 +129,25 @@ static int snapshot(const char *dstname, char *const *srcnames, uint32_t srcelem
 					uint8_t canowerwrite, int long_wait, uint8_t ignore_missing_src, int initial_batch_size) {
 	char to[PATH_MAX + 1], base[PATH_MAX + 1], dir[PATH_MAX + 1];
 	char src[PATH_MAX + 1];
+#ifdef _WIN32
+	struct stat sst = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+	            dst = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
```
dmga44: where is this constant coming from? can it be named?

Also, considering the upcoming calls of ```stat()```, are these structures not going to be filled there?

src/tools/snapshot.cc
```diff
@@ -121,10 +129,25 @@ static int snapshot(const char *dstname, char *const *srcnames, uint32_t srcelem
...
+#endif
 	int status;
 	uint32_t i, l;
 
+#ifdef _WIN32
+	for (i = 0; i < srcelements; i++) {
+		int srcname_len = strlen(srcnames[i]);
+		if (srcnames[i][srcname_len - 1] == '/' ||
+		    srcnames[i][srcname_len - 1] == '\\') {
+			srcnames[i][srcname_len - 1] = 0;
```
dmga44: Consider writing 
`srcnames[i][srcname_len - 1] = '\0';`
to make clearer you are setting the end of the string.

src/tools/snapshot.cc
```diff
@@ -204,12 +280,21 @@ static int snapshot(const char *dstname, char *const *srcnames, uint32_t srcelem
...
+							return -1;
+						}
+#else
 						if (realpath(srcnames[i], src) == NULL) {
 							printf("%s: realpath error on %s: %s\n", srcnames[i], src,
 							       strerr(errno));
 							status = -1;
 							continue;
 						}
+#endif
```
dmga44: Is it correct that the Windows code is directly returning while the Linux analogous one is setting up status and continuing?

src/tools/snapshot.cc
```diff
@@ -231,17 +323,33 @@ static int snapshot(const char *dstname, char *const *srcnames, uint32_t srcelem
...
```
dmga44: Same concern here.

src/tools/snapshot.cc
```diff
@@ -150,10 +181,17 @@ static int snapshot(const char *dstname, char *const *srcnames, uint32_t srcelem
...
+		if (GetFullPathName(dir, PATH_MAX, to, NULL) == 0) {
+			printf("%s: GetFullPathName error: %lu\n", dir, GetLastError());
+			return -1;
+		}
+#else
 		if (realpath(dir, to) == NULL) {
 			printf("%s: realpath error on %s: %s\n", dir, to, strerr(errno));
 			return -1;
 		}
+#endif
```
dmga44: Consider finding a way to try to not repeat code (and preprocessor directives) in the ```GetFullPathName``` and ```realpath``` consecutive calls.

src/tools/tools_common_functions.cc
```diff
@@ -359,20 +362,27 @@ void dirname_inplace(char *path) {
...
 	/* Find the start of the dir */
-	while (endp > path && *endp != '/') {
+	while (endp > path && (*endp != '/' && *endp != '\\')) {
 		endp--;
 	}
 
-	if (endp == path) {
+	if ((endp - path) == 2 && path[1] == ':' && path[2] == '\\') {
+		path[3] = '.';
+		path[4] = '\0';
```
dmga44, line 384: Please consider the case of mountpoints in subfolders.

dmga44 (commented):
Great work! 👍 

As a general comment, I think the changes would have looked better if they were ordered like:
- enough to be able to compile in Windows
- fixes on ```open_master_conn``` function, including the changes in the protocol.
- fixes on each of the tools, single commit or multiple commits, as you see fit.
Though it would be a chore to make it that way now.

src/common/quota_database.h
rolysr: Done

src/common/signal_handling.h
rolysr: Done

src/master/filesystem_quota.h
rolysr: In this case, I considered to use double quotes ("") for project-specific headers and angle brackets (<>) for standard library and external library headers.

src/master/filesystem_operations.cc
rolysr: Done, I reused implementation from `fs_whole_path_lookup` filesystem operation instead of reinventing a completely new one doing the same

src/tools/snapshot.cc
rolysr: Done

src/tools/snapshot.cc
rolysr: Done.

src/tools/snapshot.cc
rolysr: Done.

src/tools/snapshot.cc
rolysr: Done.

src/mount/client/saunafs_c_api.cc
rolysr: Done.

src/tools/CMakeLists.txt
rolysr: Done.

src/tools/CMakeLists.txt
rolysr: Done.

src/tools/master_functions.cc
rolysr: Yes, the tool was tested with `subfolder` and `uncpath` options during mountpoint initialization and worked as expected.

src/tools/recursive_remove.cc
rolysr: Done.

src/tools/recursive_remove.cc
rolysr: Considered.

src/tools/snapshot.cc
rolysr: Done.

src/tools/master_functions.cc
dmga44: I was referring to something like [this](https://docs.saunafs.com/windows-client#example-3).

rolysr: Done.

uristdwarf:
Here's my two cents on this:

While this is a nice bandaid to make the command-line application work on Windows, I still think we shouldn't mix code Windows and Linux code like this. We are going to have more and more of ifdefs in these programs as time goes on.

I think long-term there should be a universal library for both Windows and Linux, and then each can handle OS specifics (network, output handling etc.) as needed.

This isn't any blocker to this, just some thoughts.

src/tools/master_functions.cc
```diff
@@ -133,17 +153,126 @@ static int read_master_info(const char *name, master_info_t *info) {
...
+	size_t pos = std::string::npos;
+
+	for (int i = 0; i < n; ++i) {
+		pos = str.find(ch, pos + 1);
+		if (pos == std::string::npos) {
+			return std::string::npos;
+		}
+	}
+
+	return pos;
```
dmga44: If I understood this correctly, you're searching for the n-th occurrence of the `ch` in `str`. Would'nt be clearer to just maintain a counter while iterating through `str`?

src/tools/master_functions.cc
```diff
@@ -133,17 +153,126 @@ static int read_master_info(const char *name, master_info_t *info) {
...
+			windows_path = wsl_path.substr(pos_system_folder);
+			strncpy(rpath, windows_path.c_str(), rpath_size - 1);
+			rpath[windows_path.size()] = '\0';
+		}
+	}
+}
+#endif
+
+void get_next_path_iteration(std::string &path) {
+	size_t pos = std::min(path.find_first_of('\\'), path.find_first_of('/'));
```
dmga44, line 249: It would be nice to document a little bit what these new functions are doing.

src/tools/master_functions.cc
```diff
@@ -133,17 +153,126 @@ static int read_master_info(const char *name, master_info_t *info) {
...
 
 	rpath[0] = 0;
-	if (realpath(name, rpath) == NULL) {
-		printf("%s: realpath error on (%s): %s\n", name, rpath, strerr(errno));
+
+	std::string name_to_use = std::string(name);
+#ifdef _WIN32
+	if (name[strlen(name) - 1] == '\\' || name[strlen(name) - 1] == '/') {
+		name_to_use = std::string(name).substr(0, strlen(name) - 1);
+	}
```
dmga44: Isn't this the same as (more or less):
```suggestion

	std::string name_to_use = std::string(name);
#ifdef _WIN32
	if (name_to_use.back() == '\\' || name_to_use.back() == '/') {
		name_to_use.pop_back();
	}
```
?

src/tools/recursive_remove.cc
```diff
@@ -58,23 +59,52 @@ static int recursive_remove(const char *file_name, int long_wait) {
...
+
+	std::string name_to_use = std::string(file_name);
+#ifdef _WIN32
+	if (file_name[strlen(file_name) - 1] == '\\' || file_name[strlen(file_name) - 1] == '/') {
+		name_to_use = std::string(file_name).substr(0, strlen(file_name) - 1);
+	}
```
dmga44: Please check if the previous code suggestion about `name_to_use` works here as well.

src/tools/snapshot.cc
```diff
@@ -121,11 +131,32 @@ static int snapshot(const char *dstname, char *const *srcnames, uint32_t srcelem
...
 
-	if (stat(dstname, &dst) < 0) {  // dst does not exist
+#ifdef _WIN32
+	for (size_t i = 0; i < lookup_srcnames.size(); i++) {
+		if (lookup_srcnames[i][lookup_srcnames[i].size() - 1] == '\\' ||
+		    lookup_srcnames[i][lookup_srcnames[i].size() - 1] == '/') {
+			lookup_srcnames[i] =
+			    lookup_srcnames[i].substr(0, lookup_srcnames[i].size() - 1);
+		}
+	}
```
dmga44: Same suggestion of the `name_to_use`.

dmga44 (approved):
Great job, @rolysr , once again!

Beside my minor comments, please check if the test `test_sfstools` from the SanityChecks suite is subject to your change of the saunafs command tool in the tests.

src/tools/snapshot.cc
rolysr: Done

src/tools/recursive_remove.cc
rolysr: Done

src/tools/master_functions.cc
rolysr: Done.

src/tools/master_functions.cc
rolysr: Done.

src/tools/master_functions.cc
rolysr: Done.

src/common/quota_database.cc
```diff
@@ -19,8 +19,7 @@
  */
 
 #include "common/platform.h"
-
-#include "master/quota_database.h"
+#include "common/quota_database.h"
```
lgsilva3087: The lines were separated to prevent clang-format from sorting the "common/platform.h", which is included almost on every file.

src/master/filesystem_quota.cc
```diff
@@ -24,10 +24,11 @@
 #include <cassert>
 
 #include "common/event_loop.h"
+#include "common/small_vector.h"
+#include "common/quota_database.h"
```
lgsilva3087: Sort this please.

src/master/filesystem_quota.h
lgsilva3087: We could consider to migrate to <> always, to enforce correct relative paths (added on purpose with cmake's include_directories). This way will also help to always use paths relative to src (as intended in the main CMakeLists.txt).

src/master/filesystem_quota.h
```diff
@@ -25,6 +25,8 @@
 #include <utility>
```
lgsilva3087, line 25: We are usualy separating in a different block the system headers from our headers, even if clang-format sorts them correctly. Please consider it.

src/master/quota_database_unittest.cc
```diff
@@ -19,8 +19,7 @@
  */
 
 #include "common/platform.h"
-
-#include "master/quota_database.h"
+#include "common/quota_database.h"
```
lgsilva3087: Same here, please use a separate group for "common/platform.h" to avoid group sorting issues, until we get rid of the file.

src/tools/append_file.cc
```diff
@@ -61,8 +61,13 @@ static int append_file(const char *fname, const char *afname) {
 		printf("%s: not a file\n", fname);
 		return -1;
 	}
+#ifdef _WIN32
+	uid = 0;
+	gid = 0;
+#else
 	uid = getuid();
 	gid = getgid();
+#endif
```
lgsilva3087: This is repeated in some places, consider functions like:
```cpp
inline uint32_t getUId() {
#ifdef _WIN32
	return 0;
#else
	return getuid();
#endif
}
```

src/tools/file_repair.cc
```diff
@@ -49,13 +49,20 @@ static int file_repair(const char *fname, uint8_t correct_only_flag) {
 	if (fd < 0) {
 		return -1;
 	}
+#ifdef _WIN32
+	uint32_t uid = 0;
+	uint32_t gid = 0;
+#else
+	uint32_t uid = getuid();
+	uint32_t gid = getgid();
+#endif
```
lgsilva3087: Same here.

src/tools/quota_rep.cc
```diff
@@ -163,8 +167,13 @@ static void quota_print_rep(const std::string &path, uint32_t path_inode,
 static int quota_rep(const std::string &path, std::vector<int> requested_uids,
 					 std::vector<int> requested_gid, bool report_all, bool per_directory_quota) {
 	std::vector<uint8_t> request;
+#ifdef _WIN32
+	uint32_t uid = 0;
+	uint32_t gid = 0;
+#else
 	uint32_t uid = getuid();
 	uint32_t gid = getgid();
+#endif
```
lgsilva3087: Here also.

src/tools/quota_set.cc
```diff
@@ -47,8 +47,13 @@ static int quota_set(const std::string &path, QuotaOwner owner, uint64_t soft_in
 		return -1;
 	}
 
+#ifdef _WIN32
+	uint32_t uid = 0;
+	uint32_t gid = 0;
+#else
 	uint32_t uid = getuid();
 	uint32_t gid = getgid();
+#endif
```
lgsilva3087: Also here. If the suggestion is accepted, please search for other places.

src/tools/snapshot.cc
```diff
@@ -25,18 +25,23 @@
 #include <signal.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/types.h>
 #include <sys/stat.h>
```
lgsilva3087: sort

lgsilva3087 (approved):

src/common/quota_database.cc
rolysr: Done.

src/master/filesystem_quota.cc
rolysr: Done.

src/master/filesystem_quota.h
rolysr: Done.

src/master/quota_database_unittest.cc
rolysr: Done.

src/tools/append_file.cc
rolysr: Done.

src/tools/file_repair.cc
rolysr: Done.

src/tools/quota_rep.cc
rolysr: Done.

src/tools/quota_set.cc
rolysr: Done.

src/tools/snapshot.cc
rolysr: Done.

PR #261, opened 2024-12-19
build: Remove debian packaging related scripts and files

uristdwarf:
These files are now located at https://github.com/leil-io/debian-package

PR #300, opened 2025-02-13
fix: Add permission to DATA_PATH set by package

antuan96314:
The metalogger package was not correctly setting user and group to default DATA_PATH. This was unnoticed because the path was shared by master package and only if metalogger is install stand-alone then the error is raised.

### DIFF
--- a/debian/saunafs-metalogger.postinst
+++ b/debian/saunafs-metalogger.postinst
@@ -1,0 +1,21 @@
+#! /bin/sh
+
+# Ensure proper permissions on data dir and run dir.
+
+SAUNAFS_USER="saunafs"
+SAUNAFS_GROUP="saunafs"
+SAUNAFS_DATA_PATH="/var/lib/saunafs"
+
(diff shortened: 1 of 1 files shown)


uristdwarf:
@antuan96314 You should do this change in https://github.com/leil-io/debian-package, the deb packaging is now located and maintained there. Let me know if you need any help

antuan96314:
> @antuan96314 You should do this change in https://github.com/leil-io/debian-package, the deb packaging is now located and maintained there. Let me know if you need any help

I forgot about that repo. I just migrated the commit. Are you keeping the debian directory in both repos sync in any way? Should i delete this commit then?

uristdwarf:
> I forgot about that repo. I just migrated the commit. Are you keeping the debian directory in both repos sync in any way? 

No, there's a PR to remove the debian #261, but it's still waiting

> Should i delete this commit then?

No worries, I will remove the branch

PR #301, opened 2025-02-14
feat(uraft): Introduce floating IP manager for HA

ralcolea:
Previously, a floating IP drop triggered a new election in uRaft, sometimes leading to leader flip-flopping.

This PR introduces a FloatingIP Manager to efficiently handle floating IPs using an AF_NETLINK-based event listener. This ensures low CPU usage, automatic IP failover, and continuous monitoring.

**Key Features & Changes**
- **FloatingIP Manager Implementation**
   - Manages the floating IP lifecycle.
   - Provides automatic IP restoration on failure.
   - Supports clean start/stop functionality.

- **Netlink Event Listener**
  - AF_NETLINK + select() approach for real-time event detection.
  - Monitors RTM_DELADDR events to detect IP removal.
  - Captures network interface & IP address for accurate recovery.
  - Efficient, non-blocking design with adjustable timeout (default value 500 ms).

- **High Availability Enhancements**
  - Auto-recovery mechanism: If the socket crashes, it is restarted automatically.
  - Stop flag mechanism to gracefully exit the listener thread.
  - Failover handling: If the floating IP is lost, saunafs-uraft-helper script is used to restore the floating IP.

- **Code Quality & Performance**
  - Optimized low CPU consumption compared to polling-based methods.
  - Improved error handling and logging.
  - Added unit tests for uRaft (floating_ip_manager_unittest.cc) and integration tests in Ansible repository (test-uraft-floating-ip-manager branch).

**Advantages of this change**
- Replaces inefficient polling (arping/ip addr show) with a real-time, event-driven approach.
- Ensures high availability by detecting IP loss instantly.
 
🚀 **Performance Improvements**
✅ Netlink approach: Low CPU overhead, instant wakes up during IP drop.
❌ Polling (arping or ip addr show): High CPU usage, unnecessary checks.

🛠️ Testing Done
✅ Unit-tested IP detection & restoration.
✅ Simulated IP removal events.

### DIFF
--- a/doc/saunafs-uraft.cfg.5.adoc
+++ b/doc/saunafs-uraft.cfg.5.adoc
@@ -60,7 +60,9 @@ heartbeat messages.
 checking whether local master is alive.
 
 *URAFT_FLOATING_IP_CHECK_PERIOD*:: *[advanced]* period in milliseconds between
-checking whether the floating IP is alive.
+checking whether the floating IP is alive. A value of 0 disables floating IP
+monitoring and its automatic recovery mechanism. Adjust this setting based on
+network stability and failover requirements.
 
 == REPORTING BUGS
 
--- a/doc/saunafs-uraft.cfg.5.adoc
+++ b/doc/saunafs-uraft.cfg.5.adoc
@@ -59,6 +59,9 @@ heartbeat messages.
 *LOCAL_MASTER_CHECK_PERIOD*:: *[advanced]* period in milliseconds between
 checking whether local master is alive.
 
+*URAFT_FLOATING_IP_CHECK_PERIOD*:: *[advanced]* period in milliseconds between
+checking whether the floating IP is alive.
+
 == REPORTING BUGS
 
 Report bugs to the Github repository <https://github.com/leil/saunafs> as an
--- a/doc/saunafs-uraft.cfg.5.adoc
+++ b/doc/saunafs-uraft.cfg.5.adoc
@@ -46,6 +46,11 @@ server (default: 9421).
 
 *URAFT_FLOATING_IFACE*:: Network interface where floating IP should be managed
 (example: eth0).
+Important: If the network interface used for the floating IP becomes unavailable,
+the failover mechanism will not restore it. As a result, the floating IP cannot
+be recovered, causing SaunaFS services that depend on it to become unavailable.
+The failover mechanism is based on the URAFT_FLOATING_IP_CHECK_PERIOD option.
+See its documentation for more details.
 
 *ELECTION_TIMEOUT_MIN*:: *[advanced]* minimal timeout in milliseconds for
 election algorithm.
--- a/src/uraft/floating-ip-manager.cc
+++ b/src/uraft/floating-ip-manager.cc
@@ -24,14 +24,15 @@
 #include <sys/select.h>
 #include <sys/socket.h>
 #include <sys/syslog.h>
+#include <stop_token>
 
 
 HAFloatingIPManager::HAFloatingIPManager(const std::string &iface,
                                          const std::string &ipAddress,
                                          const int &checkPeriod)
-    : floatingIPInterface(iface),
-      floatingIPAddress(ipAddress),
-      checkFloatingIPPeriodMS(checkPeriod) { }
+    : floatingIpInterface(iface),
+      floatingIpAddress(ipAddress),
+      checkFloatingIpPeriodMS(checkPeriod) { }
 
 HAFloatingIPManager::~HAFloatingIPManager() {
 	stopEventListener();
@@ -49,11 +50,11 @@ void HAFloatingIPManager::stop() {
 	stopEventListener();
 }
 
-bool HAFloatingIPManager::isFloatingIPAlive() const {
-	return _isFloatingIPAlive;
+bool HAFloatingIPManager::isFloatingIpAlive() const {
+	return _isFloatingIpAlive;
 }
 
-bool HAFloatingIPManager::restoreFloatingIP() {
+bool HAFloatingIPManager::restoreFloatingIp() {
 	std::string command = "saunafs-uraft-helper assign-ip";
 	int result = system(command.c_str());
 
@@ -65,37 +66,36 @@ bool HAFloatingIPManager::restoreFloatingIP() {
 }
 
 void HAFloatingIPManager::startEventListener() {
-	stopListenerFlag = false;
 	if (!listenerThread.joinable()) {
 		listenerThread =
-		    std::thread(&HAFloatingIPManager::eventListenerThread, this);
-		_isFloatingIPAlive = true;
+		    std::jthread(&HAFloatingIPManager::eventListenerThread, this);
+		_isFloatingIpAlive = true;
 	}
 }
 
 void HAFloatingIPManager::stopEventListener() {
-	stopListenerFlag = true;
 	if (listenerThread.joinable()) {
-		listenerThread.join();
-		_isFloatingIPAlive = false;
+		listenerThread.request_stop();
+		_isFloatingIpAlive = false;
 	}
 }
 
-void HAFloatingIPManager::handleIPLoss(const std::string &ipAddress) {
+void HAFloatingIPManager::handleIpLoss(const std::string &ipAddress) {
 	syslog(LOG_WARNING, "[FloatingIPManager] Handling lost IP: %s",
 	       ipAddress.c_str());
-	_isFloatingIPAlive = false;
+	_isFloatingIpAlive = false;
+
 	// Trigger failover
-	if (!stopListenerFlag) { _isFloatingIPAlive = restoreFloatingIP(); }
+	const std::stop_token stopToken = listenerThread.get_stop_token();
+	if (!stopToken.stop_requested()) {
+		_isFloatingIpAlive = restoreFloatingIp();
+	}
 }
 
-void HAFloatingIPManager::eventListenerThread() {
-	constexpr int kBufferSize = 8192; // 8KB
-	char buffer[kBufferSize];
+void HAFloatingIPManager::eventListenerThread(const std::stop_token &stopToken) {
 	constexpr int kRetryTimeout = 2; // timeout before retrying
-	constexpr long kMillisecondsInOneMicrosecond = 1000;
 
-	while (!stopListenerFlag) {
+	while (!stopToken.stop_requested()) {
 		int sock_fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
 		if (sock_fd < 0) {
 			syslog(LOG_WARNING,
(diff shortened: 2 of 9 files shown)


src/uraft/floating-ip-manager.h
```diff
@@ -0,0 +1,134 @@
...
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#ifndef FLOATING_IP_MANAGER_H
+#define FLOATING_IP_MANAGER_H
```
lgsilva3087: Please use only one option to ensure the header is included once.
The project usually uses:
`#pragma once`

src/uraft/floating-ip-manager.h
```diff
@@ -0,0 +1,134 @@
...
+	std::thread listenerThread; ///< Background thread for event listening.
+	std::atomic<bool> stopListenerFlag{false}; ///< Flag to stop the event listener thread.
+	bool _isFloatingIPAlive{false}; ///< Flag indicating if the floating IP is alive.
+
+	std::string floatingIPInterface; ///< Network interface for the floating IP.
+	std::string floatingIPAddress;   ///< Floating IP address.
+	int checkFloatingIPPeriodMS = 500; ///< Period (in milliseconds) to check the floating IP status.
+};
+
+#endif  // FLOATING_IP_MANAGER_H
```
lgsilva3087: Maybe not needed according to my previous comment.

src/uraft/floating-ip-manager.h
```diff
@@ -0,0 +1,134 @@
...
+	/// network interface addresses and link status using the Netlink protocol.
+	/// It uses a Netlink socket to listen for RTM_DELADDR messages, which
+	/// indicate the removal of an IP address from an interface. If the removed
+	/// IP matches the configured floating IP, it triggers the handleIPLoss()
+	/// function to take appropriate recovery actions.
+	void eventListenerThread();
+
+protected:
+	std::thread listenerThread; ///< Background thread for event listening.
+	std::atomic<bool> stopListenerFlag{false}; ///< Flag to stop the event listener thread.
```
lgsilva3087: Consider std::jthread that is designed for this pattern.

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+				       "Restarting socket...");
+				break;  // Restart socket
+			}
+
+			for (nlh = (struct nlmsghdr *)buffer; NLMSG_OK(nlh, recv_len);
+			     nlh = NLMSG_NEXT(nlh, recv_len)) {
+				if (nlh->nlmsg_type == NLMSG_DONE) { break; }
+				if (nlh->nlmsg_type == NLMSG_ERROR) {
+					syslog(LOG_WARNING,
+					       "[FloatingIPManager] Received an error message.");
```
lgsilva3087: Does the message contains printable text?

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+			syslog(
+			    LOG_WARNING,
+			    "[FloatingIPManager] Failed to bind socket. Retrying in %ds...",
+			    kRetryTimeout);
+			close(sock_fd);
+			std::this_thread::sleep_for(std::chrono::seconds(kRetryTimeout));
+			continue;  // Retry loop
+		}
+
+		while (!stopListenerFlag) {
```
lgsilva3087: Consider to move this loop to a new function, to separate it from the retry logic.
The new function name could be pollForIpRemovalEvents.

lgsilva3087 (approved):
Great work. Please see my suggestions.

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+bool HAFloatingIPManager::restoreFloatingIP() {
+	std::string command = "saunafs-uraft-helper assign-ip";
+	int result = system(command.c_str());
+
+	if (result != 0) {
+		syslog(LOG_WARNING, "Command %s was not successful", command.c_str());
+	}
+
+	return (result == 0);
+}
```
uristdwarf: In the future, I think we should move more of the bash script stuff into the C++ code. The bash script can be notoriously slow (speed is a requirement for a program like uraft) and we should generally not rely on an outside script (however tempting it may be). Anything we can do in the bash script, we can do in C++.

However, I like that you are keeping things simple right now, and using the script. Gradually we could phase-out the script, but this is perfect start.

However, please be careful when using standard C-library functions, if these functions fail, they will often set errno which will contain actual information on the error. Worse, if you call any other function between this time and, it might change errno. See `man 3 errno` and `man system`

Please change this to actually print what error occurred using errno and strerror. read `man 3 strerror` for a thread safe version.

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+	while (!stopListenerFlag) {
+		int sock_fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+		if (sock_fd < 0) {
+			syslog(LOG_WARNING,
+			       "[FloatingIPManager] Failed to create socket. "
+			       "Retrying in %ds...",
+			       kRetryTimeout);
+			std::this_thread::sleep_for(std::chrono::seconds(kRetryTimeout));
+			continue;  // Retry loop
+		}
```
uristdwarf: As above, read `man 3 socket`. Usually standard C library functions set errno if they return less than 0, however you need to check and confirm it using `man 3 <FUNCTION_NAME>`

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+		if (bind(sock_fd, reinterpret_cast<struct sockaddr *>(&local_addr),
+		         sizeof(local_addr)) < 0) {
+			syslog(
+			    LOG_WARNING,
+			    "[FloatingIPManager] Failed to bind socket. Retrying in %ds...",
+			    kRetryTimeout);
+			close(sock_fd);
+			std::this_thread::sleep_for(std::chrono::seconds(kRetryTimeout));
+			continue;  // Retry loop
+		}
```
uristdwarf: Check if errno is set and print out actual error (see above)

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+			    0, checkFloatingIPPeriodMS * kMillisecondsInOneMicrosecond};
+
+			int result =
+			    select(sock_fd + 1, &read_fds, nullptr, nullptr, &timeout);
+
+			if (result < 0) {
+				syslog(LOG_WARNING,
+				       "[FloatingIPManager] select() function failed.");
+				break;
+			}
```
uristdwarf: Check if errno is set and print out actual error (see above)

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+			msg.msg_iov = &iov;
+			msg.msg_iovlen = 1;
+
+			ssize_t recv_len = recvmsg(sock_fd, &msg, 0);
+			if (recv_len < 0) {
+				syslog(LOG_WARNING,
+				       "[FloatingIPManager] recvmsg() failed. "
+				       "Restarting socket...");
+				break;  // Restart socket
+			}
```
uristdwarf: Check if errno is set and print out actual error (see above)

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+								handleIPLoss(ipAddress);
+							}
+						}
+						rta = RTA_NEXT(rta, attr_len);
+					}
+				}
+			}
+		}
+
+		close(sock_fd);
```
uristdwarf: Check for error

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+						    LOG_WARNING,
+						    "[FloatingIPManager] IP removed from interface: %s",
+						    networkInterface);
+					} else {
+						syslog(LOG_WARNING,
+						       "[FloatingIPManager] Failed to get "
+						       "interface name for index: %d",
+						       ifa->ifa_index);
+						continue;  // It happens when the IP is already removed
+					}
```
uristdwarf: errno is set on error, print out the actual error (see above)

> // It happens when the IP is already removed

Are you sure about this? This doesn't seem to have anything to do with IP addresses, only interfaces.
My concern here is that interfaces could be removed. This might be the only case were a leader demote may be warranted (as there is no interface, no way to actually bind to anything).

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+					// Parse attributes to extract the removed IP address
+					auto *rta = (struct rtattr *)((char *)ifa +
+					                              NLMSG_ALIGN(sizeof(*ifa)));
+					int attr_len = nlh->nlmsg_len - NLMSG_LENGTH(sizeof(*ifa));
+
+					while (RTA_OK(rta, attr_len)) {
+						if (rta->rta_type == IFA_LOCAL) { // Local IP address
+							char ipAddress[INET_ADDRSTRLEN] = {0};
+							inet_ntop(AF_INET, RTA_DATA(rta), ipAddress,
+							          sizeof(ipAddress));
```
uristdwarf: According to `man 3 inet_ntop`, this function will return NULL and set errno in case of error.

src/uraft/floating_ip_manager_unittest.cc
```diff
@@ -0,0 +1,165 @@
...
+			std::exit(EXIT_FAILURE);
+		}
+		if (system("sudo ip addr add 192.168.1.105/24 dev test0") != 0) {
+			std::cerr << "Failed to add 192.168.1.105 to test0\n";
+			std::exit(EXIT_FAILURE);
+		}
+		if (system("sudo ip link set test0 up") != 0) {
+			std::cerr << "Failed to set test0 interface up\n";
+			std::exit(EXIT_FAILURE);
+		}
```
uristdwarf, line 71: Is sudo needed here? I don't think it will work in the Google Test Framework. Regardless, we are running the tests as root, so you don't need it.

src/uraft/floating_ip_manager_unittest.cc
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	bool wasFloatingIPRestored() const { return restored; }
+
+	std::string floatingIP() const { return floatingIPAddress; }
+	std::string floatingIPIface() const { return floatingIPInterface; }
+
+private:
+	bool restored = false;
+};
```
uristdwarf: Why mocking? Why not use HAFloatingIpManager directly?

src/uraft/main.cc
```diff
@@ -48,7 +48,10 @@ void parseOptions(int argc, char **argv, uRaftController::Options &opt, bool &ma
 	("URAFT_DEMOTE_TIMEOUT", po::value<int>()->default_value(1000000000), "demote timeout (ms)")
 	("URAFT_DEAD_HANDLER_TIMEOUT", po::value<int>()->default_value(1000000000), "metadata server dead handler timeout (ms)")
 	("URAFT_CHECK_CMD_PERIOD", po::value<int>()->default_value(100), "check command status period(ms)")
-	("URAFT_STATUS_PORT", po::value<int>()->default_value(9428), "node status port");
+	("URAFT_STATUS_PORT", po::value<int>()->default_value(9428), "node status port")
+	("URAFT_FLOATING_IP", po::value<std::string>(), "floating IP address")
+	("URAFT_FLOATING_IFACE", po::value<std::string>(), "floating IP interface")
+	("URAFT_FLOATING_IP_CHECK_PERIOD", po::value<int>()->default_value(500), "floating IP check status period");
```
uristdwarf: One way of allowing to disable this floating IP check is by setting this to 0 and not using the floating IP manager.

uristdwarf: Regardless, the types should be uint, not int

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+
+bool HAFloatingIPManager::restoreFloatingIP() {
+	std::string command = "saunafs-uraft-helper assign-ip";
+	int result = system(command.c_str());
+
+	if (result != 0) {
+		syslog(LOG_WARNING, "Command %s was not successful", command.c_str());
+	}
+
+	return (result == 0);
```
uristdwarf: According to `man 3 system`:

```
       •  If  command is NULL, then a nonzero value if a shell is available, or
          0 if no shell is available.

       •  If a child process could not be created, or its status could  not  be
          retrieved,  the  return  value is -1 and errno is set to indicate the
          error.

       •  If a shell could not be executed in the child process, then  the  re‐
          turn  value  is  as  though  the  child  shell  terminated by calling
          _exit(2) with the status 127.

       •  If all system calls succeed, then the return value is the termination
          status of the child shell used to execute command.  (The  termination
          status  of  a  shell is the termination status of the last command it
          executes.)
```

You should check for these cases

src/uraft/floating-ip-manager.cc
```diff
@@ -0,0 +1,228 @@
...
+	stopEventListener();
+}
+
+bool HAFloatingIPManager::isFloatingIPAlive() const {
+	return _isFloatingIPAlive;
+}
+
+bool HAFloatingIPManager::restoreFloatingIP() {
+	std::string command = "saunafs-uraft-helper assign-ip";
+	int result = system(command.c_str());
```
uristdwarf: This system could get stuck, since there's no timeout

uristdwarf (changes requested):
First off, amazing job you have done with this. This is really great and it's nice to see some tests being added.

However, I have a few requests:

1. In all previous iterations of this feature, we've been forcing users to use this. However in case anything goes wrong, we should allow them to disable this. That means we don't have to critically patch this feature in case anything goes wrong. See one of my comments regarding an idea on how to do this.
2. You are using a lot of standard C library functions. Please double check the manuals for these (with `man 3 <FUNCTION_NAME>`) and make sure you are checking and printing the actual error messages. Make sure you are also freeing any pointers they return if they require it.

src/uraft/floating_ip_manager_unittest.cc
```diff
@@ -0,0 +1,165 @@
...
+	if (system(command.c_str()) != 0) {
+		std::cerr << "Failed to remove " + ipAddress + " from " + ipIface + "\n";
+		std::exit(EXIT_FAILURE);
+	}
+
+	// Wait for a short period to allow the listener to detect the IP loss
+	std::this_thread::sleep_for(std::chrono::seconds(1));
+
+	// Verify that handleIPLoss was called
+	EXPECT_TRUE(mockManager->isFloatingIPAlive());
```
rolysr: Consider checking `EXPECT_FALSE(mockManager->isFloatingIPAlive());` before the short period wait to have covered the case when `isFloatingIPAlive` returns false as expected.

src/uraft/floating_ip_manager_unittest.cc
```diff
@@ -0,0 +1,165 @@
...
+		std::cerr << "Failed to remove " + ipAddress + " from " + ipIface + "\n";
+		std::exit(EXIT_FAILURE);
+	}
+
+	// Wait for a short period to allow the listener to detect the IP loss
+	std::this_thread::sleep_for(std::chrono::seconds(1));
+
+	// Verify that handleIPLoss was only called for mockManager105
+	EXPECT_TRUE(mockManager105->wasFloatingIPRestored());
+	EXPECT_FALSE(mockManager->wasFloatingIPRestored());
```
rolysr: Similar consideration from my previous comment, but this time with the  `wasFloatingIPRestored()` returned value

rolysr (approved):
Well done @ralcolea. LGTM. Just a few comments added.

src/uraft/floating-ip-manager.h
ralcolea: I will use `#pragma once`.

src/uraft/floating-ip-manager.h
ralcolea: Done.

src/uraft/floating-ip-manager.h
ralcolea: Very good suggestion. Done.

src/uraft/floating-ip-manager.cc
ralcolea: Yes, it does. I added the error information for this scenario.

src/uraft/floating-ip-manager.cc
ralcolea: Very nice suggestion. After applying this suggestion, function reponsibilities are better distributed and the code is more readable and modular.

src/uraft/floating_ip_manager_unittest.cc
ralcolea: Nice catch!!! Done.

src/uraft/floating_ip_manager_unittest.cc
ralcolea: Nice catch!!! Done.

src/uraft/floating-ip-manager.cc
ralcolea: I agree with you and I really like the idea of moving most of the bash script functions into the C++ code. However, as it will take more time and the priority is to solve the leader flip-flopping issue, I decided to keep the bash script in its current state and then start moving progressively its functions to the C++ code.

I have been addressing your suggestions related with the use of C-library functions and I improved the errors logging based on `errno` as you suggested. Please, take a look again and let me know if the current version is more aligned to your expectations.

src/uraft/floating-ip-manager.cc
ralcolea: Done.

src/uraft/floating-ip-manager.cc
ralcolea: Done.

src/uraft/floating-ip-manager.cc
ralcolea: Done.

src/uraft/floating-ip-manager.cc
ralcolea: Done.

src/uraft/floating_ip_manager_unittest.cc
ralcolea: sudo is needed for these tests because, until now, unit tests were not being run as root or with sudo permissions.

From my understanding, using sudo in the Google Test framework works as long as the user running the tests has sudo privileges. IMHO, it's preferable to run only the necessary tests with sudo rather than executing all tests with elevated privileges just because a few require it.

src/uraft/floating_ip_manager_unittest.cc
ralcolea: Mocking is more appropriate here because `HAFloatingIpManager::restoreFloatingIP()` relies on the `saunafs-uraft-helper` bash script. Therefore, mocking allows us to focus on testing the intended functionality without external dependencies.

src/uraft/main.cc
ralcolea: I added the ability of disabling the floating IP manager by setting `URAFT_FLOATING_IP_CHECK_PERIOD=0` in the `saunafs-uraft.cfg`, as you suggested. I also updated the types to use uint.

src/uraft/floating-ip-manager.cc
ralcolea: I have updated the function to execute the command. Please review the changes and let me know your thoughts.

src/uraft/floating-ip-manager.cc
ralcolea: I also considered this requirement in the new version.

src/uraft/floating-ip-manager.cc
ralcolea: Done.

src/uraft/floating-ip-manager.cc
ralcolea: The documentation mentions as possible errors `ENODEV: No interface found with give name` and `ENXIO: No interface found for the index`. But, it also says that the function can fail for any of the errors specified by `socket()` or `ioctl()`. I agree with you. This error does not seem related with IP addresses, but with interfaces.

src/uraft/floating-ip-manager.cc
ralcolea: Done.

doc/saunafs-uraft.cfg.5.adoc
```diff
@@ -46,6 +46,9 @@ server (default: 9421).
 
 *URAFT_FLOATING_IFACE*:: Network interface where floating IP should be managed
 (example: eth0).
+Important: If the network interface used for the floating IP becomes unavailable,
+the failover mechanism will not restore it. As a result, the floating IP cannot
+be recovered, causing SaunaFS services that depend on it to become unavailable.
```
uristdwarf, line 51: Should mention the failover mechanism is dependent on this option: URAFT_FLOATING_IP_CHECK_PERIOD

uristdwarf (approved):

doc/saunafs-uraft.cfg.5.adoc
ralcolea: Done.

PR #303, opened 2025-02-16
fix(mount): Add mountpoint id to notifications

rolysr:
Previous version of notification area logging on the client was not able to distinguish between different mountpoints when showing notifications. This commit fixes this by adding the mountpoint id to the notification title.


antuan96314 (approved):
LGTM

dmga44 (approved):
LGTM.

PR #306, opened 2025-02-20
chore(git,cmake): Ignore CMakeUserPresets.json

lgsilva3087:
The user presets are meant to be used locally, no need to version it.


uristdwarf (approved):
Though less of an issue, if you can add it to .dockerignore as well.

ralcolea (approved):
LGTM :+1:

PR #307, opened 2025-02-21
Improve chunkserver logging

lgsilva3087:
This PR contain changes to facilitate the debugging, mainly but not limited to the Chunkserver.
The main changes are:
- The slogger is now able to add the log level and the thread id (TID) to the syslog.
- The hddRead function now prints more useful information in debug mode.
- Misc: the tests relying in the log message were updated.

### DIFF
--- a/src/slogger/slogger.cc
+++ b/src/slogger/slogger.cc
@@ -187,7 +187,10 @@ void safs::drop_all_logs() {
 bool safs::add_log_syslog([[maybe_unused]] log_level::LogLevel level) {
 #ifndef _WIN32
 	try {
-		LoggerPtr logger = spdlog::syslog_logger_mt("syslog");
+		static constexpr bool kEnableFormatting = true;
+		// Except for first and last arguments, the rest are defaults
+		LoggerPtr logger = spdlog::syslog_logger_mt("syslog", "", 0, LOG_USER, kEnableFormatting);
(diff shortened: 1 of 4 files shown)


src/slogger/slogger.cc
```diff
@@ -187,7 +187,10 @@ void safs::drop_all_logs() {
...
 #ifndef _WIN32
 	try {
-		LoggerPtr logger = spdlog::syslog_logger_mt("syslog");
+		static constexpr bool kEnableFormatting = true;
+		// Except for first and last arguments, the rest are defaults
+		LoggerPtr logger = spdlog::syslog_logger_mt("syslog", "", 0, LOG_USER, kEnableFormatting);
+		logger->set_pattern("[%t] %l: %v");
 		logger->set_level((spdlog::level::level_enum)level);
 		return true;
 	} catch (const spdlog::spdlog_ex &e) {
```
uristdwarf, line 196: This seems to only add it for syslog, why not add it for stderr as well?

uristdwarf (approved):

src/slogger/slogger.cc
lgsilva3087: For stderr was already there.

uristdwarf: OK, that makes sense.

ralcolea (approved):
LGTM :+1:

PR #309, opened 2025-02-26
fix: Inconsistent total capacity in CGI and statfs

uristdwarf:
The CGI and statfs system call where reporting different total capacities of the filesystem.

The CGI monitoring was showing the total capacity of all hard drives, regardless of limits in HDD_LEAVE_SPACE_DEFAULT option. Meanwhile, statfs was showing total capacity as space available on all hard drives + actual used bytes for chunks (or the reported total capacity on all hard drives, whichever was lower).

This commit tries to get rid of these discrepancies by discounting the HDD_LEAVE_SPACE_DEFAULT value towards the final capacity of a drive in the chunkserver. Both statfs and CGI will now show this capacity in the same way.

### DIFF
--- a/src/chunkserver/chunkserver-common/cmr_disk.cc
+++ b/src/chunkserver/chunkserver-common/cmr_disk.cc
@@ -84,6 +84,11 @@ void CmrDisk::refreshDataDiskUsage() {
 	} else {
 		setAvailableSpace(availableSpace() - leaveFreeSpace());
 	}
+	if (totalSpace() < leaveFreeSpace()) {
+		setTotalSpace(0ULL);
+	} else {
+		setTotalSpace(totalSpace() - leaveFreeSpace());
+	}
 }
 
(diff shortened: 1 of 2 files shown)


src/chunkserver/chunkserver-common/cmr_disk.cc
```diff
@@ -84,6 +84,11 @@ void CmrDisk::refreshDataDiskUsage() {
 	} else {
 		setAvailableSpace(availableSpace() - leaveFreeSpace());
 	}
+	if (totalSpace() < leaveFreeSpace()) {
+		setTotalSpace(0ULL);
```
rolysr, line 88: Consider setting the value passed here as a constant.

rolysr (approved):
LGTM.

uristdwarf:
Closing this PR, this approach is messing with balancing and I deem it too risky to fix whatever is going wrong with the test

PR #310, opened 2025-02-26
build(uraft): Update boost functions to work with boost 1.87

uristdwarf:
Uraft was causing compilation errors with boost 1.87, as some backwards compatible stuff was removed in later versions. This fixes those issues by using library calls available in Boost 1.87 and earlier versions.

### DIFF
--- a/src/uraft/main.cc
+++ b/src/uraft/main.cc
@@ -226,7 +226,7 @@ int main(int argc, char **argv) {
 		return EXIT_FAILURE;
 	}
 
-	boost::asio::io_service  io_service;
+	boost::asio::io_context  io_service;
 	uRaftController          server(io_service);
 #if (BOOST_VERSION >= 104700)
 	boost::asio::signal_set  signals(io_service, SIGINT, SIGTERM);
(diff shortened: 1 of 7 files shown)


aNeutrino:
@ralcolea we may need your help to run it against URAFT ansible e2e tests. Before we merge it.

ralcolea:
> @ralcolea we may need your help to run it against URAFT ansible e2e tests. Before we merge it.

Sure, just let me know and I will gladly join you for testing :)

uristdwarf:
Tests seem to work fine, so I think this is ready for review.

ralcolea (approved):
Great job, @uristdwarf! 👍
Nice **Boost** to uRaft 😄

src/uraft/main.cc
```diff
@@ -236,7 +236,7 @@ int main(int argc, char **argv) {
 		server.set_options(opt);
 		makePidFile(pidfile);
 #if (BOOST_VERSION >= 104700)
```
rolysr, line 238: Maybe this is a good opportunity to substitute `104700` by something describing what is available from that version that matters to our code.

src/uraft/uraft.cc
```diff
@@ -1,5 +1,6 @@
 #include "common/platform.h"
 #include "uraft.h"
+#include <boost/lexical_cast.hpp>
```
rolysr: Consider to separate these imports so that it could be separated by type as we do on there files, for instance:
```
#include "common/platform.h"

#include <boost/lexical_cast.hpp>

#include "uraft.h"
```

rolysr (approved):
LGTM. Consider my minor suggestions if applicable.

src/uraft/main.cc
uristdwarf: I think this is out of scope for now, but we need to review these ifdefs anyway in the future.

PR #312, opened 2025-02-27
fix: Show same stats as CGI in statfs call

uristdwarf:
Statfs was showing different values for available, total and used space than CGI was showing. This changes that to show the same values for statfs system call as fs_info call.


ralcolea (commented):
I like your change. However, I'm not sure if this solution works correctly for all the scenarios. How do you plan to validate it?

ralcolea (approved):
Good job @uristdwarf :+1:

dmga44 (approved):

wg1k (approved):

PR #313, opened 2025-02-28
fix(mount): Fix `htonl` dependency on Windows

rolysr:
Previous version of the code was using `htonl` function on client side from Linux specific header. This commit enable the usage of htonl function to also work for the Windows client.


ralcolea (approved):
LGTM

dmga44 (approved):

PR #318, opened 2025-03-05
feat(mount): Improve mount read algorithm

dmga44:
List of changes:
- reduce the kConservativeMultiplier constant from 2 to 1. Should
reduce the number of possibly wasted read requests.
- add priority to the ReadaheadRequestContainer. This way the most
needed requests should be picked first by the workers.

The priority is defined as the approximate time the readahead requests
will be needed in microseconds. This is achieved by putting an extra
(in addition to current timestamp) priority of 0 to the already needed
requests, and an approximated value to the ahead requests.
- try to align the extra readahead requests. The goal is to ask for
full chunks, instead of splitting it into several requests.

### DIFF
--- a/src/mount/readahead_adviser.h
+++ b/src/mount/readahead_adviser.h
@@ -91,15 +91,18 @@ public:
 	}
 
 	/*!
+	* \brief Estimate the time a readahead request will be needed.
+	*/
+	double expectedNeededTime_us(uint64_t aheadSize) {
+		return aheadSize / throughputMBps();
+	}
+
+	/*!
 	 * \brief Estimation of the size of data requested by the process
 	 * considering the last read requests made.
 	 */
 	uint64_t throughputWindow() {
-		int64_t timestamp = timer_.elapsed_us();
-		double throughput_MBps = static_cast<double>(requested_bytes_) /
-		                         (timestamp - history_.front().timestamp);
-
-		return kConservativeMultiplier * throughput_MBps * timeout_ms_
+		return kConservativeMultiplier * throughputMBps() * timeout_ms_
 		       * kBytesInOneKiB;
 	}
 
@@ -141,6 +144,25 @@ public:
 	}
 
 private:
+
+	/*!
+	* \brief Calculates an estimation of the throughput in MB/s considering the current history.
+	* The timestamp that could be provided is to save the call to ```elapsed_us``` if already done.
+	*/
+	double throughputMBps(int64_t timestamp = 0) {
+		if (timestamp == 0) {
+			timestamp = timer_.elapsed_us();
+		}
+		// Make sure non-zero division
+		double throughput_MBps = static_cast<double>(requested_bytes_) /
+		                         std::max<int64_t>(timestamp - history_.front().timestamp, 1);
+
+		if (throughput_MBps <= 0) {
+			safs::log_warn("Throughput must be greater than 0, current value: {}", throughput_MBps);
+			throughput_MBps = kNeutralThroughput;
+		}
+		return throughput_MBps;
+	}
 
 	/*!
 	 * \brief Calculates the absolute difference between two ```uint64_t``` values.
@@ -191,12 +213,10 @@ private:
 	 * \param timestamp time point used for latency estimation
 	 */
 	void adjustMaxWindowSize(int64_t timestamp) {
-		double throughput_MBps =
-		    (double)requested_bytes_ / (timestamp - history_.front().timestamp);
 		// Max window size is set on the basis of estimated throughput
 		max_window_size_ = std::min<uint64_t>(
 		    window_size_limit_,
-		    kConservativeMultiplier * throughput_MBps * timeout_ms_ * 1024);
+		    kConservativeMultiplier * throughputMBps(timestamp) * timeout_ms_ * kBytesInOneKiB);
 		max_window_size_ = std::max(max_window_size_, kInitWindowSize);
 	}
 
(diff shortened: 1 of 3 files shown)


src/mount/readdata.cc
```diff
@@ -287,24 +287,24 @@ bool ReadaheadOperationsManager::request(
 }
 
 Request ReadaheadOperationsManager::nextRequest() {
-	Request request = readaheadRequestContainer_.front();
+	Request request = readaheadRequestContainer_.top().second;
```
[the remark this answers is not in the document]
dmga44: Added a massert check at the beginning of the method because this should not happen. Thanks.

src/mount/readdata.cc
```diff
@@ -339,13 +339,18 @@ void ReadaheadOperationsManager::addExtraRequests_(
...
 
-		ReadCache::Entry *entry = rrec->cache.forceInsert(
-		    maximumRequestedOffset, satisfyingSize);
+		// Try to align extra requests to SFSCHUNKSIZE
+		uint64_t extraRequestSize = std::min<uint64_t>(
+		    satisfyingSize, SFSCHUNKSIZE - (maximumRequestedOffset % SFSCHUNKSIZE));
+		ReadCache::Entry *entry = rrec->cache.forceInsert(maximumRequestedOffset, extraRequestSize);
+
+		int64_t extraPriority =
+		    rrec->readahead_adviser.expectedNeededTime_us(maximumRequestedOffset - currentOffset);
```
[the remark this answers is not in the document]
dmga44: Added a massert check before the substraction because this should not happen. Thanks.

src/mount/readahead_adviser.h
```diff
@@ -90,16 +90,19 @@ class ReadaheadAdviser {
 		feed(offset, size, dummy);
 	}
 
+	/*!
+	* \brief Estimate the time a readahead request will be needed.
+	*/
+	double expectedNeededTime_us(uint64_t aheadSize) {
+		return aheadSize / throughputMBps();
```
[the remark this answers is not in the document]
dmga44: I uploaded a change to ensure that the ```throughputMBps``` return value is always greater than 0. Thanks.

lgsilva3087 (approved):
LGTM

src/mount/readdata.h
```diff
@@ -352,13 +363,16 @@ class ReadaheadOperationsManager {
 	 * scheduled request. If not, this value provides the starting offset to
 	 * add the extra requests.
 	 */
-	void addExtraRequests_(ReadRecord *rrec, uint64_t currentOffset,
-	                       uint64_t satisfyingSize,
+	void addExtraRequests_(ReadRecord *rrec, uint64_t currentOffset, uint64_t satisfyingSize,
```
rolysr, line 366: Is this format change needed here?

src/mount/readdata.h
```diff
@@ -311,10 +311,19 @@ class ReadaheadOperationsManager {
 	/** \brief Insert a request signaling one read worker thread to stop.
 	 *
 	 * requestsLock: UNLOCKED
-	*/
+	 */
```
rolysr, line 314: Is this extra space needed here?

src/mount/readdata.cc
```diff
@@ -339,13 +340,20 @@ void ReadaheadOperationsManager::addExtraRequests_(
...
 
-		ReadCache::Entry *entry = rrec->cache.forceInsert(
-		    maximumRequestedOffset, satisfyingSize);
+		// Try to align extra requests to SFSCHUNKSIZE
+		uint64_t extraRequestSize = std::min<uint64_t>(
+		    satisfyingSize, SFSCHUNKSIZE - (maximumRequestedOffset % SFSCHUNKSIZE));
+		ReadCache::Entry *entry = rrec->cache.forceInsert(maximumRequestedOffset, extraRequestSize);
+
+		massert(maximumRequestedOffset >= currentOffset,
+		        "Maximum requested offset should be greater than current offset");
```
rolysr: Consider here: "greater or equal than" or "not less than"

src/mount/readahead_adviser.h
```diff
@@ -142,6 +145,24 @@ class ReadaheadAdviser {
...
+	double throughputMBps(int64_t timestamp = 0) {
+		if (timestamp == 0) {
+			timestamp = timer_.elapsed_us();
+		}
+		double throughput_MBps =
+		    static_cast<double>(requested_bytes_) / (timestamp - history_.front().timestamp);
+
+		if (throughput_MBps <= 0) {
+			safs::log_warn("Throughput must be greater than 0, current value: {}", throughput_MBps);
+			throughput_MBps = 1;
```
rolysr: Why is value 1 set here in this case? Could you document this or set is as a constant? I could see that the 1 value here could help as a neutral value when calling `throughputWindow()`, is it for that?

src/mount/readahead_adviser.h
```diff
@@ -142,6 +145,24 @@ class ReadaheadAdviser {
...
+	/*!
+	* \brief Calculates an estimation of the throughput in MB/s considering the current history.
+	* The timestamp that could be provided is to save the call to ```elapsed_us``` if already done.
+	*/
+	double throughputMBps(int64_t timestamp = 0) {
+		if (timestamp == 0) {
+			timestamp = timer_.elapsed_us();
+		}
+		double throughput_MBps =
+		    static_cast<double>(requested_bytes_) / (timestamp - history_.front().timestamp);
```
rolysr: Can you always ensure that not zero division issue could appear here?

rolysr (approved):
LGTM. Just a few comments added.

src/mount/readdata.h
dmga44: No, but anyway I had to apply format to the signature of the function, so I did it all together.

src/mount/readdata.h
dmga44: It is needed to align the comment it with the upper lines.

src/mount/readdata.cc
dmga44: Applied.

src/mount/readahead_adviser.h
dmga44: You're right, it is used as a neutral, canonical value for division and multiplication. That's why it was used 1. I'm putting it as a constant.

src/mount/readahead_adviser.h
dmga44: This should not be happening in any case, because the those timestamps have a microsecond resolution. This division was used in code prior this change, so it should not be happening. Anyway, it will be valuable to ensure it. Applied.

PR #319, opened 2025-03-06
feat(chunkserver): Parallelize reads per csentry

dmga44:
This change targets making the ChunkserverEntry capable of sending
several read jobs at a time to the background workers.

The main part of the change was to make the readContinue function
able to send several read jobs into the jobs pool. To make this change
consistent, it was needed to change:
- the way a failing read jobs is being handled.
- the way the ChunkserverEntry is closed.

The job pool structure needed to be extended to fulfill the new
requirements, as well as the output buffer.

In order to configure the new behavior, the option
MAX_PARALLEL_HDD_READ_JOBS_PER_CS_ENTRY was added to the chunkserver
list of options. As a side change, the following constants were also
increased:
- kDefaultMaxBackgroundJobsPerNetworkWorker to make sure the jobs pool
could handle a larger number of parallel requests by default. The
related parameter BGJOBSCNT_PER_NETWORK_WORKER default value was also
changed.
- BuffersPool::kMaxSize to make sure there will be enough room to store
the higher number of buffers used at the same time.

A new test was added to the LongSystem tests suite to check the new parameter.


src/chunkserver/bgjobs.cc
```diff
@@ -420,34 +422,68 @@ void job_pool_disable_and_change_callback_all(void *jpool,void (*callback)(uint8
...
 
-	for (jptr = jp->jobhash[jhpos] ; jptr ; jptr=jptr->next) {
-		if (jptr->jobid==jobid) {
+	for (jptr = jp->jobhash[jhpos]; jptr; jptr = jptr->next) {
+		if (jptr->jobid == jobid) {
 			jobsUniqueLock.lock();
-			if (jptr->jstate==JSTATE_ENABLED) {
-				jptr->jstate=JSTATE_DISABLED;
-			}
+			if (jptr->jstate == JSTATE_ENABLED) { jptr->jstate = JSTATE_DISABLED; }
```
rolysr: Consider to put these indentation changes for all changes on this PR to a separate commit in order to have a cleaner history of the changes.

src/chunkserver/bgjobs.cc
```diff
@@ -420,34 +422,68 @@ void job_pool_disable_and_change_callback_all(void *jpool,void (*callback)(uint8
...
 			jobsUniqueLock.unlock();
 		}
 	}
 }
 
-void job_pool_change_callback(void *jpool,uint32_t jobid,void (*callback)(uint8_t status,void *extra),void *extra) {
+void job_pool_disable_jobs(void *jpool, std::list<uint32_t> &jobIds) {
 	TRACETHIS();
-	jobpool* jp = (jobpool*)jpool;
+	jobpool *jp = (jobpool *)jpool;
```
rolysr: Same here regarding indentation and functional changes commit changes split.

src/chunkserver/bgjobs.cc
```diff
@@ -420,34 +422,68 @@ void job_pool_disable_and_change_callback_all(void *jpool,void (*callback)(uint8
...
-	for (jptr = jp->jobhash[jhpos] ; jptr ; jptr=jptr->next) {
-		if (jptr->jobid==jobid) {
-			jptr->callback=callback;
-			jptr->extra=extra;
+	for (jptr = jp->jobhash[jhpos]; jptr; jptr = jptr->next) {
+		if (jptr->jobid == jobid) {
+			jptr->callback = callback;
+			jptr->extra = extra;
+		}
+	}
```
rolysr: Also here, related consideration with the latest comment.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -361,49 +403,51 @@ void ChunkserverEntry::readContinue() {
...
+				state = State::Close;
+				return;
+			}
+			// readPacket = std::move(packet);
+			pendingReadDataPackets.emplace_back(std::move(packet));
 
-		uint32_t readAheadBlocks = 0;
-		uint32_t maxReadBehindBlocks = 0;
+			uint32_t readAheadBlocks = 0;
+			uint32_t maxReadBehindBlocks = 0;
```
rolysr, line 453: Check for indentation here

src/chunkserver/chunkserver_entry.h
```diff
@@ -116,8 +116,7 @@ struct ChunkserverEntry {
 	static constexpr int kInvalidSocket = -1;
 	static constexpr int kInitConnectionOK = 0;
 	static constexpr int kInitConnectionFailed = -1;
-	static constexpr uint32_t kGenerateChartExpectedPacketSize =
-	    sizeof(uint32_t);
+	static constexpr uint32_t kGenerateChartExpectedPacketSize = sizeof(uint32_t);
```
rolysr: Consider if it would be useful to keep this indentation change.

src/chunkserver/network_main_thread.cc
```diff
@@ -120,6 +120,10 @@ void mainNetworkThreadReload(void) {
 	}
 	chunkReplicatorReload();
 
+	gMaxParallelHddReadJobsPerCsEntry = cfg_get_minvalue<uint16_t>(
+	    "MAX_PARALLEL_HDD_READ_JOBS_PER_CS_ENTRY",
+	    NetworkWorkerThread::kDefaultMaxParallelHddReadJobsPerCsEntry, 1);
```
rolysr, line 125: Consider to name the value `1` here as `kMinParallelHddReadJobsPerCsEntry`

rolysr (approved):
LGTM. Good work @dmga44

src/chunkserver/chunkserver_entry.cc
```diff
@@ -289,31 +290,78 @@ void ChunkserverEntry::delayedCloseCallback(uint8_t status, void *entry) {
...
 		eptr->isChunkOpen = 1;
 	}
 	if (eptr->isChunkOpen) {
 		job_close(eptr->workerJobPool, nullptr, nullptr, eptr->chunkId,
 		          eptr->chunkType);
 		eptr->isChunkOpen = 0;
 	}
-	eptr->state = State::Closed;
+	if (eptr->pendingReadJobIds.empty() && eptr->toDiscardReadJobIds.empty() && eptr->writeJobWriteId == 0) {
+		eptr->state = State::Closed;
```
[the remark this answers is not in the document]
dmga44: That is not the responsability of this function. When this function is called, the `ChunkserverEntry` is only expected to be closing the previously created jobs, that is why the status is ignored. Plus, those jobs were set disabled, which means the status won't be OK.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -289,31 +290,78 @@ void ChunkserverEntry::delayedCloseCallback(uint8_t status, void *entry) {
...
+	}
+}
+
+void ChunkserverEntry::prepareDiscardReadJobs() {
+	// We need to:
+	// - change state of packets not taken by any hdd worker
+	// - change callback in already taken ones
+	// - move packets from pending to toDiscard lists
+	job_pool_disable_jobs(workerJobPool, pendingReadJobIds);
+	job_pool_change_callback(workerJobPool, pendingReadJobIds, readDiscardCallback, this);
```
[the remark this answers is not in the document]
dmga44: The current state of the code ensures that this code is called by the network thread which already created the jobPool. Furthermore, the jobPool actually takes care of protecting the specific fields accessed by these functions.

Regarding the name change, the one selected already follows the rule used for the ChunkserverEntry members.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -289,31 +290,78 @@ void ChunkserverEntry::delayedCloseCallback(uint8_t status, void *entry) {
...
-	eptr->readJobId = 0;
 	if (status == SAUNAFS_STATUS_OK) {
-		eptr->todoReadCounter--;
+		eptr->readyReadDataPackets++;
 		eptr->isChunkOpen = 1;
-		if (eptr->todoReadCounter == 0) {
-			eptr->readContinue();
-		}
+		eptr->readContinue();
 	} else {
```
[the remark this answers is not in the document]
dmga44: > This means that the `eptr->readContinue()` call may be operating on a different state than intended

No. The `readFinishedCallback` is called for non-discarded jobs, which means no errors so far. If there is some error, the jobs and associated data will be released using `eptr->toDiscardReadDataPackets.pop_front();`, and if not, then the `eptr->readContinue()` will be called and the data will be released there.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -1022,16 +1064,16 @@ void ChunkserverEntry::testChunk(const uint8_t *data, uint32_t length) {
...
 void ChunkserverEntry::closeJobs() {
 	TRACETHIS();
-	if (readJobId > 0) {
-		job_pool_disable_job(workerJobPool, readJobId);
-		job_pool_change_callback(workerJobPool, readJobId, delayedCloseCallback,
-		                         this);
+	if (!pendingReadJobIds.empty() || !toDiscardReadJobIds.empty()) {
+		job_pool_disable_jobs(workerJobPool, pendingReadJobIds);
+		job_pool_change_callback(workerJobPool, pendingReadJobIds, delayedCloseCallback, this);
+		job_pool_change_callback(workerJobPool, toDiscardReadJobIds, delayedCloseCallback, this);
```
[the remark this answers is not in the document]
dmga44: `delayedCloseCallback` is called always, regardless of the status of the read jobs.

src/chunkserver/bgjobs.cc
```diff
@@ -277,6 +277,8 @@ void* job_worker(void *th_arg) {
 						rdargs->offset, rdargs->size, rdargs->maxBlocksToBeReadBehind,
 						rdargs->blocksToBeReadAhead, rdargs->outputBuffer);
 
+				rdargs->outputBuffer->isReadCompleted = true;
```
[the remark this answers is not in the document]
dmga44: The goal of the change is not that one.

src/chunkserver/bgjobs.cc
```diff
@@ -420,34 +422,68 @@ void job_pool_disable_and_change_callback_all(void *jpool,void (*callback)(uint8
...
-	for (jptr = jp->jobhash[jhpos] ; jptr ; jptr=jptr->next) {
-		if (jptr->jobid==jobid) {
+	for (jptr = jp->jobhash[jhpos]; jptr; jptr = jptr->next) {
+		if (jptr->jobid == jobid) {
 			jobsUniqueLock.lock();
-			if (jptr->jstate==JSTATE_ENABLED) {
-				jptr->jstate=JSTATE_DISABLED;
-			}
+			if (jptr->jstate == JSTATE_ENABLED) { jptr->jstate = JSTATE_DISABLED; }
 			jobsUniqueLock.unlock();
```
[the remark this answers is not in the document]
dmga44: For this specific function, the lock is called only once, so it makes sense.

For the new one, the lock is called the same number of packets we're working on, which should not be too much.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -324,29 +372,23 @@ void ChunkserverEntry::readFinishedCallback(uint8_t status, void *entry) {
...
-	if (readPacket) {
-		attachPacket(std::move(readPacket));
-		readPacket.reset();
-		todoReadCounter++;
+	while (readyReadDataPackets > 0 && pendingReadDataPackets.front()->outputBuffer->isReadCompleted) {
+		attachPacket(std::move(pendingReadDataPackets.front()));
+		pendingReadDataPackets.pop_front();
+		pendingReadJobIds.pop_front();
+		readyReadDataPackets--;
 	}
```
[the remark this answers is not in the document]
dmga44: `readyReadDataPackets` is only change in code accessed by a single thread, which is the network worker parent of this `ChunkserverEntry`.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -361,49 +403,51 @@ void ChunkserverEntry::readContinue() {
...
-		std::vector<uint8_t> readDataPrefix;
-		messageSerializer->serializePrefixOfCstoclReadData(
-		    readDataPrefix, chunkId, offset, thisPartSize);
-		auto packet = createDetachedPacketWithOutputBuffer(readDataPrefix);
-		if (packet == kInvalidPacket) {
-			state = State::Close;
-			return;
-		}
-		readPacket = std::move(packet);
+		while (size > 0 && pendingReadDataPackets.size() < maxParallelHddReadJobs) {
```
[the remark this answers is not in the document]
dmga44: I think this is not the place to do this, but anyway the lower limit (1) has been set to make sure previous behavior could be achieved. The reasonable upper is the number of block of a chunk, but this is intrinsically enforced.

src/chunkserver/bgjobs.cc
dmga44: I rolled back this changes, so only new lines come with the commit.

src/chunkserver/bgjobs.cc
dmga44: Same as previous comment.

src/chunkserver/bgjobs.cc
dmga44: Same response.

src/chunkserver/chunkserver_entry.cc
dmga44: It is fine :)

src/chunkserver/chunkserver_entry.h
dmga44: I rolled it back too.

src/chunkserver/network_main_thread.cc
dmga44: The rule of the rest of the file for the minimal constants is to not name them, so I kept it. Maybe it could be an idea for other refactor PR.

src/chunkserver/bgjobs.cc
```diff
@@ -439,6 +441,25 @@ void job_pool_disable_job(void *jpool,uint32_t jobid) {
...
+	job *jptr;
+
+	std::unique_lock jobsUniqueLock(jp->jobsMutex, std::defer_lock);
+	for (auto jobid : jobIds) {
+		uint32_t jhpos = JHASHPOS(jobid);
+
+		for (jptr = jp->jobhash[jhpos]; jptr; jptr = jptr->next) {
+			if (jptr->jobid == jobid) {
+				jobsUniqueLock.lock();
+				if (jptr->jstate == JSTATE_ENABLED) { jptr->jstate = JSTATE_DISABLED; }
```
lgsilva3087: Looks like this line is not heavy. Consider locking before the 'for' and unlocking after if the maximum number of expected jobs it not really high (contention could be more costly than iterating).

lgsilva3087 (approved):
LGTM. Please see my suggestion.

src/chunkserver/bgjobs.cc
dmga44: Done.

PR #324, opened 2025-03-08
Refactor chunkserver bgjobs

ralcolea:
This PR refactors the background jobs (bgjobs.{h,cc}) used by the chunkserver.

**Key changes:**
- Add the `JobPool` to manage bgjobs: previously background jobs were
  managed using raw pointers and macros, leading to complex and hard to
  maintain code. The `JobPool` class introduces modern C++ features to
  make the code more efficient, readable and easier to maintain.
  
- Refactor `hddChunkOperation`: this function handled all chunk
  operations using different argument combinations, making the code
  complex and hard to maintain. The new version separates this function
  into multiple functions, each with a single responsibility.
  
- Introduce a `ProcessJobCallback` in the `Job` structure to manage
  job execution. This simplifies the `workerThread` function of every
  job and remove redundant job arguments used in the previous version.
  
- Add documentation and unit tests for the `JobPool` class.

### DIFF
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -49,7 +49,9 @@ constexpr auto kInvalidJob = nullptr;
 
 JobPool::JobPool(uint8_t workers, uint32_t maxJobs, int *wakeupDesc) : workers(workers) {
 	int fd[2];
-	if (pipe(fd) < 0) { throw std::runtime_error("Failed to create pipe"); }
+	if (pipe(fd) < 0) {  // pipe is a critical resource for communication within the JobPool
+		throw std::runtime_error("Failed to create pipe: " + std::string(strerror(errno)));
+	}
 	rpipe = fd[0];
 	wpipe = fd[1];
 
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -44,9 +44,6 @@
 #include <utility>
 #include <vector>
 
-#define JHASHSIZE 0x400
-#define JHASHPOS(id) ((id)&0x3FF)
-
 constexpr auto kInvalidJob = nullptr;
 
 JobPool::JobPool(uint8_t workers, uint32_t maxJobs, int *wakeupDesc) : workers(workers) {
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -177,7 +177,7 @@ void JobPool::workerThread() {
 
 		if (jobState == State::Disabled) {
 			status = SAUNAFS_ERROR_NOTDONE;
-			sendStatus(jobId, status);
+			jobsUniqueLock.unlock();
 			continue;
 		}
 
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -47,10 +47,6 @@
 #define JHASHSIZE 0x400
 #define JHASHPOS(id) ((id)&0x3FF)
 
-/// This is a special value used to determine the chunk operation that will be executed
-/// by \ref hddChunkOperation function.
-constexpr uint32_t kWildCardLength = 0xFFFFFFFF;
-
 constexpr auto kInvalidJob = nullptr;
 
 
@@ -263,14 +259,51 @@ void JobPool::workerThread() {
 			status = SAUNAFS_ERROR_EINVAL;
 			break;
 		}
-		case ChunkOperation::ChunkOp:
+		case ChunkOperation::Delete:
 		{
 			auto args =
 			    std::unique_ptr<chunk_chunkop_args>(static_cast<chunk_chunkop_args *>(job->args));
-			// TODO(Crash): Refactor hddChunkOperation and separate operations and states
-			status =
-			    hddChunkOperation(args->chunkid, args->version, args->chunkType, args->newversion,
-			                      args->copychunkid, args->copyversion, args->length);
+			status = hddInternalDelete(args->chunkid, args->version, args->chunkType);
+			break;
+		}
+		case ChunkOperation::Create:
+		{
+			auto args =
+			    std::unique_ptr<chunk_chunkop_args>(static_cast<chunk_chunkop_args *>(job->args));
+			status = hddInternalCreate(args->chunkid, args->version, args->chunkType);
+			break;
+		}
+		case ChunkOperation::ChangeVersion:
+		{
+			auto args =
+			    std::unique_ptr<chunk_chunkop_args>(static_cast<chunk_chunkop_args *>(job->args));
+			status = hddInternalUpdateVersion(args->chunkid, args->version, args->newversion,
+			                                  args->chunkType);
+			break;
+		}
+		case ChunkOperation::Truncate:
+		{
+			auto args =
+			    std::unique_ptr<chunk_chunkop_args>(static_cast<chunk_chunkop_args *>(job->args));
+			status = hddTruncate(args->chunkid, args->version, args->chunkType, args->newversion,
+			                     args->length);
+			break;
+		}
+		case ChunkOperation::Duplicate:
+		{
+			auto args =
+			    std::unique_ptr<chunk_chunkop_args>(static_cast<chunk_chunkop_args *>(job->args));
+			status = hddDuplicate(args->chunkid, args->version, args->newversion, args->chunkType,
+			                      args->copychunkid, args->copyversion);
+			break;
+		}
+		case ChunkOperation::DuplicateTruncate:
+		{
+			auto args =
+			    std::unique_ptr<chunk_chunkop_args>(static_cast<chunk_chunkop_args *>(job->args));
+			status = hddDuplicateTruncate(args->chunkid, args->version, args->newversion,
+			                              args->chunkType, args->copychunkid, args->copyversion,
+			                              args->length);
 			break;
 		}
 		case ChunkOperation::Open:
@@ -471,43 +504,41 @@ uint32_t job_replicate(JobPool &jobPool, JobPool::JobCallback callback, void *ex
 	                      extra);
 }
 
-uint32_t job_chunkop(JobPool &jobPool, JobPool::JobCallback callback, void *extra, uint64_t chunkId,
-                     uint32_t chunkVersion, ChunkPartType chunkType, uint32_t newChunkVersion,
-                     uint64_t copyChunkId, uint32_t copyChunkVersion, uint32_t length) {
-    auto args = std::make_unique<chunk_chunkop_args>();
-    args->chunkid = chunkId;
-    args->version = chunkVersion;
-    args->newversion = newChunkVersion;
-    args->copychunkid = copyChunkId;
-    args->copyversion = copyChunkVersion;
-    args->length = length;
-    args->chunkType = chunkType;
-	return jobPool.addJob(JobPool::ChunkOperation::ChunkOp, args.release(), std::move(callback),
-	                      extra);
-}
-
 uint32_t job_invalid(JobPool &jobPool, JobPool::JobCallback callback, void *extra) {
 	return jobPool.addJob(JobPool::ChunkOperation::Invalid, nullptr, std::move(callback), extra);
 }
 
 uint32_t job_delete(JobPool &jobPool, JobPool::JobCallback callback, void *extra, uint64_t chunkId,
 	uint32_t chunkVersion, ChunkPartType chunkType) {
-	return job_chunkop(jobPool, std::move(callback), extra, chunkId, chunkVersion, chunkType, 0, 0,
-	                   0, 0);
+	auto args = std::make_unique<chunk_chunkop_args>();
+	args->chunkid = chunkId;
+	args->version = chunkVersion;
+	args->chunkType = chunkType;
+	return jobPool.addJob(JobPool::ChunkOperation::Delete, args.release(), std::move(callback),
+	                      extra);
 }
 
 uint32_t job_create(JobPool &jobPool, JobPool::JobCallback callback, void *extra, uint64_t chunkId,
 	uint32_t chunkVersion, ChunkPartType chunkType) {
-	return job_chunkop(jobPool, std::move(callback), extra, chunkId, chunkVersion, chunkType, 0, 0,
-	                   0, 1);
+	auto args = std::make_unique<chunk_chunkop_args>();
+	args->chunkid = chunkId;
+	args->version = chunkVersion;
+	args->chunkType = chunkType;
+	return jobPool.addJob(JobPool::ChunkOperation::Create, args.release(), std::move(callback),
+	                      extra);
 }
 
 uint32_t job_version(JobPool &jobPool, const JobPool::JobCallback &callback, void *extra,
                      uint64_t chunkId, uint32_t chunkVersion, ChunkPartType chunkType,
                      uint32_t newChunkVersion) {
 	if (newChunkVersion > 0) {
-		return job_chunkop(jobPool, callback, extra, chunkId, chunkVersion, chunkType,
-		                   newChunkVersion, 0, 0, kWildCardLength);
+		auto args = std::make_unique<chunk_chunkop_args>();
+		args->chunkid = chunkId;
+		args->version = chunkVersion;
+		args->newversion = newChunkVersion;
+		args->chunkType = chunkType;
+		return jobPool.addJob(JobPool::ChunkOperation::ChangeVersion, args.release(), callback,
+		                      extra);
 	}
 	return job_invalid(jobPool, callback, extra);
 }
@@ -515,9 +546,14 @@ uint32_t job_version(JobPool &jobPool, const JobPool::JobCallback &callback, voi
 uint32_t job_truncate(JobPool &jobPool, const JobPool::JobCallback &callback, void *extra,
 	uint64_t chunkId, ChunkPartType chunkType, uint32_t chunkVersion,
 	uint32_t newChunkVersion, uint32_t length) {
-	if (newChunkVersion > 0 && length != kWildCardLength) {
-		return job_chunkop(jobPool, callback, extra, chunkId, chunkVersion, chunkType,
-		                   newChunkVersion, 0, 0, length);
+	if (newChunkVersion > 0) {
+		auto args = std::make_unique<chunk_chunkop_args>();
+		args->chunkid = chunkId;
+		args->version = chunkVersion;
+		args->newversion = newChunkVersion;
+		args->length = length;
+		args->chunkType = chunkType;
+		return jobPool.addJob(JobPool::ChunkOperation::Truncate, args.release(), callback, extra);
 	}
 	return job_invalid(jobPool, callback, extra);
 }
@@ -525,9 +561,15 @@ uint32_t job_truncate(JobPool &jobPool, const JobPool::JobCallback &callback, vo
 uint32_t job_duplicate(JobPool &jobPool, const JobPool::JobCallback &callback, void *extra,
                        uint64_t chunkId, uint32_t chunkVersion, uint32_t newChunkVersion,
                        ChunkPartType chunkType, uint64_t chunkIdCopy, uint32_t chunkVersionCopy) {
-	if (newChunkVersion > 0 && chunkIdCopy > 0) {
-		return job_chunkop(jobPool, callback, extra, chunkId, chunkVersion, chunkType,
-		                   newChunkVersion, chunkIdCopy, chunkVersionCopy, kWildCardLength);
+	if (newChunkVersion > 0) {
+		auto args = std::make_unique<chunk_chunkop_args>();
+		args->chunkid = chunkId;
+		args->version = chunkVersion;
+		args->newversion = newChunkVersion;
+		args->chunkType = chunkType;
+		args->copychunkid = chunkIdCopy;
+		args->copyversion = chunkVersionCopy;
+		return jobPool.addJob(JobPool::ChunkOperation::Duplicate, args.release(), callback, extra);
 	}
 	return job_invalid(jobPool, callback, extra);
 }
@@ -536,9 +578,17 @@ uint32_t job_duptrunc(JobPool &jobPool, const JobPool::JobCallback &callback, vo
                       uint64_t chunkId, uint32_t chunkVersion, uint32_t newChunkVersion,
                       ChunkPartType chunkType, uint64_t chunkIdCopy, uint32_t chunkVersionCopy,
                       uint32_t length) {
-	if (newChunkVersion > 0 && chunkIdCopy > 0 && length != kWildCardLength) {
-		return job_chunkop(jobPool, callback, extra, chunkId, chunkVersion, chunkType,
-		                   newChunkVersion, chunkIdCopy, chunkVersionCopy, length);
+	if (newChunkVersion > 0) {
+		auto args = std::make_unique<chunk_chunkop_args>();
+		args->chunkid = chunkId;
+		args->version = chunkVersion;
+		args->newversion = newChunkVersion;
+		args->copychunkid = chunkIdCopy;
+		args->copyversion = chunkVersionCopy;
+		args->length = length;
+		args->chunkType = chunkType;
+		return jobPool.addJob(JobPool::ChunkOperation::DuplicateTruncate, args.release(), callback,
+		                      extra);
 	}
 	return job_invalid(jobPool, callback, extra);
 }
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -49,75 +49,6 @@
 
 constexpr auto kInvalidJob = nullptr;
 
-
-// for ChunkOp
-struct chunk_chunkop_args {
-	uint64_t chunkid,copychunkid;
-	uint32_t version,newversion,copyversion;
-	uint32_t length;
-	ChunkPartType chunkType;
-};
-
-// for Open and Close
-struct chunk_open_and_close_args {
-	uint64_t chunkid;
-	ChunkPartType chunkType;
-};
-
-// for Read
-struct chunk_read_args {
-	uint64_t chunkid;
-	uint32_t version;
-	ChunkPartType chunkType;
-	uint32_t offset,size;
-	uint8_t *crcbuff;
-	uint32_t maxBlocksToBeReadBehind;
-	uint32_t blocksToBeReadAhead;
-	OutputBuffer* outputBuffer;
-	bool performHddOpen;
-};
-
-// for Prefetch
-struct chunk_prefetch_args {
-	uint64_t chunkid;
-	uint32_t version;
-	ChunkPartType chunkType;
-	uint32_t firstBlock;
-	uint32_t nrOfBlocks;
-};
-
-// for Write
- struct chunk_write_args {
-	uint64_t chunkId;
-	uint32_t chunkVersion;
-	ChunkPartType chunkType;
-	uint16_t blocknum;
-	uint32_t offset, size;
-	uint32_t crc;
-	const uint8_t *buffer;
-};
-
-struct chunk_get_blocks_args {
-	uint64_t chunkId;
-	uint32_t chunkVersion;
-	ChunkPartType chunkType;
-	uint16_t* blocks;
-};
-
-struct chunk_legacy_replication_args {
-	uint64_t chunkid;
-	uint32_t version;
-	uint8_t srccnt;
-};
-
-struct chunk_replication_args {
-	uint64_t chunkId;
-	uint32_t chunkVersion;
-	ChunkPartType chunkType;
-	uint32_t sourcesBufferSize;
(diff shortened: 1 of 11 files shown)


src/chunkserver/bgjobs.cc
```diff
@@ -42,608 +30,358 @@
...
-	uint32_t jhpos = JHASHPOS(jobid);
-	job *jptr;
+void JobPool::changeCallback(uint32_t jobId, JobCallback callback, void *extra) {
+	std::unique_lock lock(jobsMutex);
+	auto jobIterator = jobHash.find(jobId);
+	if (jobIterator != jobHash.end()) {
+		jobIterator->second->callback = std::move(callback);
+		jobIterator->second->extra = extra;
+	}
+}
```
dmga44: Comparing with the previous code, you are adding an extra locking in this version (versus previous `job_pool_change_callback`). The reason behind previously not using it (from my understanding) is that only a single thread (the thread creating the jobPool) was accessing those members of the `job` struct (the workers were only accessing `state` and `args` fields). So, please consider not doing any locking in this function.

src/chunkserver/bgjobs.cc
```diff
@@ -42,608 +30,358 @@
...
-				jptr->jstate=JSTATE_DISABLED;
-			}
-			jptr->callback=callback;
+	auto jobIterator = jobHash.find(jobId);
+	if (jobIterator != jobHash.end()) {
+		auto callback = jobIterator->second->callback;
+		if (callback) {
+			callback(status, jobIterator->second->extra);
 		}
+		jobHash.erase(jobIterator);
```
dmga44: Comparing this function with the previous `job_pool_check_jobs`, I see a main difference, maybe I'm wrong. I think this function is only processing a single job status at a time, while the base one was cycling until the `statusQueue` was actually empty. Please check if this is true, and if so what are the consecuences.

src/chunkserver/bgjobs.cc
```diff
@@ -42,608 +30,358 @@
...
+			jobState = job->state;
+			if (job->state == State::Enabled) { job->state = State::InProgress; }
+		}
+
+		if (operation == ChunkOperation::Exit) { break; }
+
+		if (jobState == State::Disabled) {
+			status = SAUNAFS_ERROR_NOTDONE;
 			jobsUniqueLock.unlock();
+			continue;
```
dmga44: Previous version of the code was sending even the statuses of the disabled jobs, which is `SAUNAFS_ERROR_NOTDONE`, is this happening here?

P.D: I've seen a commit solving a deadlock in these lines. Maybe this code solves it as well?:
```suggestion
		if (jobState == State::Disabled) {
			status = SAUNAFS_ERROR_NOTDONE;
			jobsUniqueLock.unlock();
			sendStatus(jobId, status);
			continue;
```

src/chunkserver/bgjobs.cc
```diff
@@ -42,608 +30,358 @@
...
-	free(jp->workerthreads);
-	close(jp->rpipe);
-	close(jp->wpipe);
-	delete jp;
+		if (performHddOpen && status != SAUNAFS_STATUS_OK) {
+			int ret = hddClose(chunkId, chunkType);
+			if (ret != SAUNAFS_STATUS_OK) {
+				safs_silent_syslog(LOG_ERR,
+				                   "read job: cannot close chunk after read error (%s): %s",
+				                   saunafs_error_string(status), saunafs_error_string(ret));
```
dmga44: Consider using newer version of our logging library `safs::log_err()` in this case, for instance.

src/chunkserver/bgjobs_unittest.cc
```diff
@@ -0,0 +1,141 @@
...
+TEST_F(JobPoolTest, DisableJob) {
+	uint32_t jobId =
+	    jobPool->addJob(JobPool::ChunkOperation::Read, mockJobCallback, &counter, mockProcessJob);
+	jobPool->disableJob(jobId);
+
+	// Wait for the job to be processed
+	std::this_thread::sleep_for(std::chrono::milliseconds(kWaitTimeMs));
+
+	EXPECT_EQ(counter.load(), 0);
+}
```
dmga44: Please check if after applying suggestion of sending status of disabled jobs this test holds true.

src/chunkserver/hddspacemgr.cc
```diff
@@ -2016,53 +2016,22 @@ int hddInternalDelete(uint64_t chunkId, uint32_t version,
...
+                 ChunkPartType chunkType, uint64_t copyChunkId, uint32_t copyChunkVersion) {
+	return hddInternalDuplicate(chunkId, chunkVersion, chunkNewVersion, chunkType, copyChunkId,
+	                            copyChunkVersion);
+}
+
+int hddDuplicateTruncate(uint64_t chunkId, uint32_t chunkVersion, uint32_t chunkNewVersion,
+                         ChunkPartType chunkType, uint64_t copyChunkId, uint32_t copyChunkVersion,
+                         uint32_t length) {
+	return hddInternalDuplicateTruncate(chunkId, chunkVersion, chunkNewVersion, chunkType,
+	                                    copyChunkId, copyChunkVersion, length);
```
dmga44, line 2034: Why create compact versions of the same functions for only these ones? Why not for `hddInternalUpdateVersion`, `hddInternalDelete` and `hddInternalCreate`?

My point is why in the middle, why not all of them or none of them?

src/chunkserver/network_worker_thread.cc
```diff
@@ -318,8 +317,17 @@ void NetworkWorkerThread::addConnection(int newSocketFD) {
...
+	try {
+		if (!bgJobPool_) { throw std::runtime_error("JobPool instance is null"); }
+		csservEntries.emplace_front(newSocketFD, bgJobPool_.get());
+		csservEntries.front().lastActivity = eventloop_time();
+	} catch (const std::runtime_error &e) {
+		safs_pretty_syslog(LOG_ERR, "Failed to add connection: %s", e.what());
+		tcpclose(newSocketFD);
+	} catch (const std::exception &e) {
+		safs_pretty_syslog(LOG_ERR, "Unexpected error: %s", e.what());
+		tcpclose(newSocketFD);
```
dmga44: Though this file is not the main target of the refactor, please consider using `safs::log_err()`.

dmga44 (commented):
Great job @ralcolea !

Please check my comments, some of them actually worry me 😅 . I'm saving my approve until then :)

src/chunkserver/bgjobs.cc
ralcolea: Nice catch!! Your reasoning is right :sweat_smile:
Done.

src/chunkserver/bgjobs.cc
ralcolea: You're right. The new `checkJobs()` function processes a single job status every time is called.
The previous version processed status jobs until `statusQueue` became empty.

However, I did some tests and `receiveStatus()` function was always called only once. That's why I changed the behavior. My theory is that the poll mechanism works so fast, that even processing a single job every time is enough.

Anyways, for safety, I will revert the implementation to use the same logic as in the original function.

src/chunkserver/bgjobs.cc
ralcolea: You're right. The previous version also sent `status` for jobs having `Disabled` status. It means that `Disabled` jobs are processed the same as jobs in other states. My reasoning behind this change was that `Disabled` jobs should not be processed. That's why I removed the line sending `status` for jobs in `Disabled` state.

Maybe, my assumption was wrong and the correct solution is to keep the previous behavior by following your suggestion. I noticed this after adding the unit test for `disableJobs()` and at this moment, I decided to change it. 

Considering that I'm not sure whether my assumption is correct, I will add your suggestion to keep the original behavior of the system and I will update the unit test for `DisableJob`.

Thank you for pointing out @dmga44  :+1:

src/chunkserver/bgjobs.cc
ralcolea: Done.

src/chunkserver/bgjobs_unittest.cc
ralcolea: After sending status for disabled jobs, the original test will not pass.
Considering that job callback will run also for disabled jobs, I will modify the test to validate that `SAUNAFS_ERROR_NOTDONE` is received by the callback :)

src/chunkserver/hddspacemgr.cc
ralcolea: `hddInternalTruncate()`, `hddInternalDuplicate()` and `hddInternalDuplicateTruncate()` functions were not accessible from bgjobs files because they are only defined in `hddspacemgr.cc`.

Therefore, I created new functions in the header (`hddspacemgr.h`) to allow access to these functions from bgjobs files and apply the least number of changes in `hddspacemgr` files.

src/chunkserver/network_worker_thread.cc
ralcolea: Done.

dmga44 (approved):

src/chunkserver/network_worker_thread.cc
```diff
@@ -318,8 +317,17 @@ void NetworkWorkerThread::addConnection(int newSocketFD) {
...
+	}
```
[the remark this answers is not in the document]
ralcolea: The current implementation does not throw `JobPoolException`. Therefore, there is no point of catching exceptions that will never arrive :)

src/chunkserver/bgjobs.cc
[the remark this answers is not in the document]
ralcolea: The current logic handles all the scenarios correctly.

src/chunkserver/bgjobs.cc
```diff
@@ -42,608 +30,358 @@
...
+constexpr auto kInvalidJob = nullptr;
 
-	if (jp->statusQueue->isEmpty()) {   // first status
-		eassert(write(jp->wpipe,&status,1)==1); // write anything to wake up select
-	}
-	jp->statusQueue->put(jobid, status, nullptr, 1);
-}
+JobPool::JobPool(uint8_t workers, uint32_t maxJobs, int *wakeupDesc) : workers(workers) {
+	int fd[2];
+	if (pipe(fd) < 0) { throw std::runtime_error("Failed to create pipe"); }
```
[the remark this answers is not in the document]
ralcolea: Done.

src/chunkserver/masterconn.cc
```diff
@@ -310,7 +303,7 @@ void masterconn_create(masterconn */*eptr*/, const std::vector<uint8_t> &data) {
 	matocs::createChunk::deserialize(data, chunkId, chunkType, chunkVersion);
 	OutputPacket *outputPacket = new OutputPacket;
 	cstoma::createChunk::serialize(outputPacket->packet, chunkId, chunkType, SAUNAFS_STATUS_OK);
-	job_create(jpool, masterconn_saujobfinished, outputPacket, chunkId, chunkType, chunkVersion);
+	job_create(*jobPool, masterconn_saujobfinished, outputPacket, chunkId, chunkVersion, chunkType);
```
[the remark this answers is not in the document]
ralcolea: Done.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -391,10 +388,8 @@ void ChunkserverEntry::readContinue() {
 			maxReadBehindBlocks = std::min(totalRequestBlocks,
 					gHDDReadAhead.maxBlocksToBeReadBehind());
 		}
-
-		readJobId = job_read(workerJobPool, readFinishedCallback, this, chunkId,
-		                     chunkVersion, chunkType, offset, thisPartSize,
-		                     maxReadBehindBlocks, readAheadBlocks,
+		readJobId = job_read(*workerJobPool, readFinishedCallback, this, chunkId, chunkVersion,
+		                     chunkType, offset, thisPartSize, maxReadBehindBlocks, readAheadBlocks,
 		                     readPacket->outputBuffer.get(), !isChunkOpen);
```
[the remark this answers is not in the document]
ralcolea: The current implementation contains a lot of cases like this and this validation is outside the scope of this PR. Maybe in future PRs these suggestions can be addressed in a PR created to prevent potential null pointer dereferences.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -1029,23 +1018,20 @@
 void ChunkserverEntry::closeJobs() {
 	TRACETHIS();
 	if (readJobId > 0) {
-		job_pool_disable_job(workerJobPool, readJobId);
-		job_pool_change_callback(workerJobPool, readJobId, delayedCloseCallback,
-		                         this);
+		workerJobPool->disableJob(readJobId);
+		workerJobPool->changeCallback(readJobId, delayedCloseCallback, this);
 		state = State::CloseWait;
```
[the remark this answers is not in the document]
ralcolea: This suggestion will be addressed in future PRs.

src/chunkserver/bgjobs.cc
```diff
@@ -42,608 +30,358 @@
...
+	for (uint8_t i = 0; i < workers; ++i) {
+		jobsQueue->put(0, JobPool::ChunkOperation::Exit, nullptr, 1);
+	}
 
-	static std::atomic_uint16_t workersCounter(0);
-	std::string threadName = "jobWorker " + std::to_string(workersCounter++);
-	pthread_setname_np(pthread_self(), threadName.c_str());
+	for (auto &thread : workerThreads) {
+		if (thread.joinable()) { thread.join(); }
+	}
```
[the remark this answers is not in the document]
ralcolea: This deallocation is not needed, as the `Job` resources are deallocated correctly when they are removed from `jobsQueue` and `jobHash` data structures.

src/chunkserver/bgjobs.h
```diff
@@ -22,81 +22,342 @@
...
+		Open,
+		Close,
+		Read,
+		Prefetch,
+		Write,
+		Replicate,
+		GetBlocks
+	};
+
+public:
```
uristdwarf: nit: Two publics

src/chunkserver/masterconn.cc
```diff
@@ -321,7 +322,13 @@ void masterconn_delete(masterconn */*eptr*/, const std::vector<uint8_t>& data) {
...
 	OutputPacket* outputPacket = new OutputPacket;
 	cstoma::deleteChunk::serialize(outputPacket->packet, chunkId, chunkType, 0);
-	job_delete(jpool, masterconn_saujobfinished, outputPacket, chunkId, chunkVersion, chunkType);
+	if (jobPool) {
+		job_delete(*jobPool, masterconn_saujobfinished, outputPacket, chunkId, chunkVersion,
+		           chunkType);
+	} else {
+		safs::log_err("masterconn_delete: jobPool is null.");
+		delete outputPacket;
+	}
```
uristdwarf, line 331: What scenarios do you think where jobPool would be null? It would seem something has gone terribly wrong if it's null, but I'm not sure.

ralcolea: Although unlikely, I'm thinking in the scenario where the pipe creation fails.
In this case, an exception is thrown and `jobPool` instance is not created correctly.
For sure, there are other scenarios, but this is the first that comes to my mind :)

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
+#include <utility>
+#include <vector>
+
+constexpr auto kInvalidJob = nullptr;
+
+JobPool::JobPool(uint8_t workers, uint32_t maxJobs, int *wakeupDesc) : workers(workers) {
+	int fd[2];
+	if (pipe(fd) < 0) {  // pipe is a critical resource for communication within the JobPool
+		throw std::runtime_error("Failed to create pipe: " + std::string(strerror(errno)));
+	}
```
rolysr: Consider to add here in the run time error message something specifying that the error here is related to the `JobPool`. you could do something like: `throw std::runtime_error("JobPool: Failed to create pipe: " + std::string(strerror(errno)));` or similar for faster debugging or error troubleshooting. Consider to add this also in some other parts of the newly added code.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
+	int notLastJob = 1;
+	do {
+		notLastJob = receiveStatus(jobId, status);
+		auto jobIterator = jobHash.find(jobId);
+		if (jobIterator != jobHash.end()) {
+			auto callback = jobIterator->second->callback;
+			if (callback) { callback(status, jobIterator->second->extra); }
+			jobHash.erase(jobIterator);
+		}
+	} while (notLastJob > 0);
```
rolysr: Consider changing this by:
```suggestion
	} while (notLastJob);
```
this is just an optional suggestion in case it will fit future behavior of `receiveStatus` function.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
 
-	jobpool *jp = (jobpool*)th_arg;
-	job *jptr;
-	uint8_t *jptrarg;
-	uint8_t status, jstate;
-	uint32_t jobid;
-	uint32_t op;
+	uint32_t jobId;
+	uint32_t operation;
+	uint8_t *jobPtrArg;
```
rolysr, line 159: Consider finding a way to change this raw pointer as a smart pointer if possible according to the use case here.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
+	State jobState;
+	uint8_t status = SAUNAFS_STATUS_OK;
+
+	std::unique_lock jobsUniqueLock(jobsMutex, std::defer_lock);
 
-	std::unique_lock jobsUniqueLock(jp->jobsMutex, std::defer_lock);
+	while (true) {
+		jobsQueue->get(&jobId, &operation, &jobPtrArg, nullptr);
+		Job *job = reinterpret_cast<Job *>(jobPtrArg);
```
rolysr, line 167: Also consider applying last suggestion here too.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
-	return jp;
 }
 
-uint32_t job_pool_jobs_count(void *jpool) {
-	TRACETHIS();
-	jobpool* jp = (jobpool*)jpool;
-	return jp->jobsQueue->elements();
+void JobPool::sendStatus(uint32_t jobId, uint8_t status) {
+	std::lock_guard pipeLockGuard(pipeMutex);
+	if (statusQueue->isEmpty()) { eassert(write(wpipe, &status, 1) == 1); }
```
rolysr: Consider substituting this `1` value by a macro or constant like `STATUS_BYTE_SIZE` and substituting this code by something like: `eassert(write(wpipe, &status, STATUS_BYTE_SIZE) == STATUS_BYTE_SIZE && "JobPool: SendStatus: Failed to write status to pipe");` in order to make it easier to debug if something goes wrong when writing to the pipe.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
-			}
-			jptr->callback=callback;
-		}
+int JobPool::receiveStatus(uint32_t &jobId, uint8_t &status) {
+	uint32_t qstatus = 0;
+	std::lock_guard pipeLockGuard(pipeMutex);
+	statusQueue->get(&jobId, &qstatus, nullptr, nullptr);
+	status = qstatus;
+	if (statusQueue->isEmpty()) {
+		eassert(read(rpipe, &qstatus, 1) == 1);
```
rolysr: Consider doing something similar to previous comment if you consider it necessary.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
-	uint32_t i;
-
-	for (i = 0; i < jp->workers; i++) {
-		jp->jobsQueue->put(0, OP_EXIT, nullptr, 1);
-	}
-
-	for (i = 0; i < jp->workers; i++) {
-		zassert(pthread_join(jp->workerthreads[i], NULL));
-	}
+uint32_t job_prefetch(JobPool &jobPool, uint64_t chunkId, uint32_t  /*version*/, ChunkPartType chunkType,
```
rolysr: Why is still needed to pass the `version` as parameter here if it is not used inside of the function? I saw that there is only one place in the code where this is called and the parameter is passed, but anyway, it is never used.

src/chunkserver/bgjobs.cc
```diff
@@ -41,609 +29,360 @@
...
+	uint32_t firstBlockToBePrefetched, uint32_t blocksToBePrefetched) {
+	JobPool::ProcessJobCallback processJob = [=]() -> uint8_t {
+		return hddPrefetchBlocks(chunkId, chunkType, firstBlockToBePrefetched,
+		                         blocksToBePrefetched);
+	};
+	return jobPool.addJob(JobPool::ChunkOperation::Prefetch, nullptr, nullptr, processJob);
```
rolysr: Consider substituting both `nullptr` values here by descriptive constants of the parameter for better understanding what is being passed to the function.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -294,8 +294,7 @@ void ChunkserverEntry::delayedCloseCallback(uint8_t status, void *entry) {
 		eptr->isChunkOpen = 1;
 	}
 	if (eptr->isChunkOpen) {
-		job_close(eptr->workerJobPool, nullptr, nullptr, eptr->chunkId,
-		          eptr->chunkType);
+		job_close(*eptr->workerJobPool, nullptr, nullptr, eptr->chunkId, eptr->chunkType);
```
rolysr: Consider doing something similar to the previous comment here, or you can also move the order of these to nullptr parameters to the end of the function signature and declared them as default on the definition of the `job_close` function, consider doing these in other occurrences of similar functions.

src/chunkserver/chunkserver_entry.h
```diff
@@ -119,7 +120,7 @@ struct ChunkserverEntry {
 	static constexpr uint32_t kGenerateChartExpectedPacketSize =
 	    sizeof(uint32_t);
 
-	void* workerJobPool; // Job pool assigned to a given network worker thread
+	JobPool *workerJobPool;  // Job pool assigned to a given network worker thread
```
rolysr, line 123: Consider if it is suitable to make this raw pointer a smart pointer in this case.

src/chunkserver/masterconn.cc
```diff
@@ -553,7 +585,7 @@ void masterconn_read(masterconn *eptr) {
 
 	watchdog.start();
 	while (eptr->mode != KILL) {
-		if (job_pool_jobs_count(jpool) >= (BGJOBSCNT * 9) / 10) {
+		if (jobPool->getJobCount() >= (BGJOBSCNT * 9) / 10) {
```
rolysr, line 588: Consider setting these `9` and `10` values to named constants or consider adding a comment here for better understanding of the use of them here.

src/chunkserver/masterconn.cc
```diff
@@ -832,9 +864,17 @@ int masterconn_init(void) {
...
+	} catch (const std::runtime_error &e) {
+		safs::log_err("masterconn_init_threads: Failed to create JobPool instance: {}", e.what());
+		return -1;
+	} catch (const std::exception &e) {
+		safs::log_err("masterconn_init_threads: Failed to create JobPool instance: {}", e.what());
+		return -1;
+	}
 
-	if (jpool == nullptr) { return -1; }
+	if (jobPool == nullptr) { return -1; }
```
rolysr: Consider substituting this `-1` for some specific error name in case it has sense here for further use of this value or also adding an error log here like the ones use before in this same function.

src/chunkserver/network_worker_thread.cc
```diff
@@ -64,8 +64,15 @@ NetworkWorkerThread::NetworkWorkerThread(uint32_t nrOfBgjobsWorkers,
...
-	    job_pool_new(nrOfBgjobsWorkers, bgjobsCount, &bgJobPoolWakeUpFd_);
+	try {
+		bgJobPool_ = std::make_unique<JobPool>(nrOfBgjobsWorkers, bgjobsCount, &bgJobPoolWakeUpFd_);
+	} catch (const std::runtime_error &e) {
+		safs::log_err("NetworkWorkerThread: Failed to create JobPool instance: {}", e.what());
+		throw;
+	} catch (const std::exception &e) {
+		safs::log_err("NetworkWorkerThread: Failed to create JobPool instance: {}", e.what());
+		throw;
+	}
```
rolysr: Given that both exception catches here do the same independently from the exception type, consider if it would be better to keep them both as just one exception like: 
`catch (const std::exception &e) {
		safs::log_err("NetworkWorkerThread: Failed to create JobPool instance: {}", e.what());
		throw;
}`

rolysr (approved):
Great job @ralcolea ! Just a few style/readability suggestions for the change.

src/chunkserver/bgjobs.cc
ralcolea: Done.

src/chunkserver/bgjobs.cc
ralcolea: This was my first implementation. I changed to avoid clangd warning `Implicit conversion 'int' -> 'bool'`.

src/chunkserver/bgjobs.cc
ralcolea: IMO, smart pointers is not convenient for `jobsQueue` because usually, the raw pointer which comes in the queue is owned by another data structure and is not responsibility of the `jobsQueue` to deallocate this memory or take  ownership of this pointer. Additionally, this could cause changes in `ProducerConsumerQueue` class, which is out of the scope for this PR.

src/chunkserver/bgjobs.cc
ralcolea: In this case, `jobsQueue` stores a pointer of the original `Job`, which is already a smart pointer stored in `jobHash`.

src/chunkserver/bgjobs.cc
ralcolea: Done.

src/chunkserver/bgjobs.cc
ralcolea: Done.

src/chunkserver/bgjobs.cc
ralcolea: I did not want to change the original signature of the function despite it was not used.
I will remove `version` parameter.

src/chunkserver/bgjobs.cc
dmga44: Then, why not change the `notLastJob` type as well, altogether with the return type of `receiveStatus`?

src/chunkserver/bgjobs.cc
ralcolea: Done.

src/chunkserver/chunkserver_entry.cc
ralcolea: Done.

src/chunkserver/chunkserver_entry.h
ralcolea: `ChunkserverEntry` uses a raw pointer because the original `JobPool`, which is already a smart pointer is defined in the `NetworkWorkerThread` class and is shared between all the `ChunkserverEntry` instances.

src/chunkserver/masterconn.cc
ralcolea: I'm not completely sure about the reasoning behind these values.
My best guess is that `9` and `10` represents `90%` of `BGJOBSCNT` and the validation is done to avoid adding more jobs if the `JobPool` is reaching its limit, but that's just my theory :)

src/chunkserver/masterconn.cc
ralcolea: I added an error message with more information about the error.
I will not address the suggestion about replacing `-1` because this value is used as error code almost in every function of `masterconn.cc`. It does not make sense to change it only in this function.
This change can be addressed in another PR.

src/chunkserver/network_worker_thread.cc
ralcolea: Done.

src/chunkserver/bgjobs.cc
ralcolea: I changed to `bool` the type of `notLastJob` as well as the return type of `receiveStatus`, as both suggested.

src/chunkserver/masterconn.cc
uristdwarf: I wonder whether it's better to throw an exception or not. Either way I'd set the error level to critical.

src/chunkserver/bgjobs.h
ralcolea: Removed unneeded public :)

src/chunkserver/masterconn.cc
ralcolea: Maybe in the future we can change it, as well as the error level. However, I don't think these are stoppers to merge the changes :)

PR #326, opened 2025-03-11
feat(mount): Add mount info special inode

rolysr:
This commit introduces the initial implementation of a special inode to display information about the mountpoint. This feature will help in accessing client option values and initial parameters for external applications use.


uristdwarf (approved):

src/mount/fuse/main.cc
```diff
@@ -175,6 +178,36 @@ int fuse_mnt_check_empty(const char *mnt, mode_t rootmode, off_t rootsize) {
...
+	return arguments;
+}
+
+std::string get_current_local_time() {
+	auto now = std::chrono::system_clock::now();
+	std::time_t now_time = std::chrono::system_clock::to_time_t(now);
+	std::tm *localtm = std::localtime(&now_time);
+
+	std::ostringstream oss;
+	oss << std::put_time(localtm, "%Y-%m-%d_%H:%M:%S");
```
dmga44: If possible, consider moving the new functions, headers and other code changes in main.cc to the hidden file source code. The idea would be to be to leave here only the required changes.

dmga44: The only change which could be left here is a call to a function in sauna_client.h that stores the original command line.

src/mount/fuse/main.cc
```diff
@@ -266,6 +299,10 @@ static int mainloop(struct fuse_args *args, struct fuse_cmdline_opts *fuse_opts,
 	params.statfs_cache_timeout = gMountOptions.statfscachetimeout;
 	params.use_quota_in_volume_size = gMountOptions.usequotainvolumesize;
 
+	// Update mount info actual pid
+	mountInfo.pid = getpid();
+	buildMountInfoStr();
+
 	if (!gMountOptions.meta) {
 		SaunaClient::fs_init(params);
 	} else {
```
dmga44: Consider building the parameters arguments for the hidden file in the call to fs_init().

src/mount/mount_info.h
```diff
@@ -0,0 +1,105 @@
...
+	std::string uid;
+	std::string gid;
+#endif
+	std::string username;
+	int pid;
+	std::string version;
+	std::string commitId;
+	std::string arguments;
+	std::unique_ptr<std::map<std::string, std::string>> mountOptions;
+} MountInfo;
```
dmga44: Please modernize the _mountInfo to a class.

src/mount/readdata.cc
```diff
@@ -618,6 +619,7 @@ void read_data_init(uint32_t retries,
 	gPrefetchXorStripes = prefetchXorStripes;
 	gBandwidthOveruse = bandwidth_overuse;
 	gTweaks.registerVariable("PrefetchXorStripes", gPrefetchXorStripes);
+	tweakByMountOption["sfsprefetchxorstripes"] = "PrefetchXorStripes";
```
dmga44: Please pass the original name of the variable as a param to the registerVariable function to simplify the implementation.

src/mount/readdata.cc
```diff
@@ -629,13 +631,21 @@ void read_data_init(uint32_t retries,
 	pthread_attr_destroy(&thattr);
 
 	gTweaks.registerVariable("ReadMaxRetries", maxRetries);
+	tweakByMountOption["sfsioretries"] = "ReadMaxRetries";
```
dmga44: Please notice you're registering twice the "sfsioretries" option in tweakByMountOption.

For the specific case of "ReadMaxRetries" and "WriteMaxRetries", please consider doing the following:
- register them, but instead of "sfsioretries", as "sfsreadretries" and "sfswriteretries".
- modify the mountInfo.mountOptions got while building the mount info string to search for the last two mentioned keys (insert instead of "sfsioretries", the new "sfsreadretries" and "sfswriteretries").

dmga44 (commented):
Great job, please see my comments.

src/mount/fuse/main.cc
rolysr: Done

src/mount/fuse/main.cc
rolysr: Done

src/mount/mount_info.h
rolysr: Done

src/mount/readdata.cc

src/mount/readdata.cc

dmga44 (approved):

PR #328, opened 2025-03-11
refactor(chunkserver): Split OutputBuffer inner buffer

dmga44:
The main goal of this change is to prepare the `OutputBuffer` to emplace
bigger pieces of data at once during pread calls. This could not be
achieved previously because the buffer should have a sequence of
header, CRC, data block, header, CRC and so on.

The class `OutputBuffer::Buffer` is created for this purpose. This is a
generic class to implement the three types of buffers an `OutputBuffer`
has. Taking this into consideration, the implementation of the
`OutputBuffer` could be made simpler and more readable.

Side changes:
- increase readability of putting data into the `OutputBuffer`.
- save the allocation of a block of zeros when asking for block beyond current number of blocks in chunk.
- modernize the `WriteStatus` enum.
- change `BuffersPool` to adapt to the new format of the `OutputBuffer`. Keep track of the number of block in the buffers instead of the number of buffers, and log it from time to time.
- add documentation for the `OutputBuffer` class.
- improve the unit tests suite of the `OutputBuffer` class.

The idea is to squash all these commits.

### DIFF
--- a/src/chunkserver/buffers_pool.h
+++ b/src/chunkserver/buffers_pool.h
@@ -51,6 +51,14 @@ public:
 	std::shared_ptr<T> get(size_t headerSize, size_t numBlocks) {
 		std::unique_lock lock(mutex_);
 
+		if (operationsSinceLastLog_ == kLogAfterEveryXTimes) {
+			safs::log_trace(
+			    "BuffersPool::get {} blocks, currentNumberOfBlocks_ = {}, kMaxNumberOfBlocks = {}",
+			    numBlocks, currentNumberOfBlocks_, kMaxNumberOfBlocks);
+			operationsSinceLastLog_ = 0;
+		}
+		operationsSinceLastLog_++;
+
 		auto it = buffersMap_.find({headerSize, numBlocks});
 		if (it == buffersMap_.end() || it->second.empty()) {
 			// To make sure the allocation is not done under the lock.
@@ -63,7 +71,7 @@ public:
 
 		buffers.pop();
 		buffer->clear();
-		currentSize_--;
+		currentNumberOfBlocks_ -= numBlocks;
 
 		return buffer;
 	}
@@ -74,21 +82,36 @@ public:
 	 */
 	void put(std::shared_ptr<T> &&buffer) {
 		std::unique_lock lock(mutex_);
-		auto &buffers = buffersMap_[buffer->type()];
-		if (currentSize_ < kMaxSize) {
+
+		auto [headerSize, numBlocks] = buffer->type();
+		auto &buffers = buffersMap_[{headerSize, numBlocks}];
+
+		if (operationsSinceLastLog_ == kLogAfterEveryXTimes) {
+			safs::log_trace(
+			    "BuffersPool::put {} blocks, currentNumberOfBlocks_ = {}, kMaxNumberOfBlocks = {}",
+			    numBlocks, currentNumberOfBlocks_, kMaxNumberOfBlocks);
+			operationsSinceLastLog_ = 0;
+		}
+		operationsSinceLastLog_++;
+
+		if (currentNumberOfBlocks_ + numBlocks <= kMaxNumberOfBlocks) {
 			buffer->clear();
 			buffers.push(std::move(buffer));
-			currentSize_++;
+			currentNumberOfBlocks_ += numBlocks;
 		}
 		// To make sure the deallocation is not done under the lock.
 		lock.unlock();
 	}
 
 private:
+	/// How many operations should pass between two log messages.
+	static constexpr size_t kLogAfterEveryXTimes = 1000;
+	/// Number of operations since last log message.
+	size_t operationsSinceLastLog_ = 0;
 	/// Maximum number of buffers in the pool.
-	static constexpr size_t kMaxSize = 8192;
+	static constexpr size_t kMaxNumberOfBlocks = 8192;
 	/// Current number of buffers in the pool.
-	size_t currentSize_ = 0;
+	size_t currentNumberOfBlocks_ = 0;
 	/// Buffers pool map: a queue of buffers for each type of buffer.
 	std::map<std::pair<size_t, size_t>, std::queue<std::shared_ptr<T>>> buffersMap_;
 	/// Mutex to protect the pool.
--- a/src/chunkserver/buffers_pool.h
+++ b/src/chunkserver/buffers_pool.h
@@ -44,26 +44,28 @@ public:
 
 	/**
 	 * Gets a buffer from the pool or creates a new one.
-	 * @param capacity Requested buffer capacity.
+	 * @param headerSize The size of the header.
+	 * @param numBlocks The number of blocks.
 	 * @return The existent buffer or a newly created one.
 	 */
-	std::shared_ptr<T> get(size_t capacity) {
-		std::lock_guard lock(mutex_);
+	std::shared_ptr<T> get(size_t headerSize, size_t numBlocks) {
+		std::unique_lock lock(mutex_);
 
-		if (buffers_.empty()) {
-			return std::make_shared<T>(capacity);
+		auto it = buffersMap_.find({headerSize, numBlocks});
+		if (it == buffersMap_.end() || it->second.empty()) {
+			// To make sure the allocation is not done under the lock.
+			lock.unlock();
+			return std::make_shared<T>(headerSize, numBlocks);
 		}
 
-		auto buffer = buffers_.front();
+		auto &buffers = it->second;
+		auto buffer = buffers.front();
 
-		if (buffer->capacity() == capacity) {
-			buffers_.pop();
-			buffer->clear();
+		buffers.pop();
+		buffer->clear();
+		currentSize_--;
 
-			return buffer;
-		}
-
-		return std::make_shared<T>(capacity);
+		return buffer;
 	}
 
 	/**
@@ -71,19 +73,24 @@ public:
 	 * @param buffer The buffer to put back.
 	 */
 	void put(std::shared_ptr<T> &&buffer) {
-		std::lock_guard lock(mutex_);
-
-		if (buffers_.size() < kMaxSize) {
+		std::unique_lock lock(mutex_);
+		auto &buffers = buffersMap_[buffer->type()];
+		if (currentSize_ < kMaxSize) {
 			buffer->clear();
-			buffers_.push(std::move(buffer));
+			buffers.push(std::move(buffer));
+			currentSize_++;
 		}
+		// To make sure the deallocation is not done under the lock.
+		lock.unlock();
 	}
 
 private:
 	/// Maximum number of buffers in the pool.
 	static constexpr size_t kMaxSize = 8192;
-	/// Buffers pool (container).
-	std::queue<std::shared_ptr<T>> buffers_;
+	/// Current number of buffers in the pool.
+	size_t currentSize_ = 0;
+	/// Buffers pool map: a queue of buffers for each type of buffer.
+	std::map<std::pair<size_t, size_t>, std::queue<std::shared_ptr<T>>> buffersMap_;
 	/// Mutex to protect the pool.
 	std::mutex mutex_;
 };
--- a/src/chunkserver/chunkserver_entry.cc
+++ b/src/chunkserver/chunkserver_entry.cc
@@ -41,22 +41,22 @@
 #include "chunkserver/hdd_readahead.h"
 #include "chunkserver/hddspacemgr.h"
 #include "chunkserver/network_stats.h"
+#include "chunkserver/output_buffer.h"
 #include "common/charts.h"
-#include "output_buffer.h"
+#include "common/datapack.h"
+#include "common/event_loop.h"
+#include "common/legacy_vector.h"
+#include "common/massert.h"
+#include "common/saunafs_version.h"
+#include "common/sockets.h"
+#include "devtools/TracePrinter.h"
+#include "devtools/request_log.h"
+#include "protocol/SFSCommunication.h"
 #include "protocol/cltocs.h"
 #include "protocol/cstocl.h"
 #include "protocol/cstocs.h"
-#include "common/datapack.h"
-#include "common/event_loop.h"
-#include "common/saunafs_version.h"
-#include "common/massert.h"
-#include "protocol/SFSCommunication.h"
-#include "common/legacy_vector.h"
 #include "protocol/packet.h"
 #include "slogger/slogger.h"
-#include "common/sockets.h"
-#include "devtools/request_log.h"
-#include "devtools/TracePrinter.h"
 
 constexpr uint32_t kMaxPacketSize = 100000 + SFSBLOCKSIZE;
 constexpr uint8_t kConnectRetries = 10;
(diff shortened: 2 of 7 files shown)


src/chunkserver/chunkserver_entry.cc
```diff
@@ -1582,11 +1622,11 @@ void ChunkserverEntry::writeToSocket() {
 			massert(bytesInBufferAfter <= bytesInBufferBefore,
 					"New bytes in pack->outputBuffer after sending some data");
 			stats_bytesout += (bytesInBufferBefore - bytesInBufferAfter);
-			if (ret == OutputBuffer::WRITE_ERROR) {
+			if (ret == OutputBuffer::WriteStatus::Error) {
```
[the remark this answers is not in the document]
dmga44: All sites using these write statuses were updated.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -1582,11 +1622,11 @@ void ChunkserverEntry::writeToSocket() {
...
 				safs_silent_errlog(LOG_NOTICE, "(write) write error");
 				state = State::Close;
 				return;
-			} else if (ret == OutputBuffer::WRITE_AGAIN) {
+			} else if (ret == OutputBuffer::WriteStatus::Again) {
```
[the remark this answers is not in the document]
dmga44: Same as before.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -137,19 +137,18 @@ MessageSerializer *MessageSerializer::getSerializer(PacketHeader::Type type) {
...
 
-	uint32_t sizeOfWholePacket = PacketHeader::kSize + header.length;
 	std::unique_ptr<PacketStruct> outPacket = std::make_unique<PacketStruct>();
 	passert(outPacket);
 
-	outPacket->outputBuffer = getReadOutputBufferPool().get(sizeOfWholePacket);
+	outPacket->outputBuffer = getReadOutputBufferPool().get(packetPrefix.size(), numBlocks);
 
-	if (outPacket->outputBuffer->copyIntoBuffer(packetPrefix) !=
+	if (outPacket->outputBuffer->copyIntoBuffer(OutputBuffer::BufferType::Header, packetPrefix) !=
```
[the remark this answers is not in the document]
dmga44: It is.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -361,49 +401,51 @@ void ChunkserverEntry::readContinue() {
...
+			const uint32_t thisPartSize =
+			    std::min<uint32_t>(totalRequestSize, SFSBLOCKSIZE - thisPartOffset);
+			const uint32_t numBlocks = (thisPartSize + SFSBLOCKSIZE - 1) >> SFSBLOCKBITS;
+			const uint16_t totalRequestBlocks =
+			    (totalRequestSize + thisPartOffset + SFSBLOCKSIZE - 1) / SFSBLOCKSIZE;
+
+			std::vector<uint8_t> readDataPrefix;
+			messageSerializer->serializePrefixOfCstoclReadData(readDataPrefix, chunkId, offset,
+			                                                   thisPartSize);
+			auto packet = createDetachedPacketWithOutputBuffer(readDataPrefix, numBlocks);
```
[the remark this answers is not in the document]
dmga44: It is the correct way of calculating it.

src/chunkserver/hddspacemgr.cc
```diff
@@ -744,24 +748,26 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
...
 				uint8_t *crcBuffPointer = crcBuff;
-				put32bit(&crcBuffPointer,
-				         mycrc32(0, tmp.data() + kCrcSize + offsetWithinBlock,
-				                 size));
-				outputBuffer->copyIntoBuffer(crcBuff, kCrcSize);
+				put32bit(
+				    &crcBuffPointer,
+				    mycrc32(0, tmp.rawData(OutputBuffer::BufferType::Block) + offsetWithinBlock,
+				            size));
+				outputBuffer->copyIntoBuffer(OutputBuffer::BufferType::CRC, crcBuff, kCrcSize);
```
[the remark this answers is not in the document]
dmga44: It is.

src/chunkserver/hddspacemgr.cc
```diff
@@ -744,24 +748,26 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
...
 				outputBuffer->copyIntoBuffer(
-				    tmp.data() + kCrcSize + offsetWithinBlock, size);
+				    OutputBuffer::BufferType::Block,
+				    tmp.rawData(OutputBuffer::BufferType::Block) + offsetWithinBlock, size);
```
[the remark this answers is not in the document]
dmga44: It is.

src/chunkserver/bgjobs.cc
```diff
@@ -277,6 +277,8 @@ void* job_worker(void *th_arg) {
 						rdargs->offset, rdargs->size, rdargs->maxBlocksToBeReadBehind,
 						rdargs->blocksToBeReadAhead, rdargs->outputBuffer);
 
+				rdargs->outputBuffer->isReadCompleted = true;
```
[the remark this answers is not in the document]
dmga44: It is the intended behavior, plus, this change belongs to another PR.

src/chunkserver/buffers_pool.h
```diff
@@ -47,43 +47,49 @@ class BuffersPool {
...
 	/**
 	 * Puts a buffer back to the pool.
 	 * @param buffer The buffer to put back.
 	 */
 	void put(std::shared_ptr<T> &&buffer) {
-		std::lock_guard lock(mutex_);
-
-		if (buffers_.size() < kMaxSize) {
+		std::unique_lock lock(mutex_);
+		auto &buffers = buffersMap_[buffer->type()];
```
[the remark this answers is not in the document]
dmga44: The `type()` means that the `headerSize` and `numBlocks` are correct.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -289,31 +288,78 @@ void ChunkserverEntry::delayedCloseCallback(uint8_t status, void *entry) {
...
 		eptr->isChunkOpen = 1;
 	}
 	if (eptr->isChunkOpen) {
 		job_close(eptr->workerJobPool, nullptr, nullptr, eptr->chunkId,
 		          eptr->chunkType);
 		eptr->isChunkOpen = 0;
 	}
-	eptr->state = State::Closed;
+	if (eptr->pendingReadJobIds.empty() && eptr->toDiscardReadJobIds.empty() && eptr->writeJobWriteId == 0) {
+		eptr->state = State::Closed;
```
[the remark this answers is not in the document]
dmga44: Belongs to another PR.

src/chunkserver/output_buffer.h
```diff
@@ -32,46 +32,123 @@
...
 	void clear();
 
-	static inline size_t getAlignedSize(size_t capacity) {
-		size_t remainder = capacity % disk::kIoBlockSize;
-
-		if (remainder == 0) { return capacity; }
-
-		return capacity + disk::kIoBlockSize - remainder;
-	}
+	std::atomic_bool isReadCompleted{false};
```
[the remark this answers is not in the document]
dmga44: Yes.

src/chunkserver/output_buffer_unittest.cc
```diff
@@ -42,17 +43,23 @@ TEST(OutputBufferTests, outputBuffersTest) {
...
 
 	uint8_t buf[WRITE_SIZE];
 	memset(buf, VALUE, WRITE_SIZE);
-	ASSERT_EQ(outputBuffer.copyIntoBuffer(buf, WRITE_SIZE), WRITE_SIZE);
+	ASSERT_EQ(outputBuffer.copyIntoBuffer(OutputBuffer::BufferType::Header,
+	                                      std::vector<uint8_t>(testHeaderSize, VALUE)),
+	          testHeaderSize);
+	ASSERT_EQ(outputBuffer.copyIntoBuffer(OutputBuffer::BufferType::CRC, buf, kCrcSize), kCrcSize);
+	ASSERT_EQ(outputBuffer.copyIntoBuffer(OutputBuffer::BufferType::Block, buf, WRITE_SIZE_DATA),
+	          WRITE_SIZE_DATA);
```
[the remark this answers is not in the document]
dmga44: Indeed.

src/chunkserver/hddspacemgr.cc
```diff
@@ -656,7 +659,7 @@ static void hddReadAheadAndBehind(IChunk *chunk, uint16_t block,
 		    *chunk, firstBlockToRead,
 		    blocksToBeReadAhead + block - firstBlockToRead);
 		OutputBuffer buffer =
-		    OutputBuffer(kHddBlockSize * (block - firstBlockToRead));
+		    OutputBuffer(1, block - firstBlockToRead);
```
lgsilva3087: Consider to use a constant like kIgnoreHeaderSize = 1.

src/chunkserver/buffers_pool.h
```diff
@@ -44,46 +44,53 @@ class BuffersPool {
...
+			buffers.push(std::move(buffer));
+			currentSize_++;
 		}
+		// To make sure the deallocation is not done under the lock.
+		lock.unlock();
 	}
 
 private:
 	/// Maximum number of buffers in the pool.
 	static constexpr size_t kMaxSize = 8192;
```
lgsilva3087: Consider to use kMaxNumberOfBlocks instead of kMaxSize, for better control of the total memory size to be used by the buffer in the next steps (bigger size per write request).

Consider to log_trace the current size every defined amount of get or put operations (example every 1000 operations), to improve observability on memory usage.

lgsilva3087 (commented):
Looks good, please see my small suggestions.

src/chunkserver/hddspacemgr.cc
```diff
@@ -744,24 +748,26 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
 		status = hddReadCrcAndBlock(chunk, block, outputBuffer);
 
 		if (status == SAUNAFS_STATUS_OK) {
-			status = hddCheckCrcForFullBlock(chunk, block, outputBuffer, false);
+			status = hddCheckCrcForFullBlock(chunk, block, outputBuffer, 0, false);
```
rolysr, line 751: You can consider in these cases, if the last params to be passed are 0 and false, then you can change the `hddCheckCrcForFullBlock` for having them as default on the signature definition.

src/chunkserver/hddspacemgr.cc
```diff
@@ -744,24 +748,26 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
...
 		}
 	} else {  // Partial block
-		OutputBuffer tmp(kHddBlockSize);
+		OutputBuffer tmp(1, 1);
```
rolysr: Consider to comment these magic `1` values here.

src/chunkserver/hddspacemgr.cc
```diff
@@ -744,24 +748,26 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
...
 		status = hddReadCrcAndBlock(chunk, block, &tmp);
 
 		if (status == SAUNAFS_STATUS_OK) {  // Successful read of the full block
-			status = hddCheckCrcForFullBlock(chunk, block, &tmp, true);
+			status = hddCheckCrcForFullBlock(chunk, block, &tmp, 0, true);
```
rolysr, line 759: Same consideration here as in a previous comment.

src/chunkserver/output_buffer.cc
```diff
@@ -30,71 +30,119 @@
...
+		}
+		ssize_t ret = ::write(outputFileDescriptor, blockBuffer_.getUnflushedDataFirstIndex(),
+		                      currentRemainingBytesForFD_);
+
 		if (ret <= 0) {
-			if (ret == 0 || errno == EAGAIN) {
-				return WRITE_AGAIN;
-			}
-			return WRITE_ERROR;
+			if (ret == 0 || errno == EAGAIN) { return WriteStatus::Again; }
```
rolysr, line 62: Consider this change here:
```suggestion
			if (!ret || errno == EAGAIN) { return WriteStatus::Again; }
```

src/chunkserver/output_buffer.cc
```diff
@@ -30,71 +30,119 @@
...
+ssize_t OutputBuffer::copyIntoBuffer(BufferType type, const void *mem, size_t len) {
+	switch (type) {
+	case BufferType::Block:
+		return blockBuffer_.copyIntoBuffer(mem, len);
+	case BufferType::CRC:
+		return crcBuffer_.copyIntoBuffer(mem, len);
+	case BufferType::Header:
+		return headerBuffer_.copyIntoBuffer(mem, len);
+	default:
+		safs::log_err("Invalid buffer type");
```
rolysr: Consider to improve the logging here by specifying that this comes from the OutputBuffer

rolysr (approved):
LGTM.

src/chunkserver/output_buffer.cc
uristdwarf: I'd rather be explicit and check for 0 explicitly, even though negatives convert to true. Seems more clear to me.

I don't know if it's logged in an above function, but even then I think you should log errno here (as an error), to better trace it later if needed.

uristdwarf: Of course, if it's logged above, you probably need to remove or change the errno checking there, from man 3 errno:

```
      A common mistake is to do

           if (somecall() == -1) {
               printf("somecall() failed\n");
               if (errno == ...) { ... }
           }

       where errno no longer needs to have the value it had  upon  return  from
       somecall()  (i.e.,  it  may have been changed by the printf(3)).  If the
       value of errno should be preserved across a library  call,  it  must  be
       saved:

           if (somecall() == -1) {
               int errsv = errno;
               printf("somecall() failed\n");
               if (errsv == ...) { ... }
           }
```

src/chunkserver/output_buffer.cc
```diff
@@ -30,71 +30,119 @@
...
+	switch (type) {
+	case BufferType::Block:
+		return blockBuffer_.copyIntoBuffer(mem.data(), mem.size());
+	case BufferType::CRC:
+		return crcBuffer_.copyIntoBuffer(mem.data(), mem.size());
+	case BufferType::Header:
+		return headerBuffer_.copyIntoBuffer(mem.data(), mem.size());
+	default:
+		safs::log_err("Invalid buffer type");
+		return 0;
```
uristdwarf: I wonder if an assert is more appropriate. In what situation would it be normal to have an invalid buffer type passed?

uristdwarf (commented):
Review part 1/2?

src/chunkserver/hddspacemgr.cc
dmga44: IMO I think it won't help too much, because this would be the only affeected case. Further changes would also add up calls to this function with starting offset different than 0, so I think it would be better to keep it explicit.

src/chunkserver/hddspacemgr.cc
dmga44: The first parameter of the constructor was changed to `kIgnoreHeaderSize` to make it more readable, and added a comment on top to that line. The second `1` is because it is necessary to read a block.

src/chunkserver/hddspacemgr.cc
dmga44: Note the first answer holds.

src/chunkserver/output_buffer.cc
dmga44: I agree with Urmas, I think it is better the explicit equal check with 0.

src/chunkserver/output_buffer.cc
dmga44: Done, I also changed the log type to warn because it is not critical, it is just ignored operation.

src/chunkserver/output_buffer.cc
dmga44: I actually changed the log type to warn because it is not so critical, it is just an ignored operation.

Regarding your other question, it is almost impossible because those three types are the only defined ones. Currently all the calls to these functions only use those three buffers types.

src/chunkserver/output_buffer.cc
dmga44: Regarding @uristdwarf suggestion of logging errno here: the function which calls this `writeOutToAFileDescriptor` function logs an error using old LOG_NOTICE but does not specify. I could log it here, but then I'll have two logs about the same thing, and modifying it in an above function could already not have the initial errno.

So, I'll opt for leaving it like it is now.

src/chunkserver/hddspacemgr.cc
dmga44: Done.

src/chunkserver/buffers_pool.h
dmga44: Done, both.

uristdwarf (approved):

lgsilva3087 (approved):
LGTM

PR #332, opened 2025-03-19
feat: Add initial scripts for FoundationDB

lgsilva3087:
This pull request introduces scripts to automate the installation and
management of FoundationDB for testing purposes. The changes are
mostly:

- Add a script to install FoundationDB.
- Add script to provide functions to start and stop on-demand
  FoundationDB clusters for testing.
- Integrate into the testing framework.
- Create an app in C++ to test the connectivity to the on-demand FDB
  cluster.
- Create the FDBTests suite with one basic test.

### DIFF
--- a/tests/ci_build/install-foundationdb.sh
+++ b/tests/ci_build/install-foundationdb.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+# This script installs FoundationDB on a Debian-based system.
+# It downloads the client and server packages from the official GitHub repository,
+# installs them, and verifies the installation.
+
+set -eu -o pipefail
+
+readonly workspace="/tmp/saunafs-foundationdb-tmp"
+
+function die() {
+	echo "Error: ${*}" >&2
+
+	# Check if the workspace exists and remove it
+	if [ -d "${workspace}" ]; then
+		echo "Removing temporary workspace..."
+		rm -rf "${workspace:?}"
+	fi
+
+	exit 1
+}
+
+# Ensure the script is run as root
+if [ "$(id -u)" -ne 0 ]; then
+	die "This script must be run as root. Please use sudo."
+fi
+
+fdb_version="7.3.63"
+
+# Check if FoundationDB is already installed by using fdbcli
+if command -v fdbcli &> /dev/null; then
+	echo "FoundationDB is already installed. Version:"
+	fdbcli --version
+	exit 0
+fi
+
+# Check if wget is installed
+if ! command -v wget &> /dev/null; then
+	die "wget is not installed. Please install wget first."
+fi
+
+# Create the temporary workspace
+mkdir -p "${workspace}"
+
+# Define some useful variables
+client_pkg="foundationdb-clients_${fdb_version}-1_amd64.deb"
+server_pkg="foundationdb-server_${fdb_version}-1_amd64.deb"
+hash_suffix=".sha256"
+base_url="https://github.com/apple/foundationdb/releases/download/${fdb_version}"
+
+# Download the client and server packages
+echo "Downloading FoundationDB packages..."
+wget "${base_url}/${client_pkg}" -O "${workspace}/${client_pkg}" || die "Failed to download client package"
+wget "${base_url}/${server_pkg}" -O "${workspace}/${server_pkg}" || die "Failed to download server package"
+
+# Download the SHA256 hash file
+wget "${base_url}/${client_pkg}${hash_suffix}" -O "${workspace}/${client_pkg}${hash_suffix}" || die "Failed to download client hash file"
+wget "${base_url}/${server_pkg}${hash_suffix}" -O "${workspace}/${server_pkg}${hash_suffix}" || die "Failed to download server hash file"
+
+# Verify the downloaded packages
+echo "Verifying downloaded packages..."
+cd "${workspace}" || die "Failed to change directory to ${workspace}"
+sha256sum -c "${client_pkg}${hash_suffix}" || die "Client package verification failed"
+sha256sum -c "${server_pkg}${hash_suffix}" || die "Server package verification failed"
+cd - || die "Failed to change back to previous directory"
+
+# Install the packages
+echo "Installing FoundationDB packages..."
+dpkg -i "${workspace}/${client_pkg}" "${workspace}/${server_pkg}" || die "Failed to install FoundationDB packages"
+
+# Fix any missing dependencies
+echo "Fixing missing dependencies..."
+apt-get install -f -y || die "Failed to fix missing dependencies"
+
+# Verify the installation
+echo "Verifying FoundationDB installation..."
+if command -v fdbcli &> /dev/null; then
+	echo "FoundationDB CLI installed successfully."
+	fdbcli --version
+else
+	echo "FoundationDB CLI installation failed."
+	exit 1
+fi
+
+# Stop FoundationDB service if running after the installation
+systemctl stop foundationdb || true
+
+# Clean up temporary workspace
+echo "Removing temporary workspace..."
+rm -r "${workspace:?}"
+
+echo "FoundationDB installation completed successfully."
--- a/tests/docker/Dockerfile.test
+++ b/tests/docker/Dockerfile.test
@@ -17,6 +17,7 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
 	apt-get update && apt-get install tini && \
     ./tests/install-packages.sh;
 
+COPY ./tests/ci_build/install-foundationdb.sh ./tests/ci_build/
 COPY ./tests/setup_machine.sh ./tests/
 
 # Run the setup-test-machine.sh script
--- a/tests/setup_machine.sh
+++ b/tests/setup_machine.sh
@@ -194,6 +194,9 @@ if [ ! -f /etc/sudoers.d/saunafstest ] || ! grep -q '# FoundationDB' /etc/sudoer
 	END
 fi
 
+echo ; echo 'Install FoundationDB'
+"$script_dir/ci_build/install-foundationdb.sh"
+
 echo ; echo 'Fixing GIDs of users'
 for name in saunafstest saunafstest_{0..9}; do
 	uid=$(getent passwd "${name}" | cut -d: -f3)
--- a/tests/setup_machine.sh
+++ b/tests/setup_machine.sh
@@ -186,6 +186,14 @@ if [ ! -f /etc/sudoers.d/saunafstest ] || ! grep -q '# Client' /etc/sudoers.d/sa
 	END
 fi
 
+if [ ! -f /etc/sudoers.d/saunafstest ] || ! grep -q '# FoundationDB' /etc/sudoers.d/saunafstest > /dev/null; then
+	cat <<-'END' >>/etc/sudoers.d/saunafstest
+		# FoundationDB
+		saunafstest ALL = NOPASSWD: /usr/lib/foundationdb/fdbmonitor --conffile*
+		saunafstest ALL = NOPASSWD: /usr/bin/pkill -f fdbmonitor*
+	END
+fi
+
 echo ; echo 'Fixing GIDs of users'
 for name in saunafstest saunafstest_{0..9}; do
 	uid=$(getent passwd "${name}" | cut -d: -f3)
--- a/tests/tools/foundationdb.sh
+++ b/tests/tools/foundationdb.sh
@@ -1,15 +1,6 @@
 #!/usr/bin/env bash
 
-# Starts a temporary FoundationDB Cluster in /tmp/saunafs-fdb-test.
-# The cluster uses two servers on ports 4500 and 4501.
-# This script is intended for testing purposes only.
-
-set -eu -o pipefail
-
-readonly script_name=$(basename "$0")
 readonly workspace="/tmp/saunafs-fdb-test"
-
-readonly action="${1:-}"
 
 function create_workspace() {
 	mkdir -p "${workspace}"/{conf,data,logs}
@@ -84,20 +75,3 @@ function start_fdb_cluster() {
 function stop_fdb_cluster() {
 	cleanup_fdb_cluster
 }
-
-function main() {
-  case "${action}" in
-    start)
-      start_fdb_cluster
-      ;;
-    stop)
-      stop_fdb_cluster
-      ;;
-    *)
-      echo "Usage: ${script_name} {start|stop}"
-      exit 1
-      ;;
-  esac
-}
-
-main "$@"
--- a/tests/tools/foundationdb.sh
+++ b/tests/tools/foundationdb.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+# Starts a temporary FoundationDB Cluster in /tmp/saunafs-fdb-test.
+# The cluster uses two servers on ports 4500 and 4501.
+# This script is intended for testing purposes only.
+
+set -eu -o pipefail
+
+readonly script_name=$(basename "$0")
+readonly workspace="/tmp/saunafs-fdb-test"
+
+readonly action="${1:-}"
+
+function create_workspace() {
+	mkdir -p "${workspace}"/{conf,data,logs}
+	chown -R "$(id -un):$(id -gn)" "${workspace}"
+}
+
+function create_config_file() {
+	cat > "${workspace}/conf/foundationdb.conf" <<EOF
+[fdbmonitor]
+user = $(id -un)
+group = $(id -gn)
+lockfile = ${workspace}/fdbmonitor.pid
+
+[general]
+restart-delay = 60
+cluster-file = ${workspace}/conf/fdb.cluster
+
+[fdbserver]
+command = /usr/sbin/fdbserver
+public_address = 127.0.0.1:\$ID
+listen_address = public
+datadir = ${workspace}/data/\$ID
+logdir = ${workspace}/logs
+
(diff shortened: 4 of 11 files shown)


tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,101 @@
...
+
+function start_fdb_cluster() {
+  cleanup_fdb_cluster
+  create_workspace
+
+  # Create cluster file manually
+  if [ ! -f "${workspace}/conf/fdb.cluster" ]; then
+      echo "Creating new cluster configuration..."
+      echo "saunafstest:clusterId@127.0.0.1:4500" > "${workspace}/conf/fdb.cluster"
+      chmod 644 "${workspace}/conf/fdb.cluster"
```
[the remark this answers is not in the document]
lgsilva3087: Can you provide an example? Your proposed changes are not related to your comment.

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,101 @@
...
+  create_config_file
+
+  # Start FoundationDB with required sudo privileges
+  sudo /usr/lib/foundationdb/fdbmonitor --conffile "${workspace}/conf/foundationdb.conf" &
+
+  # Wait for fdbmonitor to initialize
+  sleep 3
+
+  # Ensure cluster is configured
+  fdbcli --exec "configure new single memory" --cluster-file "${workspace}/conf/fdb.cluster" || true
```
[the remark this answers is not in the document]
lgsilva3087: Done.

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,101 @@
...
+EOF
+}
+
+function check_cluster_status() {
+  fdbcli --exec "status" --cluster-file "${workspace}/conf/fdb.cluster"
+}
+
+function cleanup_fdb_cluster() {
+    sudo pkill -f fdbmonitor || true
+    rm -r "${workspace}" || true
```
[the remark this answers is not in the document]
lgsilva3087: OK, I got the idea, but I would like to keep it simple for now.

tests/tools/foundationdb.sh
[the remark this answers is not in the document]
lgsilva3087: I agree. I am still looking for the way to do it as you suggest but the used approaches fail.

tests/ci_build/install-foundationdb.sh
```diff
@@ -0,0 +1,60 @@
...
+
+fdb_version="7.3.63"
+client_pkg="foundationdb-clients_${fdb_version}-1_amd64.deb"
+server_pkg="foundationdb-server_${fdb_version}-1_amd64.deb"
+base_url="https://github.com/apple/foundationdb/releases/download/${fdb_version}"
+
+# Download the client and server packages
+echo "Downloading FoundationDB packages..."
+wget "${base_url}/${client_pkg}" -O "${workspace}/${client_pkg}"
+wget "${base_url}/${server_pkg}" -O "${workspace}/${server_pkg}"
```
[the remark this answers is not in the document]
lgsilva3087: Done.

tests/ci_build/install-foundationdb.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Install the packages
+echo "Installing FoundationDB packages..."
+dpkg -i "${workspace}/${client_pkg}" "${workspace}/${server_pkg}"
```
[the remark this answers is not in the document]
lgsilva3087: Done.

tests/ci_build/install-foundationdb.sh
```diff
@@ -0,0 +1,63 @@
...
+
+fdb_version="7.3.63"
+client_pkg="foundationdb-clients_${fdb_version}-1_amd64.deb"
+server_pkg="foundationdb-server_${fdb_version}-1_amd64.deb"
+base_url="https://github.com/apple/foundationdb/releases/download/${fdb_version}"
+
+# Download the client and server packages
+echo "Downloading FoundationDB packages..."
+wget "${base_url}/${client_pkg}" -O "${workspace}/${client_pkg}" || die "Failed to download client package"
+wget "${base_url}/${server_pkg}" -O "${workspace}/${server_pkg}" || die "Failed to download server package"
```
uristdwarf, line 54: You should verify the checksums and/or signatures for these deb files

tests/setup_machine.sh
```diff
@@ -266,5 +274,8 @@ done
 # shellcheck disable=SC2016
 echo ': ${SAUNAFS_LOOP_DISKS:="'"${loops[*]}"'"}' >> /etc/saunafs_tests.conf
 
+echo ; echo 'Install FoundationDB'
+"$script_dir/ci_build/install-foundationdb.sh"
+
```
uristdwarf: If you have it in the Dockerfile.test running before setup_machine.sh, it will run this script twice. Are you sure you want that (are there any side-effects from running it twice)?

tests/ci_build/install-foundationdb.sh
lgsilva3087: Sure, done. Thanks.

tests/setup_machine.sh
lgsilva3087: Hi, thanks for noticing it, no side effects but anyway I modified the script to abort if it was already installed. I keep the call in setup_machine for the developers workstations, I use it.

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,77 @@
...
```
wg1k: Please, add a safety measure on `rm` calls
```suggestion
    rm -r "${workspace:?}" || true
```

tests/ci_build/install-foundationdb.sh
```diff
@@ -0,0 +1,84 @@
...
+  echo "FoundationDB CLI installation failed."
+  exit 1
+fi
+
+# Stop FoundationDB service if running after the installation
+systemctl stop foundationdb || true
+
+# Clean up temporary workspace
+echo "Removing temporary workspace..."
+rm -r "${workspace}"
```
wg1k:
```suggestion
rm -r "${workspace:?}"
```

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+readonly workspace="/tmp/saunafs-fdb-test"
+
+function create_workspace() {
+    mkdir -p "${workspace}"/{conf,data,logs}
+    chown -R $(whoami):$(whoami) "${workspace}"
```
wg1k:
```suggestion
    chown -R "$(id -u):$(id -g)" "${workspace}"
```

uristdwarf (approved):

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function create_config_file() {
+  cat > "${workspace}/conf/foundationdb.conf" <<EOF
+[fdbmonitor]
+user = $(whoami)
+group = $(whoami)
```
wg1k:
```suggestion
user = $(id -u)
group = $(id -g)
```

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,77 @@
...
+lockfile = ${workspace}/fdbmonitor.pid
+
+[general]
+restart-delay = 60
+cluster-file = ${workspace}/conf/fdb.cluster
+
+[fdbserver]
+command = /usr/sbin/fdbserver
+public_address = 127.0.0.1:\$ID
```
wg1k, line 23: Is it expected to be accessed only from localhost?  In other case consider to bind it to an external IP, or `0.0.0.0`.

lgsilva3087: For the tests maybe localhost is enough.

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+readonly workspace="/tmp/saunafs-fdb-test"
+
+function create_workspace() {
```
wg1k, line 5: The `function` keyword is not required, as the `()` are doing the job. It is matter of preference while we are not concerned about portability, so far.

tests/ci_build/install-foundationdb.sh
lgsilva3087: Done.

tests/tools/foundationdb.sh
lgsilva3087: For some reason FoundationDB was not starting, I needed to use:
`$(id -un) and $(id -gn)`

tests/tools/foundationdb.sh
lgsilva3087: For some reason FoundationDB was not starting, I needed to use:
`$(id -un) and $(id -gn)`

tests/tools/foundationdb.sh
```diff
@@ -0,0 +1,77 @@
...
+[backup_agent]
+command = /usr/lib/foundationdb/backup_agent/backup_agent
+logdir = ${workspace}/logs
+
+[backup_agent.1]
+EOF
+}
+
+function check_cluster_status() {
+  fdbcli --exec "status" --cluster-file "${workspace}/conf/fdb.cluster"
```
wg1k: Could you please, check the indentation format?

lgsilva3087: Thanks, I changed both scripts to use tabs, should be consistent now.

tests/tools/foundationdb.sh
lgsilva3087: Done

utils/sfs_test_fdb.cc
wg1k, line 1: Might this file be moved to test focused place in the project structure, e.g., the `tests/` folder?

tests/docker/Dockerfile.test
```diff
@@ -17,6 +17,13 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
 	apt-get update && apt-get install tini && \
     ./tests/install-packages.sh;
 
+COPY ./tests/ci_build/install-foundationdb.sh ./tests/
+
+RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
+    --mount=target=/var/cache/apt,type=cache,sharing=locked \
+    rm -f /etc/apt/apt.conf.d/docker-clean && \
+    ./tests/install-foundationdb.sh || true;
+
```
wg1k: Consider this general rule: the more change-prone a layer is, the later it should be introduced in the Dockerfile, for better cache utilization.

wg1k: Given that the setup_machine is calling the script, this RUN seems not to add a value. Consider removing it, or otherwise taking the call out of the setup_machine and moving this RUN below, if you consider it will have frequent changes.

tests/ci_build/install-foundationdb.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
+
+# This script installs FoundationDB on a Debian-based system.
+# It downloads the client and server packages from the official GitHub repository,
+# installs them, and verifies the installation.
+
+set -eu -o pipefail
+
+function die() {
+	echo "${1}"
```
wg1k:
```suggestion
	echo "Error: ${*}" >&2
```

tests/ci_build/install-foundationdb.sh
```diff
@@ -0,0 +1,84 @@
...
+	exit 1
+}
+
+# Ensure the script is run as root
+if [ "$(id -u)" -ne 0 ]; then
+	die "This script must be run as root. Please use sudo."
+fi
+
+fdb_version="7.3.63"
```
rolysr, line 28: Will we be interested on keeping this `fdb_version` to this specific value or do we have intention of keeping our system always compatible with latest versions? In case of the second option, consider if it would be better to automatically fetch latest version.

tests/setup_machine.sh
```diff
@@ -186,6 +186,14 @@ if [ ! -f /etc/sudoers.d/saunafstest ] || ! grep -q '# Client' /etc/sudoers.d/sa
 	END
 fi
 
+if [ ! -f /etc/sudoers.d/saunafstest ] || ! grep -q '# FoundationDB' /etc/sudoers.d/saunafstest >/dev/null; then
```
rolysr:
```suggestion
if [ ! -f /etc/sudoers.d/saunafstest ] || ! grep -q '# FoundationDB' /etc/sudoers.d/saunafstest > /dev/null; then
```
Consider adding a space right after the latest `>`

utils/sfs_test_fdb.cc
```diff
@@ -0,0 +1,142 @@
...
+void runNetworkThread() { checkError(fdb_run_network()); }
+
+void setValue(FDBDatabase *database, size_t keySize, const uint8_t *key, size_t valueSize,
+              const uint8_t *value) {
+	FDBTransaction *transaction{};
+	checkError(fdb_database_create_transaction(database, &transaction));
+	fdb_transaction_set(transaction, key, static_cast<int>(keySize), value,
+	                    static_cast<int>(valueSize));
+	FDBFuture *commitFuture = fdb_transaction_commit(transaction);
+	checkError(fdb_future_block_until_ready(commitFuture));
```
rolysr, line 61: If an error occurs here during the `checkError` execution, will it be granted that `transaction` and `commitFuture` will correctly be freed?

utils/sfs_test_fdb.cc
```diff
@@ -0,0 +1,142 @@
...
+	checkError(fdb_database_create_transaction(database, &transaction));
+	static constexpr fdb_bool_t kUseSnapshot{0};
+	FDBFuture *getValueFuture =
+	    fdb_transaction_get(transaction, key, static_cast<int>(keySize), kUseSnapshot);
+	checkError(fdb_future_block_until_ready(getValueFuture));
+
+	fdb_bool_t valuePresent{};
+	const uint8_t *valueRead{};
+	int valueLength{};
+	checkError(fdb_future_get_value(getValueFuture, &valuePresent, &valueRead, &valueLength));
```
rolysr, line 77: Consider checking the same if applicable as suggested on the last comment.

rolysr (approved):
LGTM.

tests/ci_build/install-foundationdb.sh
lgsilva3087: Hi, the version is usually related to the API version used internally by the clients (`#define FDB_API_VERSION 730`). This one (7.3.63) is the latest stable. At least at the beginning I propose to use a fixed version, then upgrade and test with new versions and so on.

tests/setup_machine.sh
lgsilva3087: I applied it, but please notice that the rest of the script does not use this space.

utils/sfs_test_fdb.cc
lgsilva3087: If there is an error, `exit` is called, the OS will claim the memory in this specific example.

utils/sfs_test_fdb.cc
lgsilva3087: Same here. We can do it in this simple connection test (`exit`), but in the library we will need to handle it.

wg1k (approved):

tests/ci_build/install-foundationdb.sh
dmga44: As a small comment, you could also remove the workspace folder (and its content), if exists, in the die function like in the end of the script.

utils/sfs_test_fdb.cc
dmga44: I think this place is the correct one, just like the file-generate and file-validate commands, for instance.

utils/sfs_test_fdb.cc
```diff
@@ -0,0 +1,142 @@
...
+
+	if (argc == 2) { clusterFile = std::string(argv[1]); }
+
+	//  set up network
+	checkError(fdb_select_api_version(FDB_API_VERSION));
+	checkError(fdb_setup_network());
+
+	// run network thread
+	pthread_t networkThread = 0;
+	pthread_create(&networkThread, nullptr, (void *(*)(void *))runNetworkThread, nullptr);
```
dmga44: Why using here a `pthread_t` instead of a more modern `std::thread` or` std::jthread`?

tests/docker/Dockerfile.test
lgsilva3087: Fixed, now there is only one call in the setup_machine. The call in the setup_machine was unreachable before, because it is failing in the ramdisk section and the call was later.

lgsilva3087: Fixed, there is only one call now (setup_machine.sh).

utils/sfs_test_fdb.cc
lgsilva3087: Sure, changed to use std::jthread.

tests/ci_build/install-foundationdb.sh
lgsilva3087: Both applied. Thanks.

dmga44 (approved):

utils/sfs_test_fdb.cc
wg1k: I see that `utils` folder is being filled recently with test related stuff. Is that what we want? Both `tests/test_utils` or `tests/tools` seem to be  better suited. Am I wrong?

lgsilva3087: I partially agree with @ictus4u, but looks like utils was always used for the C and C++ tools used from the tests, since more than 11 years ago. For now, I will keep it there, and let's decide what to do with all those tools.

PR #333, opened 2025-03-20
feat(chunkserver): Batch CS read jobs

dmga44:
This PR targets these two features:
- make a single read job in a chunk able to read several blocks.
- make the read of the data in a single pread call for non-zoned drives.

The final goal is improving the read performance from the chunkserver side.
As a side change, a new option is added to the chunkserver list of options and a test is updated.

The idea is to squash all the commits into one.


src/protocol/cstocl.h
```diff
@@ -31,7 +31,9 @@ namespace readData {
 inline void serializePrefix(std::vector<uint8_t>& destination,
 		uint64_t chunkId, uint32_t readOffset, uint32_t readSize) {
 	// This prefix requires CRC (uint32_t) and data (readSize * uint8_t) to be appended
-	uint32_t extraSpace = serializedSize(uint32_t()) + readSize;
+	uint32_t extraSpace = (((readSize + SFSBLOCKSIZE - 1) >> SFSBLOCKBITS) *
+	                       serializedSize(uint32_t())) +
```
lgsilva3087: consider:
sizeof(uint32_t)

Which is a constexpr always evaluated at compile time.

src/chunkserver/hddspacemgr.cc
```diff
@@ -100,6 +100,9 @@
 constexpr int kErrorLimit = 2;
 constexpr int kLastErrorTime = 60;
 constexpr size_t kIgnoreHeaderSize = 1;
+constexpr uint32_t startingOffsetOfBlock(uint32_t blocknum) {
+	return blocknum * SFSBLOCKSIZE;
+}
```
lgsilva3087: Please apply the code style here, the line fits in the define number of columns.

src/chunkserver/chunkserver_entry.h
```diff
@@ -157,6 +157,7 @@ struct ChunkserverEntry {
 	std::set<uint32_t> partiallyCompletedWrites;
 
 	/* read */
+	uint16_t maxBlocksPerHddReadJob;
```
lgsilva3087: Consider:
/// Number of contiguous blocks to read from the device in one read job.

lgsilva3087 (approved):
LGTM

src/chunkserver/hddspacemgr.cc
```diff
@@ -737,23 +738,68 @@ int hddRead(uint64_t chunkId, uint32_t version, ChunkPartType chunkType,
...
+				uint32_t bytesRead = outputBuffer->copyIntoBuffer(OutputBuffer::BufferType::Block,
+				                                                  chunk, bytesToBeRead, offset);
+
+				if (bytesRead != bytesToBeRead) {
+					hddAddErrorAndPreserveErrno(chunk);
+					safs::log_warn("hddRead: file:{} - read error from block {} to {}",
+					               chunk->fullDataFilename().c_str(), initialBlock,
+					               initialBlock + numBlocks - 1);
+					hddReportDamagedChunk(chunk->id(), chunk->type());
+					return SAUNAFS_ERROR_IO;
```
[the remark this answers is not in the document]
dmga44: `hddReportDamagedChunk` is actually being called.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -404,6 +391,37 @@ void ChunkserverEntry::readFinishedCallback(uint8_t status, void *entry) {
...
+	    readDataPrefix, chunkId, offset, std::min<uint32_t>(jobSize, SFSBLOCKSIZE - jobOffset));
+
+	auto packet = createDetachedPacketWithOutputBuffer(readDataPrefix, numBlocks);
+
+	for (uint32_t i = 0; i < numBlocks; i++) {
+		if (i > 0) {  // first block is already serialized
+			readDataPrefix.clear();
+			messageSerializer->serializePrefixOfCstoclReadData(
+			    readDataPrefix, chunkId, offset - jobOffset + (i * SFSBLOCKSIZE), SFSBLOCKSIZE);
+		}
```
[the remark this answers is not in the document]
dmga44: The `readDataPrefix` needs to be clear before every serialization, that's what I'm ensuring here.

The calculations should not overflow because the initialization of those values make it so.

The inner csentry structures won't change during this call because only one thread access it.

src/protocol/cstocl.h
dmga44: Applied.

src/chunkserver/hddspacemgr.cc

src/chunkserver/chunkserver_entry.h
dmga44: Applied.

lgsilva3087 (approved):
Great. LGTM.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -136,28 +136,15 @@ MessageSerializer *MessageSerializer::getSerializer(PacketHeader::Type type) {
 	return &singleton;
 }
 
-std::unique_ptr<PacketStruct>
-ChunkserverEntry::createDetachedPacketWithOutputBuffer(
+std::unique_ptr<PacketStruct> ChunkserverEntry::createDetachedPacketWithOutputBuffer(
```
rolysr, line 139: Is it needed to keep these indentation changes?

src/chunkserver/network_main_thread.cc
```diff
@@ -120,6 +120,8 @@ void mainNetworkThreadReload(void) {
 	}
 	chunkReplicatorReload();
 
+	gMaxBlocksPerHddReadJob = cfg_get_minvalue<uint16_t>(
+	    "MAX_BLOCKS_PER_HDD_READ_JOB", NetworkWorkerThread::kDefaultMaxBlocksPerHddReadJob, 1);
```
rolysr, line 124: What would happen if a `0` value is passed through the configuration option `MAX_BLOCKS_PER_HDD_READ_JOB`?

rolysr (approved):
LGTM. Take a look at my suggestions.

src/chunkserver/chunkserver_entry.cc
dmga44: Well, I wanted to change the signature of the method and could apply the new width rule that we have, so I did it. I think it looks better this way.

src/chunkserver/network_main_thread.cc
dmga44: The third parameter, which is 1, is the minimum value this option could get. So, if a 0 is passed, then the value used is 1, which current behavior in dev.

PR #337, opened 2025-03-27
fix: Fix SQLite issue

dmga44:
An issue appeared when several processes/threads started doing
concurrent operations on a SQLite database in our system. The issue originated from
having different ReadRecords for each those processes while also doing
writes on the same file the cache was about. Solution is to clear the
cache after chunk is locked and prepared to be written.

A couple of tests were added to test the fix.

Co-authored-by: aNeutrino <pm@leil.io>

### DIFF
--- a/src/mount/readdata.cc
+++ b/src/mount/readdata.cc
@@ -672,8 +672,8 @@ void read_inode_ops(uint32_t inode) { // attributes of inode have been changed -
 	}
 }
 
-// Ongoing write on this inode and chunkId, force reconnect and clear cache for this chunk
-void read_inode_reconnect_clear(uint32_t inode, uint32_t chunkId) {
+// Ongoing write on this inode and chunkIndex, force reconnect and clear cache for this chunk
+void read_inode_reconnect_and_clear_cache(uint32_t inode, uint32_t chunkIndex) {
 	std::unique_lock gMutexLock(gMutex);
 
 	ReadRecordRange range = gActiveReadRecords.equal_range(inode);
@@ -682,7 +682,7 @@ void read_inode_reconnect_clear(uint32_t inode, uint32_t chunkId) {
 		if (!it->second->expired) {
 			it->second->refreshCounter = REFRESHTICKS;  // force reconnect on forthcoming access
 			std::unique_lock inodeLock(it->second->mutex);
-			it->second->cache.selective_clear(chunkId);  // clear cache for this chunk
+			it->second->cache.selective_clear(chunkIndex);  // clear cache for this chunk
 		}
 	}
 }
--- a/src/mount/readdata.cc
+++ b/src/mount/readdata.cc
@@ -672,6 +672,21 @@ void read_inode_ops(uint32_t inode) { // attributes of inode have been changed -
 	}
 }
 
+// Ongoing write on this inode and chunkId, force reconnect and clear cache for this chunk
+void read_inode_reconnect_clear(uint32_t inode, uint32_t chunkId) {
+	std::unique_lock gMutexLock(gMutex);
+
+	ReadRecordRange range = gActiveReadRecords.equal_range(inode);
+
+	for (auto it = range.first; it != range.second; ++it) {
+		if (!it->second->expired) {
+			it->second->refreshCounter = REFRESHTICKS;  // force reconnect on forthcoming access
+			std::unique_lock inodeLock(it->second->mutex);
+			it->second->cache.selective_clear(chunkId);  // clear cache for this chunk
+		}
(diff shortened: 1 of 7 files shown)


src/mount/readdata.cc
```diff
@@ -672,6 +672,21 @@ void read_inode_ops(uint32_t inode) { // attributes of inode have been changed -
...
+	ReadRecordRange range = gActiveReadRecords.equal_range(inode);
+
+	for (auto it = range.first; it != range.second; ++it) {
+		if (!it->second->expired) {
+			it->second->refreshCounter = REFRESHTICKS;  // force reconnect on forthcoming access
+			std::unique_lock inodeLock(it->second->mutex);
+			it->second->cache.selective_clear(chunkId);  // clear cache for this chunk
+		}
+	}
+}
```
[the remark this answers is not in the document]
dmga44: The mutexes are nested because each of them is required for protection of the structures they protect (`gMutex` protects the `gActiveReadRecords` and the `inodeLock` protects the cache). The issue is that in some other places in the code the `ReadRecord` were are iterating could be released, so if the `gMutex` is not locked then we would have SEGFAULT if we follow your proposal.

src/mount/readdata_cache.h
```diff
@@ -410,6 +410,20 @@ class ReadCache {
...
+		auto it = entries_.begin();
+		while (it != entries_.end()) {
+			// If the entry is not in the chunk, skip it
+			if (it->endOffset() < chunkStart || it->offset >= chunkEnd) {
+				++it;
+				continue;
+			}
+			it = erase(it);
+		}
+	}
```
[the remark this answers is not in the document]
dmga44: The erase makes sure that the return value is the next iterator.

src/mount/readdata_cache.h
```diff
@@ -410,6 +410,20 @@ class ReadCache {
 		}
 	}
 
+	void selective_clear(uint32_t chunkId) {
```
lgsilva3087: Consider chunkIndex instead of chunkId, which is used for other purposes.

src/mount/readdata_cache.h
```diff
@@ -410,6 +410,20 @@ class ReadCache {
...
+		uint64_t chunkStart = chunkId * SFSCHUNKSIZE;
+		uint64_t chunkEnd = chunkStart + SFSCHUNKSIZE;
+		auto it = entries_.begin();
```
lgsilva3087, line 423: I don't know the expected number of entries, if needed, please consider:
- Using lower_bound to go directly to the interesting ones (assuming the intrusive set keeps them sorted).
- Divide the condition to avoid iterating over the last entries after the chunkEnd:
```
if (it->offset >= chunkEnd) {
	break;
}
```

src/mount/readdata.h
```diff
@@ -382,6 +382,7 @@ uint32_t read_data_get_total_read_timeout_ms();
 bool read_data_get_prefetchxorstripes();
 
 void read_inode_ops(uint32_t inode);
+void read_inode_reconnect_clear(uint32_t inode, uint32_t chunkId);
```
lgsilva3087: Consider renaming it to: read_inode_reconnect_and_clear_cache.

lgsilva3087 (approved):
Looks good, please see my suggestions.

tests/test_suites/LongSystemTests/test_sqlite_stress_test_long.sh
```diff
@@ -0,0 +1,26 @@
+timeout_set 1 hour
```
rolysr: Consider to check the total time execution for this test to be as long as the other from this same test suite.

dmga44: I reduced the amount of ops it does in total, I also reduced the expected running time.

src/mount/readdata.h
dmga44: Done.

src/mount/readdata_cache.h
dmga44: Indeed, renamed, thanks.

src/mount/readdata_cache.h
dmga44: I thought that in the end, these kind of workloads where several threads write and read though the same mountpoint to the same file would perform bad anyway and this would be the less impacting part, but I get your point it would be nice not to waste time here.

I've applied the suggestions to make it iterate over the possible set of entries that could overlap the target chunk. Thanks.

rolysr (approved):

PR #343, opened 2025-04-08
feat(mount): Reduce write memory usage

dmga44:
This PR targets the apparent high memory usage of our client after
some write operation is performed.

The usage of new/delete[] on the write cache blocks during intensive
writes could lead to an increased memory fragmentation. The
substitution with an analogous vector seeks to reduce it, while
reducing the complexity of maintaining the code.

It also adds the malloctrimperiod option to the FUSE client. This
option targets faster release of the memory used by the client the
Linux kernel. By default, it keeps current behavior (let the OS reclaim
the memory).

Co-authored-by: Rolando Sánchez Ramos <rolysr@leil.io>


src/mount/write_cache_block.cc
```diff
@@ -33,12 +33,13 @@ WriteCacheBlock::WriteCacheBlock(uint32_t chunkIndex, uint32_t blockIndex, Type
 		  to(0),
 		  type(type) {
 	sassert(blockIndex < SFSBLOCKSINCHUNK);
-	blockData = new uint8_t[SFSBLOCKSIZE];
+	blockData.reserve(SFSBLOCKSIZE);
+	blockData.resize(SFSBLOCKSIZE);
```
lgsilva3087: Please check if only the call to resize is needed here.

src/mount/write_cache_block.cc
```diff
@@ -33,12 +33,13 @@ WriteCacheBlock::WriteCacheBlock(uint32_t chunkIndex, uint32_t blockIndex, Type
...
 }
 
 WriteCacheBlock::WriteCacheBlock(WriteCacheBlock&& block) noexcept {
-	blockData = block.blockData;
-	block.blockData = nullptr;
+	blockData = std::move(block.blockData);
+	block.blockData.clear();
```
lgsilva3087: Clearing the moved could be redundant here, because the ownership is now transferred to blockData. Please check if removing this line produces the same effect.

src/mount/fuse/mount_config.cc
```diff
@@ -324,6 +326,9 @@ void usage(const char *progname) {
 "    -o limitglibcmallocarenas=N  limit glibc malloc arenas to given value - prevents "
 				"from using huge amount of virtual memory. Use it in constrained memory "
 				"environments (default: %u)\n"
+"    -o malloctrimperiod=MSEC    set period of malloc trim in milliseconds. Could prove useful "
```
lgsilva3087: Consider something like:

"set the malloc trim period in milliseconds. This option can be useful in memory-constrained scenarios. When set to 0, periodic trimming is disabled."

lgsilva3087 (approved):
LGTM, please see my comments.

src/mount/sauna_client.cc
```diff
@@ -118,6 +119,19 @@ struct ReaddirSession {
...
+std::jthread gMallocTrimThread;
+
+void mallocTrimThread(const std::stop_token& stop, unsigned mallocTrimePeriod_ms) {
+	pthread_setname_np(pthread_self(), "mallocTrimThread");
+
+	while (!stop.stop_requested()) {
+		std::this_thread::sleep_for(std::chrono::milliseconds(mallocTrimePeriod_ms));
+		malloc_trim(0);
+	}
+}
```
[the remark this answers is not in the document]
dmga44: Well, I have not found any equivalent of malloc_trim for Windows that doesn't make me change dependencies. For now it has not being an issue.

src/mount/sauna_client.cc
```diff
@@ -118,6 +119,19 @@ struct ReaddirSession {
 	}
 };
 
+#ifndef _WIN32
+std::jthread gMallocTrimThread;
+
+void mallocTrimThread(const std::stop_token& stop, unsigned mallocTrimePeriod_ms) {
+	pthread_setname_np(pthread_self(), "mallocTrimThread");
```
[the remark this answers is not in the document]
dmga44: Done.

src/mount/fuse/main.cc
```diff
@@ -702,6 +703,13 @@ int main(int argc, char *argv[]) try {
 	gLimitGlibcArenas = gMountOptions.limitglibcmallocarenas;
 	tuneMalloc();
 
+	if (gMountOptions.malloctrimperiod > 0 && gMountOptions.malloctrimperiod < 100) {
+		// Try to avoid malloc_trim() being called too often
+		fprintf(stderr, "malloc trim period too small (%u ms) - increased to 100 ms\n",
+				gMountOptions.malloctrimperiod);
+		gMountOptions.malloctrimperiod = 100;
```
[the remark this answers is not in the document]
dmga44: Done.

src/mount/fuse/mount_config.cc
dmga44: Applied.

src/mount/write_cache_block.cc
dmga44: I've changed it to `std::vector<uint8_t> blockData = std::vector<uint8_t>(SFSBLOCKSIZE);` in the .h file to use it as default constructor.

src/mount/sauna_client.cc
```diff
@@ -118,6 +119,19 @@ struct ReaddirSession {
 	}
 };
 
+#ifdef __linux__
+std::jthread gMallocTrimThread;
+
+void mallocTrimThread(const std::stop_token& stop, unsigned mallocTrimePeriod_ms) {
+	pthread_setname_np(pthread_self(), "mallocTrimThread");
```
uristdwarf: This is a C library function and returns. You need check the return value if it ran successfully and then the errno if not (see man `pthread_setname_np`)

uristdwarf (approved):

ralcolea (approved):

src/mount/write_cache_block.cc
dmga44: Done.

src/mount/sauna_client.cc
dmga44: I actually did the same as other occurrences of this function in our code, but I think it would be nice to check it as you proposed.

Applied.

PR #346, opened 2025-04-09
feat: Disable creation of empty files for full disk

### LINKED ISSUE #317: SaunaFS Integration with logging tool via NFS Ganesha bad behavior - empty buckets creation
### Overview
We attempted to integrate SaunaFS as an archive solution with the logging tool using the NFS Ganesha Client. However, we encountered several issues that led us to abandon SaunaFS in favor of a traditional storage solution.

### Issues Identified
1. **Capacity Mismatch**
   - The capacity values displayed in the CGI interface differed from those shown in the terminal.
   
2. **Mounting Failure**
   - The SaunaFS storage did not mount automatically to the logging tool when using the `mount` command.
   
3. **Behavior at Full Capacity**
   - Once SaunaFS reached full capacity, logging tool continued to write empty buckets instead of ceasing logging operations.

### Alternative Solution Implemented
Due to these issues, we opted to use a traditional storage approach by deploying a VM and utilizing NFS Ganesha (`FSAL_VFS`) to export a VFS-based filesystem. Instead of exporting SaunaFS, we exported a local folder via NFS Ganesha.

Results of Alternative Solution
1. Logging tool correctly stopped writing logs when storage reached full capacity (no empty buckets were written).
2. The storage mounted automatically when using the `mount` command.
3. Capacity values were accurately reflected.

### Conclusion
The alternative solution resolved all issues we encountered with SaunaFS. To consider SaunaFS in the future, the following improvements would be necessary:
- Ensure consistency in capacity reporting between CGI and terminal.
- Implement correct behavior when storage reaches full capacity, preventing unnecessary writes of empty buckets.

ralcolea:
**Summary**
This PR introduces a new configuration option for the master server to control metadata creation behavior when disk space is depleted.

Previously, even when disk space was exhausted, metadata for files and directories could still be created—leading to the presence of empty files or folders. This update prevents such metadata from being created for folders when the system is out of space.

**Key Changes**

- A new option was added: `CREATE_EMPTY_FOLDERS_WHEN_SPACE_DEPLETED` (default: 1).
   - When set to `0`, metadata for empty folders will not be created if there is no available space.
   - This helps prevent inconsistent or misleading filesystem state when disk space is full.

- The new option modify the behavior of `fs_mkdir` function:
   - If the option is disabled and the disk is full, `fs_mkdir` operation returns `SAUNAFS_ERROR_NOSPACE` and skip metadata creation.

**Test coverage**

Considering the new option affects Linux and NFS clients, both clients were tested (with option enabled and disabled) under full-disk conditions:

- `test_nfs_ganesha_allow_empty_folders_metadata_after_full_disk.sh:` validates that enabling the option (default value) allows NFS clients to create empty folders when the space is depleted (expected behavior).

- `test_nfs_ganesha_disable_empty_folders_metadata_after_full_disk.sh:` validates that disabling the option (set value = 0)  prevent NFS clients to create empty folders when the space is depleted.

- `test_allow_empty_folders_metadata_after_full_disk.sh:` validates that enabling the option (default value) allows Linux clients to create empty folders when the space is depleted (expected behavior).

- `test_disable_empty_folders_metadata_after_full_disk.sh:` validates that disabling the option (set value = 0)  prevent Linux clients to create empty folders when the space is depleted.

**Related Issue**
Fixes: #317

### DIFF
--- a/doc/sfsmaster.cfg.5.adoc
+++ b/doc/sfsmaster.cfg.5.adoc
@@ -262,6 +262,12 @@ Set to either 1 to enable, or 0 to disable (default is 0)
 *PROMETHEUS_HOST (EXPERIMENTAL)*:: Host address where Prometheus metric data
 can be collected, must be in the format of HOST:PORT (default 0.0.0.0:9499)
 
+*CREATE_EMPTY_FOLDERS_WHEN_SPACE_DEPLETED (EXPERIMENTAL)* When enabled,
+this option allows the system to create metadata for empty folders when
+storage space is depleted. This helps preserving the intended file and
+directory structure. When this option is disabled, no metadata for folders
+will be created if there is insufficient storage space (default: 1).
+
 *LOG_LEVEL*:: Setup logging. Uses the environment variable SAUNAFS_LOG_LEVEL or
 config value LOG_LEVEL to determine logging level. Valid log levels are
 - 'trace'
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -28,6 +28,7 @@
 
 #include "common/attributes.h"
 #include "common/event_loop.h"
+#include "errors/saunafs_error_codes.h"
 #include "filesystem_metadata.h"
 #include "master/changelog.h"
 #include "master/chunks.h"
@@ -45,10 +46,18 @@
 #include "master/task_manager.h"
 #include "metrics/metrics.h"
 #include "protocol/matocl.h"
+#include "slogger/slogger.h"
 
 std::array<uint32_t, FsStats::Size> gFsStatsArray = {{}};
 
 [[maybe_unused]] static const char kAclXattrs[] = "system.richacl";
+
+inline bool isDepletedSpace() {
+	uint64_t totalSpace = 0;
+	uint64_t availableSpace = 0;
+	matocsserv_getspace(&totalSpace, &availableSpace);
+	return (totalSpace < SFSCHUNKSIZE || availableSpace < SFSCHUNKSIZE);
+}
 
 void fs_retrieve_stats(std::array<uint32_t, FsStats::Size> &output_stats) {
 	output_stats = gFsStatsArray;
@@ -1007,6 +1016,7 @@ uint8_t fs_mknod(const FsContext &context, uint32_t parent, const HString &name,
 	    fsnodes_quota_exceeded_dir(wd, {{QuotaResource::kInodes, 1}})) {
 		return SAUNAFS_ERROR_QUOTA;
 	}
+
 	static_cast<FSNodeDirectory *>(wd)->case_insensitive =
 	    context.sesflags() & SESFLAG_CASEINSENSITIVE;
 	p = fsnodes_create_node(ts, static_cast<FSNodeDirectory*>(wd), name, type, mode, umask, context.uid(), context.gid(), 0,
@@ -1055,6 +1065,14 @@ uint8_t fs_mkdir(const FsContext &context, uint32_t parent, const HString &name,
 	    fsnodes_quota_exceeded_dir(wd, {{QuotaResource::kInodes, 1}})) {
 		return SAUNAFS_ERROR_QUOTA;
 	}
+
+	if (gDisableEmptyFoldersMetadataOnFullDisk) {
+		if (isDepletedSpace()) {
+			safs::log_err("fs_mkdir: not enough space to create a folder");
+			return SAUNAFS_ERROR_NOSPACE;
+		}
+	}
+
 	static_cast<FSNodeDirectory *>(wd)->case_insensitive =
 	    context.sesflags() & SESFLAG_CASEINSENSITIVE;
 	p = fsnodes_create_node(ts, static_cast<FSNodeDirectory *>(wd), name, FSNode::kDirectory, mode,
--- a/src/admin/dump_config_command.cc
+++ b/src/admin/dump_config_command.cc
@@ -129,6 +129,7 @@ const static std::unordered_map<std::string, std::string> defaultOptionsMaster =
     {"SNAPSHOT_INITIAL_BATCH_SIZE_LIMIT", "10000"},
     {"FILE_TEST_LOOP_MIN_TIME", "3600"},
     {"PRIORITIZE_DATA_PARTS", "1"},
+    {"CREATE_EMPTY_FOLDERS_WHEN_SPACE_DEPLETED", "1"},
 };
 
 const static std::unordered_map<std::string, std::string> defaultOptionsShadow = {
--- a/src/data/sfsmaster.cfg.in
+++ b/src/data/sfsmaster.cfg.in
@@ -332,3 +332,8 @@
 ## port where it will listen for scrapers. Must in the format of HOST:PORT.
 ## (Default: 0.0.0.0:9499)
 # PROMETHEUS_HOST=0.0.0.0:9499
+
+## This option allows the system to create metadata for empty folders when
+## storage space is depleted.
+## (Default: 1)
+# CREATE_EMPTY_FOLDERS_WHEN_SPACE_DEPLETED = 1
--- a/src/master/filesystem.cc
+++ b/src/master/filesystem.cc
@@ -57,6 +57,8 @@ static bool gAutoRecovery = false;
 bool gMagicAutoFileRepair = false;
 bool gAtimeDisabled = false;
 
+bool gDisableEmptyFoldersMetadataOnFullDisk = false;
+
 uint32_t gTestStartTime;
 
 static bool gSaveMetadataAtExit = true;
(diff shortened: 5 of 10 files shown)


src/master/filesystem_operations.cc
```diff
@@ -45,11 +46,19 @@
...
 
 std::array<uint32_t, FsStats::Size> gFsStatsArray = {{}};
 
 [[maybe_unused]] static const char kAclXattrs[] = "system.richacl";
 
+inline bool isDepletedSpace() {
+	uint64_t totalSpace = 0;
+	uint64_t availableSpace = 0;
+	matocsserv_getspace(&totalSpace, &availableSpace);
+	return (totalSpace < SFSBLOCKSIZE || availableSpace < SFSBLOCKSIZE);
```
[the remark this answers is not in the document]
ralcolea: `SFSBLOCKSIZE` is the minimal value for us. If it's needed in the future, we can set a configurable threshold for this value, as you suggest :)

uristdwarf (approved):
While it might mitigate #317, the actual bug still needs to be found why the application is creating empty files

tests/test_suites/ShortSystemTests/test_allow_empty_files_metadata_after_full_disk.sh
```diff
@@ -0,0 +1,26 @@
...
+CHUNKSERVERS=1 \
+	USE_RAMDISK=YES \
+	CHUNKSERVER_EXTRA_CONFIG="READ_AHEAD_KB = 1024|MAX_READ_BEHIND_KB = 2048" \
+	MOUNT_EXTRA_CONFIG="sfscachemode=NEVER,sfsioretries=10" \
+	setup_local_empty_saunafs info
+
+cd ${info[mount0]}
+
+# Write a big file to deplete the space of 2 GiB
+assert_failure dd if=/dev/zero of="${info[mount0]}/largefile" bs=1M count=4096 \
```
antuan96314:
```suggestion
# Write a big file to deplete the space of 4 GiB
assert_failure dd if=/dev/zero of="${info[mount0]}/largefile" bs=1M count=4096 \
```
The comment about the space is not matching the size of the generated files

tests/test_suites/ShortSystemTests/test_allow_empty_files_metadata_after_full_disk.sh
```diff
@@ -0,0 +1,26 @@
...
+assert_failure dd if=/dev/zero of="${info[mount0]}/1MiBfile.1" bs=1M count=1
+assert_failure dd if=/dev/zero of="${info[mount0]}/1MiBfile.2" bs=1M count=1
+
+# Try to create a folder, which should be successful
+assert_success mkdir "${info[mount0]}/empty.folder"
+
+# Validate empty files were created
+assert_success ls -lh "${info[mount0]}/1MiBfile.1"
+assert_success ls -lh "${info[mount0]}/1MiBfile.2"
+
```
antuan96314:
```suggestion
assert_success touch "${info[mount0]}/file.empty"
assert_success touch "${info[mount0]}/empty.folder/file.empty"
assert_equal 0 $(stat "${info[mount0]}/file.empty" --format="%s")
assert_equal 0 $(stat "${info[mount0]}/empty.folder/file.empty" --format="%s")
```
Metadata operations should be successful even when there is no space left.

tests/test_suites/ShortSystemTests/test_disable_empty_files_metadata_after_full_disk.sh
```diff
@@ -0,0 +1,32 @@
...
+	USE_RAMDISK=YES \
+	MASTER_EXTRA_CONFIG="CREATE_EMPTY_FILES_AND_FOLDERS_WHEN_SPACE_DEPLETED = 0" \
+	CHUNKSERVER_EXTRA_CONFIG="READ_AHEAD_KB = 1024|MAX_READ_BEHIND_KB = 2048" \
+	MOUNT_EXTRA_CONFIG="sfscachemode=NEVER,sfsioretries=10" \
+	setup_local_empty_saunafs info
+
+cd ${info[mount0]}
+
+# Write a big file to deplete the space of 2 GiB
+assert_failure dd if=/dev/zero of="${info[mount0]}/largefile" bs=1M count=4096 \
```
antuan96314:
```suggestion
# Write a big file to deplete the space of 2 GiB
assert_failure dd if=/dev/zero of="${info[mount0]}/largefile" bs=1M count=2048 \
```
To match value on the comment line.

tests/test_suites/ShortSystemTests/test_disable_empty_files_metadata_after_full_disk.sh
```diff
@@ -0,0 +1,32 @@
...
+# Wait for data to be written and space to be synced
+sleep 3
+
+ls -lh "${info[mount0]}"
+
+# Try to create a file, which should fail
+assert_failure dd if=/dev/zero of="${info[mount0]}/1MiBfile" bs=1M count=1
+
+# Try to create a folder, which should fail
+assert_failure mkdir "${info[mount0]}/empty.folder"
```
antuan96314: This type of operation should not fail. If the cluster has no space left, this metadata operations are still valid and should passed. Same as `touch file.empty`. Only write operation to create or append data to chunks should fail.

doc/sfsmaster.cfg.5.adoc
```diff
@@ -225,6 +225,13 @@ specified time in seconds (default is 3600). It's possible for the loop to take
 more time if the master server is busy or the machine doesn't have enough
 processing power to make all the needed calculations.
 
+*CREATE_EMPTY_FILES_AND_FOLDERS_WHEN_SPACE_DEPLETED* When enabled (default: 1),
```
antuan96314: I don't see the benefit of this option just to force a failure of some operations that should passed when there is no space left on the cluster. I think that if cluster is full only write and append data to chunks operations should fail but `mkdir dir` and `touch dir/file.empty` should still pass without issues.

src/master/filesystem_operations.cc
```diff
@@ -1001,6 +1010,14 @@
 	    fsnodes_quota_exceeded_dir(wd, {{QuotaResource::kInodes, 1}})) {
 		return SAUNAFS_ERROR_QUOTA;
 	}
+
+	if (gDisableEmptyFilesMetadataOnFullDisk) {
+		if (isDepletedSpace()) {
+			safs::log_warn("fs_mknod: not enough space to create a file");
```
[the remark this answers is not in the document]
wg1k: The opposite: It is expected to ring the bells when no space is available, warning priority might be conservative.

src/master/filesystem_operations.cc
```diff
@@ -1049,6 +1066,14 @@
 	    fsnodes_quota_exceeded_dir(wd, {{QuotaResource::kInodes, 1}})) {
 		return SAUNAFS_ERROR_QUOTA;
 	}
+
+	if (gDisableEmptyFilesMetadataOnFullDisk) {
+		if (isDepletedSpace()) {
+			safs::log_warn("fs_mkdir: not enough space to create a folder");
```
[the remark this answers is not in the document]
wg1k: Same comment. It is expected to ring the bells when no space is available, warning priority might be conservative.

tests/test_suites/ShortSystemTests/test_allow_empty_files_metadata_after_full_disk.sh
wg1k: The comment refers to the 2GiB of ramdisk, however it is not specific enough allowing for confusion, like here.
```suggestion
# Write a big file to deplete the space of 2 GiB of RAMDISK
assert_failure dd if=/dev/zero of="${info[mount0]}/largefile" bs=1M count=4096 \
```

ralcolea: The real space is 2GiB and 4GiB file is copied to deplete this space.
I can update it, but it was written intentionally :)

wg1k (commented):
Why to make it optional rather than fixing the behavior in a deterministic way?

tests/test_suites/ShortSystemTests/test_disable_empty_files_metadata_after_full_disk.sh
ralcolea: Same case as before.

uristdwarf:
> Why to make it optional rather than fixing the behavior in a deterministic way?

Because this is not normal POSIX behavior. Empty files can still be created if there's enough metadata space on other filesystems.

uristdwarf:
To add, this is a mitigation for #317, not a fix.

tests/test_suites/ShortSystemTests/test_disable_empty_files_metadata_after_full_disk.sh
uristdwarf: The question I have is when these write operations happen? After client has accepted the write or when client is actually writing the chunks? From my discussion with @dmga44 it may be the latter. But it was mostly about chunkservers, not master though.

uristdwarf: Because if it's the latter, then it may be too late

ralcolea: I agree with you. That's why this option is by default enabled and allows metadata operations (as `touch file.empty`) even if space is depleted.

doc/sfsmaster.cfg.5.adoc
ralcolea: This option is not intended for us, but for the team who registered issue [317](https://github.com/leil-io/saunafs/issues/317). Unfortunately, their application does not handle correctly the empty files created by our system when the space of the cluster is depleted.

antuan96314 (approved):

wg1k (approved):

src/master/filesystem_operations.cc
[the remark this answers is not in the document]
ralcolea: I agree with your suggestion. Applied for `fs_mkdir`, which was the only function needed in the final version.

src/master/filesystem_operations.cc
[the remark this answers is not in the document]
ralcolea: I agree. Applied.

PR #348, opened 2025-04-13
fix(mount): Fix client to master response delay

rolysr:
Recent changes introduced new client-to-master communication functions, which caused the client to block when interacting with older versions of the master.

This commit ensures backward compatibility by preventing blocking behavior during communication with older master versions. Instead, the client will now display a proper warning message, encouraging users to update the master to a compatible version.


uristdwarf (approved):

dmga44 (approved):
LGTM.

PR #354, opened 2025-04-17
fix: Fix flaky SQLIite tests

dmga44:
The issue experienced in the SQLite tests seems to be related with the
following sequence of occurences:
- some thread starts asking for getattr at the db-wal (SQLite
write-ahead log).
- some other thread starts doing a write operation in that file.
- the value got from write_data_getmaxfleng is the one from before the
start of the write,
- the write (caching of the new data) finishes and updates the file
length kept in client. It also removes previous entries of that inode
from the DirEntryCache.
- the getattr asks to master to solve the getattr call, which didn't
have the size after the write yet, so it caches a wrong file size value
(lower than the one expected after write).
- the write data is flushed and the actual size in master side is
updated.
- some read is performed considering the size of the file lower than
the one expected after the write, and this read finds out that the file
is actually larger than expected: disk I/O error.

The main change is to move the call to write_data_getmaxfleng in the getattr
call from before the comunication with master to the moment after
comunicating with master. This change should cover all scenarios of
failing to retrieve correct file size.

Sometimes the check was not passing though the expected message was
seen at the end of the run log. It has been changed the way the
expected message is being searched for.

The tests were moved back to the regular testing.


ralcolea (approved):
LGTM :+1:

rolysr (approved):

PR #355, opened 2025-04-17
fix: Fix masterconn_canexit()

dmga44:
The masterconn_canexit function was crashing the master(shadow) process
when the first part of the condition checked was failing: the
gMasterConn was a nullptr and the second part of the OR was
SEGFAULTing.


ralcolea (approved):
Nice catch!!! @dmga44  :+1: :fire: :muscle:

lgsilva3087 (approved):
LGTM

PR #358, opened 2025-04-20
fix: chunkserver-common linking errors in GCC-13

ralcolea:
PR #339 introduced a dependency on the `saunafs_error_string()` function in the `chunkserver-common` library, specifically in files like `cmr_disk.cc` and `chunk_trash_manager.cc.` This led to undefined reference errors during linking with some compilers such as GCC 13, while newer compilers like GCC 14 did not report this issue.

This commit updates the `CMakeLists.txt` of the `chunkserver-common` library to explicitly link against the `sfserror` library, resolving the linking errors with GCC 13.


lgsilva3087 (approved):
LGTM

dmga44 (approved):

rolysr (approved):

PR #359, opened 2025-04-21
ci: Add clang-tidy comments for PR's

uristdwarf:
There are two seperate actions: One for running clang-tidy with no secrets, and another with secrets. This ensures that if the former is compromised, the latter secrets should be safe.

### DIFF
--- a/.github/workflows/post-static-analysis.yml
+++ b/.github/workflows/post-static-analysis.yml
@@ -1,0 +1,93 @@
+# Secure workflow with access to repository secrets and GitHub token for posting analysis results
+name: Post the static analysis results
+
+on:
+  workflow_run:
+    workflows: [ "Static analysis" ]
+    types: [ completed ]
+
+jobs:
+  clang-tidy-results:
+    # Trigger the job only if the previous (insecure) workflow completed successfully
+    if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-22.04
+    permissions:
+      pull-requests: write
+      # OPTIONAL: auto-closing conversations requires the `contents` permission
+      contents: write
+    steps:
+    - name: Download analysis results
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const artifacts = await github.rest.actions.listWorkflowRunArtifacts({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: ${{ github.event.workflow_run.id }},
+          });
+          const matchArtifact = artifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "clang-tidy-result"
+          })[0];
+          const download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: "zip",
+          });
+          const fs = require("fs");
+          fs.writeFileSync("${{ github.workspace }}/clang-tidy-result.zip", Buffer.from(download.data));
+    - name: Extract analysis results
+      run: |
+        mkdir clang-tidy-result
+        unzip -j clang-tidy-result.zip -d clang-tidy-result
+    - name: Set environment variables
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const assert = require("node:assert").strict;
+          const fs = require("fs");
+          function exportVar(varName, fileName, regEx) {
+              const val = fs.readFileSync("${{ github.workspace }}/clang-tidy-result/" + fileName, {
+                  encoding: "ascii"
+              }).trimEnd();
+              assert.ok(regEx.test(val), "Invalid value format for " + varName);
+              core.exportVariable(varName, val);
+          }
+          exportVar("PR_ID", "pr-id.txt", /^[0-9]+$/);
+          exportVar("PR_HEAD_REPO", "pr-head-repo.txt", /^[-./0-9A-Z_a-z]+$/);
+          exportVar("PR_HEAD_SHA", "pr-head-sha.txt", /^[0-9A-Fa-f]+$/);
+    - uses: actions/checkout@v4
+      with:
+        repository: ${{ env.PR_HEAD_REPO }}
+        ref: ${{ env.PR_HEAD_SHA }}
+        persist-credentials: false
+    - name: Redownload analysis results
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const artifacts = await github.rest.actions.listWorkflowRunArtifacts({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: ${{ github.event.workflow_run.id }},
+          });
+          const matchArtifact = artifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "clang-tidy-result"
+          })[0];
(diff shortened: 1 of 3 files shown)


dmga44 (approved):

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,77 @@
...
+jobs:
+  clang-tidy:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        submodules: true
+        fetch-depth: 0
+    - name: Setup anew (or from cache) vcpkg
```
ralcolea: Please, rebase this branch on top of dev and update this part with the current `run-unit-and-ganesha-tests.yml` workflow. This part suffered changes with vcpkg.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,77 @@
...
+      uses: lukka/run-vcpkg@v11
+
+    - name: Install packages
+      run: |
+        sudo ./tests/install-packages.sh
+    - name: Fetch base branch
```
ralcolea, line 33: Consider separating the jobs with a blank line to improve readability.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Create results directory
+      run: |
+        mkdir clang-tidy-result
+    - name: Analyze
+      run: |
+        git diff -U0 "$(git merge-base HEAD "upstream/${{ github.event.pull_request.base.ref }}")" | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
+
+    - name: Save PR metadata
+      run: |
+        echo "${{ github.event.number }}" > clang-tidy-result/pr-id.txt
```
ralcolea, line 91: I checked the output of this workflow and it only contains PR metadata.
What is the expected output of this workflow?
The list of warnings and suggestions offered by clang-tidy for the code?
I would like to see at least an example of the expected output to understand better the scope of this workflow.

.github/workflows/post-static-analysis.yml
```diff
@@ -0,0 +1,93 @@
...
+          })[0];
+          const download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: "zip",
+          });
+          const fs = require("fs");
+          fs.writeFileSync("${{ github.workspace }}/clang-tidy-result.zip", Buffer.from(download.data));
+    - name: Extract analysis results
```
ralcolea, line 39: Consider separating different steps with a blank line to improve readability.

.github/workflows/post-static-analysis.yml
```diff
@@ -0,0 +1,93 @@
...
+
+on:
+  workflow_run:
+    workflows: [ "Static analysis" ]
+    types: [ completed ]
+
+jobs:
+  clang-tidy-results:
+    # Trigger the job only if the previous (insecure) workflow completed successfully
+    if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
```
ralcolea, line 12: This workflow did not run as part of this PR. I was checking with some LLMs and it says that `post-static-analysis` workflow should be merged first to the default branch.

Can you confirm whether this is needed or the cause(s) preventing this workflow to run as part of this PR?

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,77 @@
+# Insecure workflow with limited permissions that should provide analysis results through an artifact
+name: Static analysis
+
+on: pull_request
```
ralcolea, line 4: According to Copilot, `workflow_run` event is triggered by workflows that are run on the `push`, `workflow_dispatch`, or `pull_request_target` events, but not for `pull_request`.

It suggested to change `pull_request` to `pull_request_target` event.

ralcolea (commented):
I shared some comments, despite this PR is outdated.
Please, rebase this branch on top of dev and give some feedback about its current status :)

.github/workflows/static-analysis.yml
uristdwarf: Seems to work fine.

.github/workflows/static-analysis.yml
uristdwarf: When there are code changes in the C++ code, it should output whether there are any clang-tidy warnings with the changes. Then post-static-analysis will analyze them. If no code has changed, then nothing will be outputted.

.github/workflows/post-static-analysis.yml
uristdwarf: I believe that's the case, yes.

.github/workflows/static-analysis.yml
uristdwarf: From the documentation:

pull_request:

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated. For activity related to pull request reviews, pull request review comments, or pull request comments, use the [pull_request_review](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review), [pull_request_review_comment](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review_comment), or [issue_comment](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#issue_comment) events instead. For information about the pull request APIs, see [Objects](https://docs.github.com/en/graphql/reference/objects#pullrequest) in the GraphQL API documentation or [REST API endpoints for pull requests](https://docs.github.com/en/rest/pulls).


pull_request_target:

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

Additionally, it's mentioned:

> Warning

> For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch. To help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered. For more information, see [Keeping your GitHub Actions and workflows secure: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) on the GitHub Security Lab website.

TL;DR: Copilot's idea is terrible.

ralcolea: Understood. Thank you for the explanation :)

.github/workflows/static-analysis.yml
ralcolea: Understood.

ralcolea (approved):
Nice job @uristdwarf 👍 
Please, remember to remove the test commit before merging :)

PR #362, opened 2025-04-23
ci: Use any build machines for building

uristdwarf:
With new build agents, we can afford to build both Ubuntu 24.04 and clang in any build machine available


ralcolea (approved):
Very nice change @uristdwarf :+1:

rolysr (approved):

PR #365, opened 2025-04-24
fix(main): Remove valgrind warning with clang

lgsilva3087:
When built with clang, the valgrind test was warning about a leak in the pam_open_session function call. This change ensures proper cleanup.


ralcolea (approved):
Nice catch @lgsilva3087 :+1: :clap:

dmga44 (approved):

PR #366, opened 2025-04-24
Fix: Correct quota per directory calculation

rolysr:
Previous changes related to adding quotas per directory as part of the volume size calculation did not correctly display the used space when a quota was set on the root directory. This issue was specifically observed on the Windows client. This commit resolves the issue.

### DIFF
--- a/src/master/matoclserv.cc
+++ b/src/master/matoclserv.cc
@@ -2054,16 +2054,37 @@ void matoclserv_sau_get_self_quota(matoclserventry *eptr, const uint8_t *data, u
 	std::vector<std::string> info;
 	uint8_t status;
 	deserializePacketVersionNoHeader(data, length, version);
-	
+
+	auto foundContextRootInodeResult = [&](uint32_t rootInode) {
+		for (const auto &result : results) {
+			if (result.entryKey.owner.ownerType == QuotaOwnerType::kInode && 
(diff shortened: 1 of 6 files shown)


src/master/matoclserv.cc
```diff
@@ -2054,16 +2054,37 @@ void matoclserv_sau_get_self_quota(matoclserventry *eptr, const uint8_t *data, u
...
+
+		if (!foundContextRootInodeResult(rootInode)) {
+			auto ino = fsnodes_id_to_node(rootInode);
+			statsrecord rootInodeStatRec;
+			fsnodes_get_stats(ino, &rootInodeStatRec);
+			results.emplace_back(
+				QuotaEntry{QuotaEntryKey{QuotaOwner{QuotaOwnerType::kInode, rootInode},
+											QuotaRigor::kUsed, QuotaResource::kSize},
+											rootInodeStatRec.size});
+		}
```
dmga44, line 2089: Please check if the size member of the `statsrecord` is the appropiate size of the displayed as used when mounting as a folder.

dmga44 (approved):
LGTM, please check my comment.

src/master/matoclserv.cc
rolysr: Done.

antuan96314 (approved):

PR #367, opened 2025-04-25
Fix chunkserver locking

lgsilva3087:
This change includes the following commits:
- Removal of unneeded mutex unlock before signaling a CV.
- Improve helgrind suppressions for GCC and Clang.
- Ensure the chunk is release even if the read failed (should fix pread_EIO related tests).

Co-authored-by: guillex <guillex@leil.io>


uristdwarf (approved):
Looks good!

As in https://github.com/leil-io/saunafs/pull/372#pullrequestreview-2798284296, consider keeping the debug logs you added before as trace logs for the future if needed. More logs are always good (as long as it's the right level)

rolysr (approved):

ralcolea (commented):
Great job @dmga44 and @lgsilva3087 :+1: :fire: :muscle:

PR #368, opened 2025-04-25
fix(clang): Fix error when prometheus is disabled

lgsilva3087:
Clang was reporting an unused resource when prometheus was disabled. This change moves the involved declaration to the block where it is used.


dmga44 (approved):

ralcolea (approved):
LGTM

PR #371, opened 2025-04-28
refactor(mount): Modernize pthread_mutex_t to std::mutex

dmga44:
List of changes:
- changed the pthread_mutex_t in several places for the modern
equivalent of std::mutex. This change also involved changing the
initialization and destruction scheme of several structures. As a
result the PthreadMutexWrapper was removed from our code.
- move the specific functions that were directly locking stats mutex
into the stats.cc file and substitute its occurrences for its
equivalent. Also, curly brackets were added and a goto: statement was
removed.
- a bit of extra refactor in the oplog.cc file: the pthread_cond_t was
modernized to std::condition_variable.

Co-authored-by: guillex <guillex@leil.io>

### DIFF
--- a/src/mount/fuse/sfs_meta_fuse.cc
+++ b/src/mount/fuse/sfs_meta_fuse.cc
@@ -50,14 +50,14 @@ struct dirbuf {
 	bool wasread;
 	uint8_t *p;
 	size_t size;
-	pthread_mutex_t lock;
+	std::mutex lock;
 };
 
 typedef struct _pathbuf {
 	int changed;
 	char *p;
 	size_t size;
-	pthread_mutex_t lock;
+	std::mutex lock;
 } pathbuf;
 
 #define READDIR_BUFFSIZE 50000
@@ -593,15 +593,7 @@ void sfs_meta_opendir(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 	if (ino == SPECIAL_INODE_ROOT || ino == SPECIAL_INODE_META_TRASH ||
 	    ino == SPECIAL_INODE_META_UNDEL || ino == SPECIAL_INODE_META_RESERVED) {
 
-		if (!(dirinfo = (dirbuf*)malloc(sizeof(dirbuf)))) {
-			safs_pretty_syslog(LOG_EMERG, "out of memory");
-			return;
-		}
-
-		if (pthread_mutex_init(&(dirinfo->lock), NULL)) {
-			free(dirinfo);
-			return;
-		}
+		dirinfo = new dirbuf();
 
 		dirinfo->p = NULL;
 		dirinfo->size = 0;
@@ -611,10 +603,7 @@ void sfs_meta_opendir(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 		if (fuse_reply_open(req,fi) == -ENOENT) {
 			fi->fh = 0;
 			free(dirinfo->p);
-			bool lock_destroy_failed = pthread_mutex_destroy(&(dirinfo->lock));
-			free(dirinfo);
-			if (lock_destroy_failed)
-				return;
+			delete dirinfo;
 		}
 	} else {
 		fuse_reply_err(req, ENOTDIR);
@@ -667,14 +656,13 @@ void sfs_meta_readdir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off, st
 		fuse_reply_err(req,EINVAL);
 		return;
 	}
-	pthread_mutex_lock(&(dirinfo->lock));
+	std::lock_guard lock(dirinfo->lock);
 	if (!dirinfo->wasread || (dirinfo->wasread && off==0)) {
 		if (dirinfo->p!=NULL) {
 			free(dirinfo->p);
 		}
 		if (ino == SPECIAL_INODE_ROOT) {
 			replyDirReadRoot(req, off, size);
-			pthread_mutex_unlock(&(dirinfo->lock));
 			return;
 		}
 		dirbuf_meta_fill(dirinfo, ino);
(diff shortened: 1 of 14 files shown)


src/mount/stats.h
```diff
@@ -40,6 +42,6 @@ statsnode* stats_get_subnode(statsnode *node, const char *name, uint8_t absolute
 uint64_t* stats_get_counterptr(statsnode *node);
 void stats_reset_all(void);
 void stats_show_all(char **buff, uint32_t *leng);
-void stats_lock(void);
-void stats_unlock(void);
 void stats_term(void);
+void stats_inc(uint8_t id, std::vector<uint64_t *> &statsptr, uint64_t inc = 1);
+void stats_dec(uint8_t id, std::vector<uint64_t *> &statsptr);
```
uristdwarf: Newline here

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -564,15 +564,7 @@ void sfs_meta_opendir(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
...
-		if (!(dirinfo = (dirbuf*)malloc(sizeof(dirbuf)))) {
-			safs_pretty_syslog(LOG_EMERG, "out of memory");
-			return;
-		}
-
-		if (pthread_mutex_init(&(dirinfo->lock), NULL)) {
-			free(dirinfo);
-			return;
-		}
+		dirinfo = new dirbuf();
```
ralcolea, line 596: This `dirbuf` instance is a good candidate to become smart pointer.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -691,22 +678,13 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
-		if (!(pathinfo = (pathbuf*)malloc(sizeof(pathbuf)))) {
-			safs_pretty_syslog(LOG_EMERG, "out of memory");
-			return;
-		}
-
-		if (pthread_mutex_init(&(pathinfo->lock), NULL)) {
-			free(pathinfo);
-			return;
-		}
+		pathinfo = new pathbuf();
```
ralcolea, line 742: This `pathbuf` instance is a good candidate to become a smart pointer.

src/mount/oplog.cc
```diff
@@ -216,13 +215,30 @@ unsigned long oplog_newhandle(int hflag) {
...
 	fhhead = fhptr;
-	pthread_mutex_unlock(&opbufflock);
 	return fhptr->fh;
 }
 
 void oplog_releasehandle(unsigned long fh) {
 	fhentry **fhpptr,*fhptr;
-	pthread_mutex_lock(&opbufflock);
+	std::lock_guard lock(opbufflock);
+	fhpptr = &fhhead;
```
ralcolea, line 223: Why not calling `oplog_releasedata()` function here if the logic is the same?

src/mount/special_open.cc
```diff
@@ -45,19 +45,15 @@ static void open(const Context &ctx, FileInfo *fi) {
 
 namespace InodeStats {
 static void open(const Context &ctx, FileInfo *fi) {
-	sinfo *statsinfo;
-	statsinfo = (sinfo*) malloc(sizeof(sinfo));
+	sinfo *statsinfo = new sinfo();
```
ralcolea: Consider using `auto *statsinfo` to avoid duplicating the type name.

ralcolea (approved):
Nice job @dmga44 and @lgsilva3087 :+1: :fire: :muscle: 
I added some minor suggestions, but none of them are stoppers to merge!!

src/mount/oplog.cc
```diff
@@ -216,13 +215,30 @@ unsigned long oplog_newhandle(int hflag) {
...
+	while ((fhptr = *fhpptr)) {
+		if (fhptr->fh==fh) {
```
antuan96314:
```suggestion
		if (fhptr->fh == fh) {
```

src/mount/oplog.cc
```diff
@@ -216,13 +215,30 @@ unsigned long oplog_newhandle(int hflag) {
...
+			fhptr->refcount--;
+			if (fhptr->refcount==0) {
```
antuan96314:
```suggestion
			if (!fhptr->refcount) {
```

antuan96314 (approved):

rolysr (approved):

src/mount/fuse/sfs_meta_fuse.cc
dmga44: I think it would be hard, please notice a few lines down this line:
`fi->fh = (unsigned long)pathinfo;`
which means that the `pathbuf` structure created in the client should be passed to the upper layer of the FUSE in a `uint64_t` field. So, if we would make the `pathbuf` (or the `dirbuf` structure that could be applied this comment also) a smart pointer, then we would lose the data after the open call because we can't copy or move the smart pointer to the reply.

src/mount/oplog.cc
dmga44: Very much true 👍 . Applied.

src/mount/special_open.cc
dmga44: Applied.

src/mount/oplog.cc
dmga44: Applied formatter over these lines, thanks.

src/mount/oplog.cc
dmga44: Applied.

src/mount/fuse/sfs_meta_fuse.cc
dmga44: Replied down, please check.

src/mount/fuse/sfs_meta_fuse.cc
ralcolea: You're right. I missed the lines you mentioned :sweat_smile: 
Nice catch :)

src/mount/fuse/sfs_meta_fuse.cc
ralcolea: You were right. I already saw what you mentioned :)

PR #372, opened 2025-04-28
fix: Fix truncate tests

dmga44:
The truncate tests in our testing framework have been unstable since
some time ago, failing from time to time. This commit targets to fix
those failures.

The detected failing pattern is the following:
- some truncate operation needs to be done in such a way that needs to
recalculate the last block of the parity parts blocks by writing some zeroes.
- the caching (write_blocks function) of those zeroes happens in such a
way that some of those blocks are processed first and some later on.
- the first blocks are got by the write workers, and synced to the
chunkservers with its load of zeroes.
- during the work of that write worker, the WriteChunkLocator (in this
case a TruncateWriteChunkLocator) is moved from the inodedata where it
was taken and never moved back.
- the second batch of blocks of zeroes is got by another write worker
but the inodedata which needs to have the correct locator (the one that
maintains the lockId) does not have it, so the locate and locking of
the chunk fails with SAUNAFS_ERROR_LOCKED.
- the write algorithm keeps retrying until time expires.

The proposed solution is the following:
- implement a shouldReset() member for the WriteChunkLocator with a
specific implementation for TruncateWriteChunkLocator.
- check for it right before going to destroy the locator in processJob
and right before going to reset the inodedata's locator in
write_job_delayed_end.
- the shouldReset function always returns true for the
WriteChunkLocator as it was doing. In the case of a
TruncateWriteChunkLocator, it compares the target size of the truncate
and the current committed size to the chunkservers.

### DIFF
--- a/src/mount/chunk_locator.h
+++ b/src/mount/chunk_locator.h
@@ -113,6 +113,10 @@ public:
 		return locationInfo_;
 	}
 
+	virtual bool shouldReset() const {
+		return true;
+	}
+
 protected:
 	WriteChunkLocator(uint32_t inode, uint32_t index, uint32_t lockId)
 			: inode_(inode),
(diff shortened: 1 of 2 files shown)


lgsilva3087 (approved):
LGTM, great.

ralcolea (approved):

uristdwarf (approved):
My only comment would be perhaps to include the logging you used to catch this bug as trace logs, as they can be useful in the future.

PR #373, opened 2025-04-30
chore(cmake): Pin the C standard to C23

lgsilva3087:
The default C standard changes from compilers or even compiler versions. This change pins the C standard for C23. The change also contains:

- Misc changes to modernize the file.
- Move the project definition to the beginning (recommended).


rolysr (approved):

ralcolea (approved):
LGTM :+1:

PR #374, opened 2025-05-02
build: Set GIT_* definitions to N/A if no git info generated

uristdwarf:
Previously, you were required to add ifdefs for GIT_INFO and GIT_COMMIT. This change falls back to "N/A" in the CMake file if it's not defined in any other way (e.g GENERATE_GIT_INFO is off).


uristdwarf:
For context, I found build bug while building debian packages (here: https://salsa.debian.org/hpc-team/saunafs), client was not using ifdefs as required for this (I've disabled git generation on the debian packages). So I decided if the definition is going to be used elsewhere other than common/version.h, it should be at least defined properly in CMakeLists.txt

lgsilva3087 (approved):

antuan96314 (approved):

PR #376, opened 2025-05-05
fix(tools): Fix saunafs command folder resolution

rolysr:
Previous changes to the path resolution for `saunafs` commands did not correctly handle folders with spaces in their names when executed. This issue was specifically present in the Windows version of the `saunafs` commands, particularly for paths formatted like `folder name\`. This commit solves that issue.


dmga44 (approved):
LGTM.

uristdwarf (approved):

PR #379, opened 2025-05-08
fix(master): Large trash causing crash

uristdwarf:
The current FUSE metadata mount is reading the entire trash from master, however the protocol is limited to UINT32_MAX - 8 bytes. When the trash data exceeds this, it's causing wrapping to the uint32 variable, which is then used to determine the malloc size of the memory to write the data, causing a buffer overflow. Besides, it will allocate a lot of memory for large trash sizes and hang master as it tries to respond with the *ENTIRE* trash to client.

The fix to this is limiting the amount of files that master will return. Currently, this is 1 million. There are some drawbacks with this approach: To see newer trash entries, you will need to delete or undelete files first.

A more general solution is actually implemented, which allows buffering the trash from client (SAU_CLTOMA_FUSE_GETTRASH) using max_entries. Unfortunately, this was never implemented for the FUSE meta mount. Another commit should be made to make the clients use this. But the focus of this commit is fixing CLTOMA_FUSE_GETTRASH packet, so metadata clients using this will not crash master.

### DIFF
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -39,6 +39,7 @@
 #include "master/filesystem_periodic.h"
 #include "master/filesystem_quota.h"
 #include "master/fs_context.h"
+#include "slogger/slogger.h"
 
 #ifndef NDEBUG
   #include "master/personality.h"
@@ -725,6 +726,7 @@ void fsnodes_getpath(FSNodeDirectory *parent, FSNode *child, std::string &path) 
 
 
 #ifndef METARESTORE
+constexpr uint32_t kOldPathContainerLimit = 1000000;
 
 template<class T>
 static inline uint32_t getdetachedsize(const T &data) {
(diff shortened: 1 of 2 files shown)


src/master/filesystem_node.cc
```diff
@@ -725,20 +726,27 @@ void fsnodes_getpath(FSNodeDirectory *parent, FSNode *child, std::string &path)
 
 
 #ifndef METARESTORE
+const uint32_t kOldPathContainerLimit = 1000000;
```
lgsilva3087: Consider to use constexpr to explicitely ask for compile time.

lgsilva3087: Consider making it a configuration option instead, this way the documentation will appear in the sfsmaster.cfg file, alerting the user about the limitation.

src/master/filesystem_node.cc
```diff
@@ -725,20 +726,27 @@ void fsnodes_getpath(FSNodeDirectory *parent, FSNode *child, std::string &path)
...
 
 template<class T>
 static inline uint32_t getdetachedsize(const T &data) {
 	static_assert(std::is_same<T, TrashPathContainer>::value
 	              || std::is_same<T, ReservedPathContainer>::value, "unsupported container");
```
lgsilva3087, line 734: Off-topic: looks a good place to use C++20's concepts to replace static assertions.

lgsilva3087 (approved):
Please see my suggestions.

src/master/filesystem_node.cc
uristdwarf: The main problem I see making it configurable is that it introduces user error to this problem. I would want to avoid that.

Adding user documentation is a good idea though.

src/master/filesystem_node.cc
uristdwarf: I could look into that later, I don't think it will be the last time I will be in this part of code.

ralcolea (approved):
Good job @uristdwarf :+1:

PR #391, opened 2025-05-13
fix(uraft): Remove floating IP on sfsmaster kill

ralcolea:
Previously, floating IP manager was not properly stopped when the sfsmaster process was terminated unexpectedly (e.g., kill sfsmaster). As a result, the failover mechanism incorrectly restored the floating IP to a node that was no longer the Leader. This led to multiple nodes holding the same floating IP simultaneously, causing IP conflicts and disrupting SaunaFS services.

This commit ensures the floating IP manager is correctly stopped during demotion events triggered by abnormal termination of the sfsmaster process.

### DIFF
--- a/src/uraft/floating-ip-manager.cc
+++ b/src/uraft/floating-ip-manager.cc
@@ -87,13 +87,12 @@ void HAFloatingIPManager::stopEventListener() {
 }
 
 void HAFloatingIPManager::handleIpLoss(const std::string &ipAddress) {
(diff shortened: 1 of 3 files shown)


lgsilva3087 (approved):
Great, LGTM.

antuan96314 (approved):

rolysr (approved):

wg1k (approved):

PR #393, opened 2025-05-15
ci: Add merge queue branches to CI

uristdwarf:
Idea is to explore using Github merge queues, to better coordinate merging multiple PR's that are approved and passed CI initially, but require updating each time one is merged to dev.

Note that this is not yet enabled. This commit just makes it possible to enable it.


rolysr (approved):

ralcolea (approved):
LGTM :+1:

PR #395, opened 2025-05-19
refactor: Remove deprecated method

wg1k:
Remove the deprecated changelogsMigrateFrom_1_6_29 method. It was
meant to run only once in an old version of the system. This commit
also removes the test created for verifying this function.

BREAKING CHANGE: Earlier versions than 1.9.29 changelog file names
might not be supported anymore.


ralcolea (approved):
LGTM :+1:

rolysr (approved):
LGTM.

PR #396, opened 2025-05-19
fix(chunkserver): Improve write buffers alignment

lgsilva3087:
Extend support to read from sockets into aligned buffers directly (without later memcpy) for the new following scenarios:

- Chunkservers in a standard replication chain.
- ChunkReplicator::replicate.

Reading into the aligned buffers helps with later DIRECT_IO support.


ralcolea (approved):

src/chunkserver/chunk_replicator.cc
```diff
@@ -28,6 +28,7 @@
 #include <string>
 
 #include "chunkserver/g_limiters.h"
+#include <common/aligned_allocator.h>
```
dmga44: Minor suggestion, but please consider using quotes for the includes.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -1263,8 +1263,7 @@ void ChunkserverEntry::checkNextPacket() {
 	const uint8_t *ptr;
 
 	if (state == State::WriteForward) {
-		if (mode == Mode::Data && inputPacket.bytesLeft == 0 &&
-		    fwdBytesLeft == 0) {
+		if (mode == Mode::Data && inputPacket.bytesLeft == 0 && fwdBytesLeft == 0) {
 			ptr = headerBuffer;
 			type = get32bit(&ptr);
 			opSize = get32bit(&ptr);
```
dmga44: Please notice that these if/else blocks are very similar (the actual changes inside the nested ifs are equal). These lines could be changed to:
```cpp
if ((state != State::WriteForward || fwdBytesLeft == 0) && mode == Mode::Data ||
	    inputPacket.bytesLeft == 0) {
		ptr = headerBuffer;
		type = get32bit(&ptr);
		opSize = get32bit(&ptr);

		mode = Mode::Header;
		inputPacket.bytesLeft = PacketHeader::kSize;
		inputPacket.startPtr = headerBuffer;

		if (inputPacket.useAlignedMemory) {
			gotPacket(type, inputPacket.alignedBuffer.data() + kIOAlignedOffset, opSize);
		} else {
			gotPacket(type, inputPacket.packet.data(), opSize);
		}
	}
```

dmga44 (approved):
Great work @lgsilva3087 .

src/chunkserver/chunkserver_entry.cc
dmga44: The downside is that it is less readable, just a suggestion.

src/chunkserver/chunk_replicator.cc
lgsilva3087: Fixed, sorry, it was added automatically.

src/chunkserver/chunkserver_entry.cc
lgsilva3087: I rewrote it like this:

```cpp
void ChunkserverEntry::checkNextPacket() {
	TRACETHIS();

	auto processNextPacket = [this]() {
		const uint8_t *ptr = headerBuffer;
		uint32_t type = get32bit(&ptr);
		uint32_t opSize = get32bit(&ptr);

		mode = Mode::Header;
		inputPacket.bytesLeft = PacketHeader::kSize;
		inputPacket.startPtr = headerBuffer;

		if (inputPacket.useAlignedMemory) {
			gotPacket(type, inputPacket.alignedBuffer.data() + kIOAlignedOffset, opSize);
		} else {
			gotPacket(type, inputPacket.packet.data(), opSize);
		}
	};

	if (state == State::WriteForward) {
		// the current packet has been fully read from the socket and written to the next
		// chunkserver in the chain
		if (mode == Mode::Data && inputPacket.bytesLeft == 0 && fwdBytesLeft == 0) {
			processNextPacket();
		}
	} else {
		// the current packet has been fully read from the socket
		if (mode == Mode::Data && inputPacket.bytesLeft == 0) { processNextPacket(); }
	}
}
```

PR #397, opened 2025-05-20
refactor(fuse): Partially refactor FUSE metadata mountpoint and add testing

uristdwarf:
This commit refactors out the metadata specific operations (that isn't touching the FUSE layer quite yet) to it's own library (saunafs_fuse_lib). The idea is to make it more easily testable the functions used while refactoring.

* Add saunafs_fuse_lib
* Refactor a bunch of metadata functions
* Write tests for those functions
* Add a more human-readable Attributes structure in common
* Cleanup headers in sauna_client (Some of them caused linkage problems to the new library)

### DIFF
--- a/src/common/attributes.h
+++ b/src/common/attributes.h
@@ -22,7 +22,61 @@
 
 #include "common/platform.h"
 
+#include <fcntl.h>
+#include <sys/stat.h>
+
+#include "protocol/SFSCommunication.h"
+
 #include <array>
 #include <cstdint>
 
 typedef std::array<uint8_t, 35> Attributes;
+
+constexpr int saunaFileTypeToPosix(unsigned char type) {
+	switch (type) {
+	case TYPE_BLOCKDEV:
+		return S_IFBLK;
+	case TYPE_CHARDEV:
+		return S_IFCHR;
+	case TYPE_DIRECTORY:
+		return S_IFDIR;
+	case TYPE_FIFO:
+		return S_IFIFO;
+	case TYPE_FILE:
+		return S_IFREG;
+	case TYPE_SOCKET:
+		return S_IFSOCK;
+	case TYPE_SYMLINK:
+		return S_IFLNK;
+	default:
+		return 0;
+	}
+}
+
+struct PosixFileAttributes {
+	uint8_t type;
+	uint16_t permissions;
+	uint32_t userID;
+	uint32_t groupID;
+	uint32_t accessTime;
+	uint32_t modificationTime;
+	uint32_t creationTime;
+	uint32_t hardLinkCount;
+	uint64_t size;
+
+	constexpr struct stat getStat() const {
+		struct stat fileStat{};
+		fileStat.st_mode = saunaFileTypeToPosix(type) | permissions;
+		fileStat.st_size = static_cast<off_t>(size);
+		// TODO(Urmas): Why are we returning to clients blocks of 512?
+		// Instead of S_BLKSIZE, it should SFSBLOCKSIZE
+		fileStat.st_blocks = (static_cast<off_t>(size) + S_BLKSIZE - 1) / S_BLKSIZE;
+		fileStat.st_uid = userID;
+		fileStat.st_gid = groupID;
+		fileStat.st_atime = accessTime;
+		fileStat.st_mtime = modificationTime;
+		fileStat.st_ctime = creationTime;
+		fileStat.st_nlink = hardLinkCount;
+		return fileStat;
+	}
+};
--- a/src/common/attributes_unittest.cc
+++ b/src/common/attributes_unittest.cc
@@ -1,0 +1,77 @@
+/*
+   Copyright 2025      Leil Storage OÜ
+
+   This file is part of SaunaFS.
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// NOLINTBEGIN
+
+#include <gtest/gtest.h>
+
+#include "common/attributes.h"
+
+TEST(Attributes, TypeToPosix) {
+
+	EXPECT_EQ(saunaFileTypeToPosix('q'), 0010000);
+	EXPECT_EQ(saunaFileTypeToPosix('c'), 0020000);
+	EXPECT_EQ(saunaFileTypeToPosix('d'), 0040000);
+	EXPECT_EQ(saunaFileTypeToPosix('b'), 0060000);
+	EXPECT_EQ(saunaFileTypeToPosix('f'), 0100000);
+	EXPECT_EQ(saunaFileTypeToPosix('l'), 0120000);
+	EXPECT_EQ(saunaFileTypeToPosix('s'), 0140000);
+	EXPECT_EQ(saunaFileTypeToPosix('\t'), 0);
+
+}
+
+TEST(Attributes, PosixFileAttrsToStat) {
+	PosixFileAttributes attrs = {
+		.type = 'f',
+		.permissions = 0456,
+		.userID = 567,
+		.groupID = 234,
+		.accessTime = 10,
+		.modificationTime = 412,
+		.creationTime = 794,
+		.hardLinkCount = 50,
+		.size = 1024,
+	};
+	struct stat converted = attrs.getStat();
+
+	EXPECT_EQ(converted.st_mode, 0100456U);
+	EXPECT_EQ(converted.st_size, 1024);
+	EXPECT_EQ(converted.st_uid, attrs.userID);
+	EXPECT_EQ(converted.st_gid, attrs.groupID);
+	EXPECT_EQ(converted.st_atime, attrs.accessTime);
+	EXPECT_EQ(converted.st_mtime, attrs.modificationTime);
+	EXPECT_EQ(converted.st_ctime, attrs.creationTime);
+	EXPECT_EQ(converted.st_nlink, attrs.hardLinkCount);
+	EXPECT_EQ(converted.st_blocks, 2U);
+
+	attrs.size = 1025;
+	converted = attrs.getStat();
+	EXPECT_EQ(converted.st_blocks, 3U);
+	attrs.size = 0;
+	converted = attrs.getStat();
+	EXPECT_EQ(converted.st_blocks, 0U);
+	attrs.size = 1;
+	converted = attrs.getStat();
+	EXPECT_EQ(converted.st_blocks, 1U);
+	attrs.size = 512;
+	converted = attrs.getStat();
+	EXPECT_EQ(converted.st_blocks, 1U);
+
+}
+
+// NOLINTEND
--- a/src/mount/fuse/main.cc
+++ b/src/mount/fuse/main.cc
@@ -41,6 +41,7 @@
 #include "mount/g_io_limiters.h"
 #include "mount/mastercomm.h"
 #include "mount/masterproxy.h"
+#include "mount/mount_info.h"
 #include "mount/option_casing_normalization.h"
 #include "mount/readdata.h"
 #include "mount/stats.h"
@@ -291,7 +292,7 @@ static int mainloop(struct fuse_args *args, struct fuse_cmdline_opts *fuse_opts,
 
 	struct fuse_session *se;
 	if (gMountOptions.meta) {
-		sfs_meta_init(gMountOptions.debug, gMountOptions.entrycacheto, gMountOptions.attrcacheto);
+		sfs_meta_init(gMountOptions.entrycacheto, gMountOptions.attrcacheto);
 		se = fuse_session_new(args, &sfs_meta_oper, sizeof(sfs_meta_oper), (void *)conn_opts);
 	} else {
 		se = fuse_session_new(args, &sfs_oper, sizeof(sfs_oper), (void *)conn_opts);
--- a/src/mount/fuse/sfs_fuse.cc
+++ b/src/mount/fuse/sfs_fuse.cc
@@ -19,6 +19,7 @@
  */
 
 #include "common/platform.h"
+
 #include "mount/fuse/sfs_fuse.h"
 
 #include <atomic>
@@ -33,11 +34,12 @@
 #include "common/special_inode_defs.h"
 #include "common/time_utils.h"
 #include "mount/fuse/lock_conversion.h"
+#include "mount/group_cache.h"
 #include "mount/sauna_client.h"
 #include "mount/sauna_client_context.h"
 #include "mount/thread_safe_map.h"
-#include "protocol/cltoma.h"
 #include "protocol/SFSCommunication.h"
+#include "protocol/cltoma.h"
 
 #include "grp.h"
 #include "pwd.h"
--- a/src/mount/fuse/sfs_meta_fuse.cc
+++ b/src/mount/fuse/sfs_meta_fuse.cc
@@ -18,17 +18,15 @@
    along with SaunaFS  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "common/massert.h"
 #include "common/platform.h"
-#include "mount/fuse/sfs_meta_fuse.h"
 
 #include <errno.h>
+#include <fuse_lowlevel.h>
 #include <pthread.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
-#include <fuse_lowlevel.h>
 #include <algorithm>
 #include <array>
 #include <cstdint>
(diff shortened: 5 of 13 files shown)


src/mount/fuse/fuselib/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+#  Copyright 2015-2017 Skytechnology sp. z o.o.
+#  Copyright 2023      Leil Storage OÜ
```
rolysr: Should this be changed to 2025 here? Should we also include the `Skytechnology sp. z o.o.` copyright?

src/mount/fuse/fuselib/metadata.cc
```diff
@@ -0,0 +1,88 @@
...
+	stbuf->st_blocks = 0;
+	stbuf->st_uid = 0;
+	stbuf->st_gid = 0;
+	stbuf->st_atime = now;
+	stbuf->st_mtime = now;
+	stbuf->st_ctime = now;
+
+	switch (inode) {
+	case SPECIAL_INODE_ROOT:
+		stbuf->st_nlink = 4;
```
rolysr: Maybe would be useful to add a little comment on why each of these `st_nlink` values.

src/mount/fuse/fuselib/metadata.cc
```diff
@@ -0,0 +1,88 @@
...
+	case SPECIAL_INODE_META_RESERVED:
+		stbuf->st_nlink = 2;
+		stbuf->st_mode = S_IFDIR | (nonRootAllowedToUseMeta() ? allowReadExecAll : allowReadExecRoot);
+		break;
+	default:
+		break;
+	}
+}
+
+uint32_t metadataNameToinode(const std::string &name) {
```
rolysr: Consider changing it to `metadataNameToInode`

src/mount/fuse/fuselib/metadata.cc
```diff
@@ -0,0 +1,88 @@
...
+	try {
+		inode = std::stoul(name, &end, base16);
+		if (name.at(end) == '|' && name.at(end + 1) != 0) {
+			return inode;
+		}
+	} catch (std::exception &e) {
+		safs::log_err("Could not convert metadata name to inode: {}", e.what());
+	}
+
+	return 0;
```
rolysr: you can consider returning `inode` variable here or modify the function so that when returning `inode` value, this one will contain the expected value for all cases.

src/mount/fuse/fuselib/metadata.h
```diff
@@ -0,0 +1,79 @@
...
+#include "common/attributes.h"
+#include "common/special_inode_defs.h"
+#include "mount/sauna_client.h"
+#include "protocol/SFSCommunication.h"
+
+enum : uint16_t {
+	READDIR_BUFFSIZE = 50000,
+	PATH_SIZE_LIMIT = 1024,
+	META_ROOT_MODE = 0555,
+};
```
rolysr: Would it be better to add some comment on why these values were selected? Or is there a way to understand this why if needed?

src/mount/fuse/sfs_meta_fuse.h
```diff
@@ -37,4 +37,4 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi);
 void sfs_meta_release(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi);
 void sfs_meta_read(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off, struct fuse_file_info *fi);
 void sfs_meta_write(fuse_req_t req, fuse_ino_t ino, const char *buf, size_t size, off_t off, struct fuse_file_info *fi);
-void sfs_meta_init(int debug_mode_in,double entry_cache_timeout_in,double attr_cache_timeout_in);
+void sfs_meta_init(double entry_cache_timeout_in,double attr_cache_timeout_in);
```
rolysr: Consider adding spaces between params here.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -873,11 +775,9 @@ void sfs_meta_write(fuse_req_t req, fuse_ino_t ino, const char *buf, size_t size
 	fuse_reply_write(req,size);
 }
 
-void sfs_meta_init(int debug_mode_in,double entry_cache_timeout_in,double attr_cache_timeout_in) {
-	debug_mode = debug_mode_in;
+void sfs_meta_init(double entry_cache_timeout_in,double attr_cache_timeout_in) {
```
rolysr: Consider also here adding spaces between params.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -343,7 +260,7 @@ void sfs_meta_unlink(fuse_req_t req, fuse_ino_t parent, const char *name) {
 		fuse_reply_err(req,EACCES);
 		return;
 	}
-	inode = sfs_meta_name_to_inode(name);
+	inode = metadataNameToinode(name);
```
rolysr: When changing function name to `metadataNameToInode` remember to update it in all these places.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -52,26 +53,15 @@ struct dirbuf {
...
-#define META_ROOT_MODE 0555
+};
 
 // 0x0124 = 0444
-#ifdef MASTERINFO_WITH_VERSION
-static Attributes masterinfoattr={{'f', 0x01,0x24, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,1, 0,0,0,0,0,0,0,14}};
-#else
-static Attributes masterinfoattr={{'f', 0x01,0x24, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,1, 0,0,0,0,0,0,0,10}};
-#endif
+constexpr Attributes kMasterInfoAttr={{'f', 0x01,0x24, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,1, 0,0,0,0,0,0,0,14}};
```
rolysr: Will we use MASTERINFO_WITH_VERSION always for now on? Won't this break any compatibility or configuration?

src/common/attributes_unittest.cc
```diff
@@ -0,0 +1,76 @@
...
+TEST(Attributes, TypeToPosix) {
+
+	EXPECT_EQ(saunaFileTypeToPosix('q'), 0010000);
+	EXPECT_EQ(saunaFileTypeToPosix('c'), 0020000);
+	EXPECT_EQ(saunaFileTypeToPosix('d'), 0040000);
+	EXPECT_EQ(saunaFileTypeToPosix('b'), 0060000);
+	EXPECT_EQ(saunaFileTypeToPosix('f'), 0100000);
+	EXPECT_EQ(saunaFileTypeToPosix('l'), 0120000);
+	EXPECT_EQ(saunaFileTypeToPosix('s'), 0140000);
+	EXPECT_EQ(saunaFileTypeToPosix('\t'), 0);
```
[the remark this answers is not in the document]
uristdwarf: This is intended. Stuff in SFSCommunication.h is not meant to be changed. In general, I believe that test code should not rely on definitions/variables in actual code, as those can change and introduce breaking changes which this test would then not catch..
I was thinking of using named variables, but I don't think this increases readability.

Regarding using things like S_IFIFO, gtest does not allow using EXPECT_EQ between different types (even if they are comparable), which causes problems using these defines.

src/mount/fuse/fuselib/metadata.cc
```diff
@@ -0,0 +1,88 @@
...
+uint32_t metadataNameToinode(const std::string &name) {
+	uint32_t inode = 0;
+	size_t end = 0;
+	const int base16 = 16;
+
+	try {
+		inode = std::stoul(name, &end, base16);
+		if (name.at(end) == '|' && name.at(end + 1) != 0) {
+			return inode;
+		}
```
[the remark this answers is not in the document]
uristdwarf: Both are the previous behavior of this function. I'm not changing this behavior in the refactor, for now.

src/mount/fuse/fuselib/metadata_unittest.cc
```diff
@@ -0,0 +1,188 @@
...
+	EXPECT_EQ(fileStat.st_mode, 0140000U);
+
+	resetStat(1, 'b', fileStat);
+	EXPECT_EQ(fileStat.st_mode, 0060000U);
+
+	resetStat(1, 'c', fileStat);
+	EXPECT_EQ(fileStat.st_mode, 0020000U);
+
+	resetStat(1, '\t', fileStat);
+	EXPECT_EQ(fileStat.st_mode, 0U);
```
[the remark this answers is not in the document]
uristdwarf: Addressed above why this won't be done

src/mount/fuse/fuselib/metadata.cc
```diff
@@ -0,0 +1,88 @@
...
+#include <string>
+#include "mount/fuse/fuselib/metadata.h"
+
+#include "common/attributes.h"
+#include "common/special_inode_defs.h"
+#include "mount/exports.h"
+
+void sfs_meta_stat(uint32_t inode, struct stat *stbuf) {
+	time_t now = 0;
+	now = time(nullptr);
```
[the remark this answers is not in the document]
uristdwarf: @gemini-code-assist This is a relic from the previous function

src/mount/fuse/sfs_meta_fuse.cc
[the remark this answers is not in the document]
uristdwarf: @gemini-code-assist This variable is actually not needed anymore

src/mount/fuse/fuselib/CMakeLists.txt
uristdwarf: I think this copyright info is a mistake, because it's usually not included in CMakeLists.txt

src/mount/fuse/fuselib/metadata.cc
uristdwarf: I sadly don't know why. I've set it to 4 because that's what it was in the function I refactored.

src/mount/fuse/fuselib/metadata.cc
uristdwarf: inode can still be changed and returning it here could be wrong. Changing it would require complicating the function, which IMO does not help.

src/common/attributes.h
```diff
@@ -22,7 +22,69 @@
...
+		return S_IFSOCK;
+		break;
+	case TYPE_BLOCKDEV:
+		return S_IFBLK;
+		break;
+	case TYPE_CHARDEV:
+		return S_IFCHR;
+		break;
+	default:
+		break;
```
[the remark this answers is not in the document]
aNeutrino: @gemini-code-assist I agree especially that in the future one may accidentally put code between return and break and this would be death code which may be not noticed...
therefore I vote for removing breaks each time we use return. 
also in `default` we should put `return 0;` to be consistent
and avoid another potential illusion that code between `switch` and `return 0` which we have know would be applied to all !

src/common/attributes_unittest.cc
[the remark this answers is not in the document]
aNeutrino: @uristdwarf forgot to mention you @gemini-code-assist  let me repeat his comment:

> This is intended. Stuff in SFSCommunication.h is not meant to be changed. In general, I believe that test code should not rely on definitions/variables in actual code, as those can change and introduce breaking changes which this test would then not catch..
> I was thinking of using named variables, but I don't think this increases readability.
> 
> Regarding using things like S_IFIFO, gtest does not allow using EXPECT_EQ between different types (even if they are comparable), which causes problems using these defines.

src/common/attributes.h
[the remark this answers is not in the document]
aNeutrino: @gemini-code-assist your commit sugestion is very wrong... you propose to remove all cases ! instead of  removing breaks and adding `return 0;` in `defalut:` and removing return 0; after switch or replace it with return unreachable_place in function.

src/mount/fuse/fuselib/metadata.cc
aNeutrino: @gemini-code-assist do you know why ?

src/mount/fuse/sfs_meta_fuse.cc
aNeutrino: capital `I` was missing in `Inode`

aNeutrino (dismissed):
please read comments before merging :)

src/mount/fuse/fuselib/metadata.h
```diff
@@ -0,0 +1,83 @@
...
+
+struct MetaStat {
+	std::string name;
+	inode_t inode{};
+	char type{};
```
antuan96314: Consider using the new type `FSNode` recently introduced at [PR-492](https://github.com/leil-io/saunafs/pull/492)

antuan96314 (approved):

dmga44 (approved):
Great job @uristdwarf (and @rolysr for the follow-up).

src/common/attributes.h
```diff
@@ -1,7 +1,7 @@
 /*
    Copyright 2013-2014 EditShare
    Copyright 2013-2015 Skytechnology sp. z o.o.
-   Copyright 2023      Leil Storage OÜ
+   Copyright 2023-2025 Leil Storage OÜ
```
lgsilva3087: Consider to not change this lines. It means from 2023 to present.

src/mount/fuse/main.cc
```diff
@@ -42,6 +42,7 @@
 #include "mount/mastercomm.h"
 #include "mount/masterproxy.h"
 #include "mount/option_casing_normalization.h"
+#include "mount/mount_info.h"
```
lgsilva3087, line 44: Please sort.

src/mount/fuse/sfs_fuse.cc
```diff
@@ -34,6 +34,7 @@
 #include "common/time_utils.h"
 #include "mount/fuse/lock_conversion.h"
 #include "mount/sauna_client.h"
+#include "mount/group_cache.h"
```
lgsilva3087, line 37: Sort

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -2,7 +2,7 @@
    Copyright 2005-2010 Jakub Kruszona-Zawadzki, Gemius SA
    Copyright 2013-2014 EditShare
    Copyright 2013-2019 Skytechnology sp. z o.o.
-   Copyright 2023      Leil Storage OÜ
+   Copyright 2023-2025 Leil Storage OÜ
```
lgsilva3087: See my previous comment about this.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -36,13 +36,14 @@
 #include <span>
 
 #include "common/datapack.h"
-#include "common/type_defs.h"
-#include "protocol/SFSCommunication.h"
-#include "slogger/slogger.h"
 #include "common/special_inode_defs.h"
+#include "common/type_defs.h"
```
lgsilva3087, line 39: Sorting

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -855,12 +761,9 @@ void sfs_meta_write(fuse_req_t req, fuse_ino_t ino, const char *buf, size_t size
...
-void sfs_meta_init(int debug_mode_in,double entry_cache_timeout_in,double attr_cache_timeout_in) {
-	debug_mode = debug_mode_in;
+void sfs_meta_init(double entry_cache_timeout_in, double attr_cache_timeout_in) {
 	entry_cache_timeout = entry_cache_timeout_in;
 	attr_cache_timeout = attr_cache_timeout_in;
-	if (debug_mode) {
-		fprintf(stdout, "cache parameters: entry_cache_timeout=%.2f attr_cache_timeout=%.2f\n",
-		        entry_cache_timeout, attr_cache_timeout);
-	}
+	fmt::print(stderr, "cache parameters: entry_cache_timeout={} attr_cache_timeout={}\n",
```
lgsilva3087: There are two differences here. Was it intended?
1. Printing to stderr instead of stdout.
2. The specific precision was dropped.

src/mount/sauna_client.h
```diff
@@ -22,23 +22,20 @@
...
-#include <unordered_set>
-#include <utility>
 #include <vector>
 
 #include "common/chunk_with_address_and_label.h"
 #include "common/exception.h"
-#include "common/type_defs.h"
 #include "mount/group_cache.h"
+#include "common/stat_defs.h"
+#include "common/type_defs.h"
```
lgsilva3087: Sorting

src/common/attributes.h
```diff
@@ -22,7 +22,60 @@
...
+#include <fcntl.h>
+#include <sys/stat.h>
+#include "protocol/SFSCommunication.h"
+
 #include <array>
 #include <cstdint>
 
 typedef std::array<uint8_t, 35> Attributes;
+
+inline constexpr int saunaFileTypeToPosix(unsigned char type) {
```
lgsilva3087: This inline in a header file could be redundant. Please check.

src/common/attributes.h
uristdwarf: I don't think that's the case. I would think it means what year was created and what year this file was significantly modified. But I don't have a strong opinion.

src/mount/fuse/sfs_meta_fuse.cc
uristdwarf: 1. Intended
2. Not intended

PR #401, opened 2025-05-21
build: Fix master not building on gcc12 when prometheus missing

uristdwarf:
On GCC12 debian, the master would not build because it's expecting jthread to be defined. However, in the ifdefs it's not included. Fix is to just include an empty destroy function that does nothing if prometheus is not compiled with.

### DIFF
--- a/src/metrics/metrics.cc
+++ b/src/metrics/metrics.cc
@@ -49,6 +49,18 @@
 
 namespace metrics {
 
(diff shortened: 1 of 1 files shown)


antuan96314 (approved):

ralcolea (approved):
It looks good. Please, rebase your branch and merge after all the tests will pass 👍

PR #406, opened 2025-05-22
perf: Skip metadata dump during chunk registration

ralcolea:
This commit prevents metadata dumps from running while chunks registration is active. This avoids a potential performance degradation caused by running two resource-intensive operations in parallel. The improvement is especially noticeable in large-scale deployments, where chunks registration can take several minutes and collide with the periodic metadata dump.

Additionally, logs were added to the dumping logic to better identify different stages of the forking process.

### DIFF
--- a/src/master/filesystem.cc
+++ b/src/master/filesystem.cc
@@ -119,11 +119,16 @@ static void metadataPollServe(const std::vector<pollfd> &pdesc) {
 }
 
 void fs_periodic_storeall() {
-	auto result = gMetadataBackend->fs_storeall(
-	    MetadataDumper::kBackgroundDump);  // ignore error
+	// Prevent metadata dump while chunks registration is in progress to prevent slowing down
+	// the chunks registration process
+	auto isChunkRegistrationInProgress = !gTimeoutSinceLastChunkRegistration.expired();
+	if (isChunkRegistrationInProgress) {
+		safs::log_info(
+		    "periodic metadata dump was skipped while chunks registration is in progress");
+		return;
+	}
 
-	safs_silent_syslog(LOG_DEBUG, "periodic metadata dump: %s",
-	                   (SAUNAFS_STATUS_OK == result) ? "success" : "failure");
+	gMetadataBackend->fs_storeall(MetadataDumper::kBackgroundDump);  // ignore error
 }
 
 void fs_term(void) {
--- a/src/master/matocsserv.cc
+++ b/src/master/matocsserv.cc
@@ -70,6 +70,15 @@ enum{KILL, CONNECTED};
 double gLoadFactorPenalty = 0.;
 bool gPrioritizeDataParts = true;
 
+// A safe threshold of 15 seconds to determine whether chunks registration is in progress and
+// prevent dumping metadata.
+constexpr auto kTimeoutForChunkRegistration = std::chrono::seconds(15);
+
+// Default value of 1 second to initialize the Timeout instance.
+// This value expires fast enough to not affect metadata dump when the cluster starts empty and
+// there is no chunks registration.
+Timeout gTimeoutSinceLastChunkRegistration(std::chrono::seconds(1));
+
 struct matocsserventry {
 	matocsserventry() : inputPacket(MaxPacketSize) {}
 
@@ -1112,6 +1121,8 @@ void matocsserv_sau_register_chunks(matocsserventry *eptr, const std::vector<uin
 			chunk_server_has_chunk(eptr, chunk.id, chunk.version, slice_traits::standard::ChunkPartType());
 		}
 	}
+	// Chunks registration is in progress, so we reset the timeout
+	gTimeoutSinceLastChunkRegistration = Timeout(kTimeoutForChunkRegistration);
 }
 
 void matocsserv_sau_register_space(matocsserventry *eptr, const std::vector<uint8_t>& data) {
--- a/src/master/metadata_dumper.cc
+++ b/src/master/metadata_dumper.cc
@@ -19,6 +19,7 @@
  */
 
 #include "common/platform.h"
+#include "slogger/slogger.h"
 
 #include "master/metadata_dumper.h"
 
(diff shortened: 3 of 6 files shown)


tests/test_suites/ShortSystemTests/test_metadata_prevent_dump_and_chunks_registration.sh
```diff
@@ -0,0 +1,44 @@
+timeout_set '1 minutes'
+period=10
+
+CHUNKSERVERS=1 \
+	MASTERSERVERS=1 \
+	MASTER_EXTRA_CONFIG="METADATA_DUMP_PERIOD_SECONDS = 10`
+			`|MAGIC_DEBUG_LOG = ${TEMP_DIR}/log|LOG_FLUSH_ON=DEBUG" \
```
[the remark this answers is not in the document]
ralcolea: Multiple lines using backsticks works correct for this test and is used in other tests.
Thank you anyway for your suggestion.

src/master/matocsserv.cc
```diff
@@ -1112,6 +1115,8 @@ void matocsserv_sau_register_chunks(matocsserventry *eptr, const std::vector<uin
 			chunk_server_has_chunk(eptr, chunk.id, chunk.version, slice_traits::standard::ChunkPartType());
 		}
 	}
+	// Reset the timer to 15 seconds after the last chunk registered
+	gTimeoutSinceLastChunkRegistration = Timeout(std::chrono::seconds(15));
```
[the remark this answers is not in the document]
ralcolea: I did not want to add a new parameter for an scenario that is unlikely to happen.
The 15-second timeout is appropriate for this scenario.
In case it will be needed, we can change this value or add a configurable parameter in the future.

rolysr: Maybe you could consider to put this value into a constant like `kDefaultTimeoutSinceLastChunkRegistration` or if not needed maybe add a text describing why this exact value was selected.

src/master/matocsserv.cc
```diff
@@ -70,6 +70,9 @@ enum{KILL, CONNECTED};
 double gLoadFactorPenalty = 0.;
 bool gPrioritizeDataParts = true;
 
+// Default value of 1 second to not affect metadata dump immediately
+Timeout gTimeoutSinceLastChunkRegistration(std::chrono::seconds(1));
```
rolysr: Consider adding a more descriptive comment here on why 1 second was the selected value.

rolysr (approved):
LGTM. Good job @ralcolea !

src/master/matocsserv.cc
ralcolea: Done.

src/master/matocsserv.cc
ralcolea: Nice suggestion :) Done.

src/master/filesystem.cc
```diff
@@ -119,11 +119,18 @@ static void metadataPollServe(const std::vector<pollfd> &pdesc) {
...
 void fs_periodic_storeall() {
-	auto result = gMetadataBackend->fs_storeall(
-	    MetadataDumper::kBackgroundDump);  // ignore error
+	// Prevent metadata dump while chunks registration is in progress to prevent slowing down
+	// the chunks registration process
+	auto isChunkRegistrationInProgress = !gTimeoutSinceLastChunkRegistration.expired();
+	if (isChunkRegistrationInProgress) {
+		safs::log_debug(
+		    "periodic metadata dump was skipped while chunks registration is in progress");
+		return;
```
uristdwarf: I think this should be info, not debug

src/master/filesystem.cc
```diff
@@ -119,11 +119,18 @@ static void metadataPollServe(const std::vector<pollfd> &pdesc) {
...
+	}
 
-	safs_silent_syslog(LOG_DEBUG, "periodic metadata dump: %s",
-	                   (SAUNAFS_STATUS_OK == result) ? "success" : "failure");
+	auto result = gMetadataBackend->fs_storeall(MetadataDumper::kBackgroundDump);  // ignore error
+	safs::log_debug("periodic metadata dump: {}",
+	                (SAUNAFS_STATUS_OK == result) ? "success" : "failure");
```
uristdwarf: Ditto. This is very important info

src/master/metadata_dumper.cc
uristdwarf, line 1: I'm less sure about the debugs in this file. But I would consider info because these logs don't happen that often, and they are still useful.

tests/test_suites/ShortSystemTests/test_metadata_prevent_dump_and_chunks_registration.sh
```diff
@@ -0,0 +1,44 @@
...
+while ! is_time_for_metadata_dump; do
+	echo "Waiting for next metadata dump, current seconds: $(date +%S)"
+	sleep 1
+done
+
+# Confirm the metadata dump is skipped during chunks registration in the master
+assert_eventually "grep -q 'metadata dump was skipped' ${TEMP_DIR}/log"
+
+# Confirm the metadata dump is successful in the shadow
+assert_eventually "grep -q 'periodic metadata dump: success' ${TEMP_DIR}/log"
```
uristdwarf, line 43: Maybe it's also prudent to check that it eventually dumps as well?

antuan96314 (approved):

src/master/filesystem.cc
ralcolea: Nice suggestion!! Done.

src/master/filesystem.cc
ralcolea: Done.

src/master/metadata_dumper.cc
ralcolea: Done.

tests/test_suites/ShortSystemTests/test_metadata_prevent_dump_and_chunks_registration.sh
ralcolea: I liked your idea. I extended the test to include your suggestion.

PR #408, opened 2025-05-23
ci: Prevent duplicate GitHub action workflow runs

uristdwarf:
This PR prevents running twice the Github Actions pipeline used for Ganesha and unit tests.
Now, the pipeline runs only for Pull Requests.

### DIFF
--- a/.github/workflows/run-unit-and-ganesha-tests.yml
+++ b/.github/workflows/run-unit-and-ganesha-tests.yml
@@ -1,11 +1,8 @@
 name: Run Unit and Ganesha Tests
 
 on:
-    push:
     merge_group:
     pull_request:
-        branches:
-            - "*"
 
 jobs:
     Run-Unit-Tests-and-Ganesha:


.github/workflows/run-unit-and-ganesha-tests.yml
```diff
@@ -13,6 +13,8 @@ jobs:
 
         env:
             SAUNAFS_TEST_TIMEOUT_MULTIPLIER: 2
+        if: github.event_name != 'pull_request' ||
```
ralcolea: Why not just remove Ganesha and unit tests pipeline for pull requests and use the result of the pipeline performed for the branch?

The only doubt is how this works for PRs created from external contributors. If in this case, the pipeline works for the branch, that needs to be created anyway, I think should be enough.
What do you think @uristdwarf?

uristdwarf: Because we might need to run outside PR's.
Alternatively, we could make it PR only and remove the branch (As the GH action is now slower than the Jenkins branch pipeline)

ralcolea: I also liked the idea of making it PR only. If you agree, we can add it to this PR :)

uristdwarf: Totally.

ralcolea: @uristdwarf I pushed a commit enabling the pipeline only for Pull Requests.
Please, confirm if you agree with this proposal or share any suggestion you might have :)

uristdwarf: LGTM

uristdwarf: You could probably remove my initial commit though and/or squash it

ralcolea: @uristdwarf I removed your initial commit and rebased the branch on top of dev.
I will assign reviewers to proceed with the PR.

Thank you very much for your time :)

lgsilva3087 (approved):
LGTM

rolysr (approved):
LGTM.

PR #410, opened 2025-05-25
fix: Modify maxReadAheadRequests default value

antuan96314:
Prevents timeouts and network overloads when small files are read.
May impact performance of sequential reading on big files.

### DIFF
--- a/doc/sfsmount.1.adoc
+++ b/doc/sfsmount.1.adoc
@@ -170,7 +170,7 @@ Set percentage of system memory used for max value of read cache size (default: 
 Define number of read workers (default: 30).
 
 *-o maxreadaheadrequests=*'N'::
-Define number of readahead requests per inode (default: 5).
+Define number of readahead requests per inode (default: 2).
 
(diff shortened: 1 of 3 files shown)


rolysr (commented):
Consider adding a description for the need of changing default value.

uristdwarf:
Please mark this as a breaking change per: https://www.conventionalcommits.org/en/v1.0.0/

Also, as @rolysr mentions, please add a description why we are changing. It's because having this value too high causes write/read timeouts with chunkservers (because they are being overloaded with these readahead requests).

antuan96314:
> Please mark this as a breaking change per: https://www.conventionalcommits.org/en/v1.0.0/
> 
> Also, as @rolysr mentions, please add a description why we are changing. It's because having this value too high causes write/read timeouts with chunkservers (because they are being overloaded with these readahead requests).

Why is this considered as BREAKING_CHANGE ?

lgsilva3087 (approved):
LGTM

dmga44 (approved):

PR #411, opened 2025-05-26
fix: Relax constraints in malloc tests

dmga44:
This change relaxes the check on the expected memory used in the malloc
arenas tests as it does not always imply a reduce in memory.


lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM :+1:

PR #413, opened 2025-05-26
Improve path by inode feature

rolysr:
This PR contains the following changes:
- Refactor path by inode special inode related code.
- Add check for preventing coredump on master side when requesting the name of deleted inode in a path for the path by inode special hidden file.

### DIFF
--- a/src/mount/path_by_inode.h
+++ b/src/mount/path_by_inode.h
@@ -1,0 +1,34 @@
+/*
+
+ Copyright 2023 Leil Storage OÜ
+
+ This file is part of SaunaFS.
+
+ SaunaFS is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, version 3.
+
+ SaunaFS is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "common/platform.h"
+
+#include <condition_variable>
+#include <mutex>
+#include <string>
+
+struct InodePathInfo {
+    std::string pathByInode;
+    inode_t inode = 0; 
+    std::mutex mtx;
+    std::condition_variable cv;
+    bool locked = false;
+};
+
+inline InodePathInfo gInodePathInfo;
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -941,13 +941,13 @@ EntryParam lookup(Context &ctx, inode_t parent, const char *name) {
 		if (endptr == nullptr || *endptr != '\0') {
 			throw RequestException(SAUNAFS_ERROR_EINVAL);
 		}
-		std::unique_lock<std::mutex> lock(InodePathByInode::inodePathInfo.mtx);
-		InodePathByInode::inodePathInfo.cv.wait(lock, [inode] {
-			return !InodePathByInode::inodePathInfo.locked ||
-			       InodePathByInode::inodePathInfo.inode == inode;
+		std::unique_lock<std::mutex> lock(gInodePathInfo.mtx);
+		gInodePathInfo.cv.wait(lock, [inode] {
+			return !gInodePathInfo.locked ||
+			       gInodePathInfo.inode == inode;
 		});
-		InodePathByInode::inodePathInfo.locked = true;
-		InodePathByInode::inodePathInfo.inode = inode;
+		gInodePathInfo.locked = true;
+		gInodePathInfo.inode = inode;
 		std::string fullPath = "";
 		int lookupStatus = SAUNAFS_STATUS_OK;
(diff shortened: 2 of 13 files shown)


src/mount/special_release.cc
```diff
@@ -105,7 +105,6 @@ static void release(FileInfo *fi) {
 	std::unique_lock<std::mutex> lock(inodePathInfo.mtx);
 	fi->fh = 0;
 	if (inodePathInfo.locked) {
-		free(inodePathInfo.pathByInode);
```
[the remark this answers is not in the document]
rolysr: Done.

src/mount/sauna_client.cc
```diff
@@ -951,14 +951,13 @@ EntryParam lookup(Context &ctx, Inode parent, const char *name) {
...
-			InodePathByInode::inodePathInfo.cv.notify_one();
+			inodePathInfo.locked = false;
+			inodePathInfo.cv.notify_one();
 			lock.unlock();
 			throw RequestException(status);
 		}
 		status = SAUNAFS_STATUS_OK;
-		InodePathByInode::inodePathInfo.pathByInode = new char[fullPath.length() + 1];
-		std::strcpy(InodePathByInode::inodePathInfo.pathByInode, fullPath.c_str());
+		inodePathInfo.pathByInode = fullPath;
```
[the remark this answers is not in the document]
rolysr: Done.

src/mount/path_by_inode.h
```diff
@@ -0,0 +1,39 @@
...
+    std::condition_variable cv;
+    bool locked = false;
+
+    InodePathInfo() = default;
+
+    InodePathInfo(const std::string& path, inode_t ino)
+        : pathByInode(path), inode(ino) {}
+};
+
+inline InodePathInfo inodePathInfo;
```
dmga44: Though I like the idea of this refactor, it completely drives away this special inode from the other special inodes, so if it were for me, I'd keep it the way it was and then refactor all the special inodes.

It appears this struct is supposed to be a singleton, please consider:
- renaming the variable to `gInodePathInfo`.
- deleting the default copy and move constructors from the struct.

dmga44 (approved):
Please consider my comment.

src/mount/path_by_inode.h
rolysr: Done. I kept the struct definition here as I wanted to clean the code from the `special_inodes.h` file and in the case this feature needs more changes, we could better maintain the logic and information handling, which could change.

tests/test_suites/TestTemplates/test_path_by_inode.inc
```diff
@@ -41,11 +42,17 @@ assert_equals "data5" "$(cat folder/file2)"
...
 
 # check removing files by their path
-rm $(cat .saunafs_path_by_inode/2)
+# first check the case when the file was deleted but
+# is still on the filesystem trash
+saunafs settrashtime 2 file1
+rm file1
 assert_failure cat file1
+assert_success cat .saunafs_path_by_inode/2
+assert_equals "file1 (trash)" "$(cat .saunafs_path_by_inode/2)"
```
ralcolea, line 58: I think in the test is missing the validation of path by inode for reserved files, which is also included as part of the improvement.

tests/test_suites/TestTemplates/test_path_by_inode.inc
```diff
@@ -1,5 +1,6 @@
 USE_RAMDISK=YES \
-	MOUNT_0_EXTRA_CONFIG="sfscachemode=NEVER,sfsdirentrycacheto=0" \
+	MOUNT_0_EXTRA_CONFIG="sfscachemode=NEVER,sfsreportreservedperiod=1,sfsdirentrycacheto=0" \
```
ralcolea, line 2: IMHO, using templates for testing offer more advantages when there is a common logic repeated across multiple tests. In this scenario, using templates is valuable because makes easier the maintenance of these tests.

However, when a template is used for a single test, is more difficult to understand and follow the logic of the test using the template.

My advice is to keep the template if you plan to add more tests related with the path by inode feature.
Otherwise is better to remove the template and move all the logic to the test.

ralcolea (commented):
Good work @rolysr.
Please, see my comments.

tests/test_suites/TestTemplates/test_path_by_inode.inc
rolysr: Done.

tests/test_suites/TestTemplates/test_path_by_inode.inc
rolysr: Yes, in this case, the reason of using a template here is because the logic for testing this feature on Windows client is the same that it is on the template. The difference is that this feature for the Linux client contains some more extra checking related to permissions, that are mainly focus on Linux environment.

ralcolea: Understood. In this case, the template is justified :)

tests/test_suites/TestTemplates/test_path_by_inode.inc
ralcolea: Perfect :+1:

ralcolea (approved):
Great work @rolysr!!! :+1:

lgsilva3087 (approved):
LGTM

PR #417, opened 2025-05-28
fix: Refactor saunafs-admin command

antuan96314:
A wrong line was printed when help argument was passed to saunafs-admin because the parser treated the value as argument instead of command.
 - Creates dedicated funcions to print help command and reuse logic.
 - More consistent return values from helper funtions.
 - Cleaner error message formatting with correct identations.
 - Clearer separation between options and commands.
 - Maintained all the original functionality while making it more modular.

### DIFF
--- a/src/admin/main.cc
+++ b/src/admin/main.cc
@@ -54,6 +54,41 @@
 #include "common/sockets.h"
 #include <common/version.h>
 
+int printVersion() {
+	std::cout << "SaunaFS Admin Tool\n\n";
+	std::cout << "Version: " << common::version() << "\n";
+	return 0;
+}
+
+void printCommandHelp(const SaunaFsAdminCommand* command) {
+	command->usage();
+	if (!command->supportedOptions().empty()) {
+		std::cerr << "  Possible command-line options:\n";
+		for (const auto& optionWithDescription : command->supportedOptions()) {
+			std::cerr << "\n    " << optionWithDescription.first << "\n";
+			std::cerr << "      " << optionWithDescription.second << "\n";
+		}
+	}
+	std::cerr << std::endl;
+}
+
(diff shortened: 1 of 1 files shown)


rolysr (approved):
LGTM.

src/admin/main.cc
```diff
@@ -53,6 +53,40 @@
...
+		std::cerr << "  Possible command-line options:\n";
+		for (const auto& optionWithDescription : command->supportedOptions()) {
+			std::cerr << "\n    " << optionWithDescription.first << "\n";
+			std::cerr << "      " << optionWithDescription.second << "\n";
+		}
+	}
+	std::cerr << std::endl;
+}
+
+int printHelp(const std::string& programName, const std::vector<const SaunaFsAdminCommand*>& commands, bool _printVersion = false) {
```
ralcolea: Line too long.

src/admin/main.cc
```diff
@@ -53,6 +53,40 @@
...
+	if (_printVersion) { printVersion(); }
+	std::cerr << std::endl;
+	std::cerr << "Usage:\n";
+	std::cerr << "    " << programName << " COMMAND [OPTIONS...] [ARGUMENTS...]\n\n";
+	std::cerr << "Available COMMANDs:\n\n";
+	for (auto command : commands) {
```
ralcolea: Use `const auto *` to supress compiler warning.

ralcolea (approved):
Great job @antuan96314 :+1: :muscle:

src/admin/main.cc
antuan96314: Fixed

src/admin/main.cc
antuan96314: Fixed

PR #420, opened 2025-05-29
feat!: Remove deprecated sfscachefiles option

ralcolea:
The `-o sfscachefiles` mount option, previously marked as deprecated and equivalent to `-o sfscachemode=YES`, has now been fully removed.

Changes include:
- Removal of the option from documentation and help text.
- Elimination of related parsing, validation, and warning logic.
- Removal of the `cachefiles` field from the `sfsopts_` structure.

This simplifies the mount option handling and completes the transition to `sfscachemode` ahead of the SaunaFS v5.0.0 release.

BREAKING CHANGE:
Users relying on -o sfscachefiles must switch to -o sfscachemode=YES. The old option is no longer recognized and will cause an error if used.

### DIFF
--- a/doc/sfsmount.1.adoc
+++ b/doc/sfsmount.1.adoc
@@ -111,10 +111,6 @@ Print some SaunaFS-specific debugging information.
 *-o sfscachemode=*'CACHEMODE'::
 Set cache mode (see *DATA CACHE MODES*; default is AUTO).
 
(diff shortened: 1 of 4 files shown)


rolysr (approved):
LGTM.

uristdwarf (approved):

PR #424, opened 2025-06-02
fix(master): Remove unavailable log spam

uristdwarf:
This can be quite normal when chunkservers have not yet started. It reports the total amount of unavailable files in one single log. Previous behaviour can be returned by setting log level to trace.


lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM :+1:

PR #425, opened 2025-06-02
fix(restore.cc): Correctly convert int to uint8_t in changelog restore

uristdwarf:
When parsing changelogs, the do_<operation> returns int's. Sometimes these are within range of SAUNAFS_ERROR_CODES (which is max 54). But sometimes these return -1 when they have trouble parsing the lines (or potentially other unknown values). saunafs_error_string however only accepts uint8_t, which means there's an implicit conversion and could cause unknown/faulty error messages.

This fixes that by explicitly checking the -1 to mean failure in parsing lines. For other values, an assert is used when NDEBUG is set (so the programmer is aware that they are returning a wrong value) or the reasonable default of UNKNOWN error with a log explaining why.

### DIFF
--- a/src/master/restore.cc
+++ b/src/master/restore.cc
@@ -928,12 +928,22 @@ int restore_line(const char* filename, uint64_t lv, const char* line) {
 #endif
 		safs_pretty_syslog(LOG_ERR, "%s:%" PRIu64 ": unknown entry '%s'", filename, lv, ptr);
 	} else if (status != SAUNAFS_STATUS_OK) {
+		uint8_t errorStatus = 0;
+		if (status == -1) {
+			errorStatus = SAUNAFS_ERROR_PARSE;
+		} else if (status > SAUNAFS_ERROR_MAX || status < 0) {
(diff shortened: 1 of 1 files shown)


lgsilva3087 (approved):
LGTM

src/master/restore.cc
```diff
@@ -928,12 +928,22 @@ int restore_line(const char* filename, uint64_t lv, const char* line) {
 #endif
 		safs_pretty_syslog(LOG_ERR, "%s:%" PRIu64 ": unknown entry '%s'", filename, lv, ptr);
 	} else if (status != SAUNAFS_STATUS_OK) {
+		uint8_t errorStatus = 0;
+		if (status == -1) {
```
rolysr, line 932: If there is an specific useful difference between `status == -1` and status equal to any other negative value, it think it would be useful to difference it with a constant, denoting the specific error here and avoiding using a magic number.

rolysr (approved):
LGTM.

src/master/restore.cc
uristdwarf: This would require changing this in a lot of of other places to make sense, so it's outside of the scope of this PR. But I don't disagree with this.

PR #427, opened 2025-06-03
feat(chunkserver): Make struct masterconn a class

lgsilva3087:
This change prepares the masterconn module of the chunkserver to support multiple MDS connections in the future. The main changes are:

- MasterConn is now a class with more responsibilities, which fosters later support for multiple connections.
- Simpler callbacks handling from the global masterconn functions.
- Simplified polling.
- Small documentation (will continue).

Next commits will move the MasterConn class to own files, to keep simple the management of multiple connections in masterconn.h and masterconn.cc.

### DIFF
--- a/src/chunkserver/init.h
+++ b/src/chunkserver/init.h
@@ -39,11 +39,11 @@ inline const std::vector<RunTab> runTabs = {
     RunTab{loadPlugins, "plugin manager"}, RunTab{hddInit, "hdd space manager"},
     // Has to be before "masterconn"
     RunTab{mainNetworkThreadInit, "main server module"},
+    RunTab{masterconn_init_threads, "master connection module - threads"},
     RunTab{masterconn_init, "master connection module"},
     RunTab{chartsdata_init, "charts module"}};
 
 /// Functions to call delayed after the initialization is correct
 inline const std::vector<RunTab> lateRunTabs = {
-    RunTab{masterconn_init_threads, "master connection module - threads"},
     RunTab{hddLateInit, "hdd space manager - threads"},
     RunTab{mainNetworkThreadInitThreads, "main server module - threads"}};
--- a/src/chunkserver/masterconn.cc
+++ b/src/chunkserver/masterconn.cc
@@ -309,7 +309,7 @@ public:
 		}
 	}
 
-	void servePoll(const std::vector<pollfd> &pdesc) {
+	void handlePollErrors(const std::vector<pollfd> &pdesc) {
 		// Check if the socket has been closed or has an error.
 		if (pDescPos_ >= 0 && (pdesc[pDescPos_].revents & (POLLHUP | POLLERR))) {
 			if (mode_ == ConnectionMode::CONNECTING) {
@@ -318,7 +318,9 @@ public:
 				mode_ = ConnectionMode::KILL;
 			}
 		}
+	}
 
+	void servePoll(const std::vector<pollfd> &pdesc) {
 		if (mode_ == ConnectionMode::CONNECTING) {
 			// Check if the connection has been established.
 			if (socketFD_ >= 0 && pDescPos_ >= 0 && (pdesc[pDescPos_].revents & POLLOUT)) {
@@ -422,7 +424,6 @@ public:
 				return;
 			}
 
-			// stats_bytesout += bytesWritten;
 			bytesOut_ += bytesWritten;
 			pack.bytesSent += bytesWritten;
 
@@ -486,9 +487,8 @@ public:
 		auto *outputPacket = new OutputPacket;
 		cstoma::createChunk::serialize(outputPacket->packet, chunkId, chunkType, SAUNAFS_STATUS_OK);
 		if (jobPool_) {
-			job_create(
-			    *jobPool_, [this](uint8_t status, void *packet) { sauJobFinished(status, packet); },
-			    outputPacket, chunkId, chunkVersion, chunkType);
+			job_create(*jobPool_, sauJobFinished(this), outputPacket, chunkId, chunkVersion,
+			           chunkType);
 		} else {
 			safs::log_err("masterconn_create: jobPool is null.");
 			delete outputPacket;
@@ -504,9 +504,8 @@ public:
 		auto *outputPacket = new OutputPacket;
 		cstoma::deleteChunk::serialize(outputPacket->packet, chunkId, chunkType, 0);
 		if (jobPool_) {
-			job_delete(
-			    *jobPool_, [this](uint8_t status, void *packet) { sauJobFinished(status, packet); },
-			    outputPacket, chunkId, chunkVersion, chunkType);
+			job_delete(*jobPool_, sauJobFinished(this), outputPacket, chunkId, chunkVersion,
+			           chunkType);
 		} else {
 			safs::log_err("masterconn_delete: jobPool is null.");
 			delete outputPacket;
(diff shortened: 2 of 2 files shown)


rolysr (approved):
LGTM.

ralcolea (approved):
Great work @lgsilva3087 :+1: :fire: :muscle:

src/chunkserver/init.h
```diff
@@ -39,11 +39,11 @@ inline const std::vector<RunTab> runTabs = {
     RunTab{loadPlugins, "plugin manager"}, RunTab{hddInit, "hdd space manager"},
     // Has to be before "masterconn"
     RunTab{mainNetworkThreadInit, "main server module"},
+    RunTab{masterconn_init_threads, "master connection module - threads"},
     RunTab{masterconn_init, "master connection module"},
```
antuan96314, line 43: Consider merging this two init functions for simplicity.

src/chunkserver/masterconn.cc
```diff
@@ -78,619 +91,731 @@ struct MasterConn {
...
+	eptr->providePollDescriptors(pdesc);
 
-	if (eptr->mode == ConnectionMode::FREE || eptr->sock < 0) { return; }
+	// Add the descriptor for listening for background jobs finishing.
+	gJobFDpDescPos = -1;
 
-	if (eptr->mode == ConnectionMode::CONNECTED) {
+	if (eptr->mode() == ConnectionMode::CONNECTED) {
 		pdesc.emplace_back(gJobFD, POLLIN, 0);
 		gJobFDpDescPos = static_cast<int32_t>(pdesc.size() - 1);
```
dmga44: Nice job here, though I want to point at a couple of things:
- the order of the two fd's to poll at are switched (it was previously the bgJobsPool listener and then the socket from master, and now it is the other way around) but this should not make any difference because the positions of each one are clear.
- as the bgJobsPool was kept global, then it means if there were several master connections, all of them were going there. Going a bit forward in that direction, it would be not easy to determine which master sent the job in the first place (in order to send the reply).

src/chunkserver/masterconn.cc
```diff
@@ -700,86 +825,46 @@ void masterconn_send_status() {
...
-		if (eptr->mode == ConnectionMode::CONNECTING) {
-			masterconn_connecttest(eptr);
-		} else {
-			eptr->mode = ConnectionMode::KILL;
-		}
+	
+	// Check if there are any background jobs to process.
+	if ((eptr->mode() == ConnectionMode::CONNECTED) && gJobFDpDescPos >= 0 &&
+	    (pdesc[gJobFDpDescPos].revents & POLLIN)) {
+		gJobPool->processCompletedJobs();
```
dmga44, line 839: Please consider the following:
- in the previous implementation this piece of code was called after checking the status of the polling on the master connection fd, so it was clear that state of the connection was CONNECTED and this operation was safe (or safer).
- the proposed implementation does not has this kind of check, so the connection might died and needs to be tested (and reconnected if necessary). The issue comes when trying to reply to the master some of the tasks it sent, which could fail in this scenario.

src/chunkserver/masterconn.cc
dmga44: These are just things to think about, are not worrying for now.

src/chunkserver/masterconn.cc
```diff
@@ -78,619 +91,731 @@ struct MasterConn {
...
+				if (errno != EAGAIN) {
+					safs_silent_errlog(LOG_NOTICE, "write to Master error");
+					mode_ = ConnectionMode::KILL;
+				}
+				return;
+			}
 
-void masterconn_gotpacket(MasterConn *eptr, PacketHeader header, const MessageBuffer& message) try {
-	switch (header.type) {
+			// stats_bytesout += bytesWritten;
```
dmga44: This seems to be some old code commented, consider removing it.

src/chunkserver/masterconn.cc
```diff
@@ -78,619 +91,731 @@ struct MasterConn {
...
+		ChunkPartType chunkType = slice_traits::standard::ChunkPartType();
+		uint32_t chunkVersion;
+
+		matocs::createChunk::deserialize(data, chunkId, chunkType, chunkVersion);
+		auto *outputPacket = new OutputPacket;
+		cstoma::createChunk::serialize(outputPacket->packet, chunkId, chunkType, SAUNAFS_STATUS_OK);
+		if (jobPool_) {
+			job_create(
+			    *jobPool_, [this](uint8_t status, void *packet) { sauJobFinished(status, packet); },
+			    outputPacket, chunkId, chunkVersion, chunkType);
```
dmga44: Consider the following:
- define the following funtion in the .cc (and .h for the purpose of the change):
```cpp
std::function<void(uint8_t status, void *packet)> sauJobFinished(void *eptr) {
	auto *masterConn = static_cast<MasterConn *>(eptr);
	return [masterConn](uint8_t status, void *packet) {
		masterConn->sauJobFinished(status, packet);
	};
}
```
- substitute each of the lambdas in the job calls by a call to that function, for instance:
```cpp
			job_create(*jobPool_, ::sauJobFinished(this), outputPacket, chunkId, chunkVersion,
			           chunkType);
```

This is just to make the code look better IMO, left to your consideration.

dmga44 (commented):
Overall, great job @lgsilva3087 🔥 !!

Please check out my comments, I think one of them is important.

As an extra minor suggestion, consider updating the logging to `safs::log_*` equivalent in the lines touched.

src/chunkserver/masterconn.cc
lgsilva3087: Order restored.

src/chunkserver/masterconn.cc
lgsilva3087: Done.

src/chunkserver/masterconn.cc
lgsilva3087: Removed.

src/chunkserver/masterconn.cc
lgsilva3087: Done.

src/chunkserver/init.h
lgsilva3087: I like the idea, but those functions will have more substance if we decide to add the second JobPool (for replications) and multiple MasterConn(s). I will keep them for now, with strong willing to unify them if possible.

dmga44 (approved):

src/chunkserver/masterconn.cc
```diff
@@ -78,620 +89,728 @@ struct MasterConn {
...
+		uint32_t chunkVersion;
+
+		matocs::createChunk::deserialize(data, chunkId, chunkType, chunkVersion);
+		auto *outputPacket = new OutputPacket;
+		cstoma::createChunk::serialize(outputPacket->packet, chunkId, chunkType, SAUNAFS_STATUS_OK);
+		if (jobPool_) {
+			job_create(*jobPool_, sauJobFinished(this), outputPacket, chunkId, chunkVersion,
+			           chunkType);
+		} else {
+			safs::log_err("masterconn_create: jobPool is null.");
```
ralcolea: Consider updating the logs with the new function names for the functions involved in chunk operations.
For example, replace `"masterconn_create"` with `"createChunk"`.

ralcolea (approved):
Great job @lgsilva3087!!! :+1: :fire: :muscle:

src/chunkserver/masterconn.cc
lgsilva3087: Done.

wg1k (approved):

PR #433, opened 2025-06-05
feat(tools)!: Remove default timeout from some saunafs operations

uristdwarf:
The timeout was always default 30 seconds. But this didn't make any
sense, because the master would continue performing the operation
regardless. Worse, some functions were cleanly stopping the operations
on master if SIGINT or SIGTERM was send, but this timeout would not and
simply close the connection once timeout reached. Old behaviour could be
mostly restored with `timeout`.

* Add some logging for the task ID

BREAKING CHANGE: Existing scripts that rely on this -l behaviour will
break.

Signed-off-by: Urmas Rist <urmas@leil.io>

### DIFF
--- a/src/tools/file_info.cc
+++ b/src/tools/file_info.cc
@@ -29,14 +29,12 @@
 #include "tools/tools_commands.h"
 #include "tools/tools_common_functions.h"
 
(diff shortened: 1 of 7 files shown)


dmga44 (approved):
LGTM.

ralcolea (approved):
LGTM :+1:

PR #436, opened 2025-06-09
feat: Speedup replication/rebalance

dmga44:
This PR includes several changes that should speedup the replication and rebalance operations in our system. Further details in each commit.

### DIFF
--- a/src/chunkserver/chunk_replicator.cc
+++ b/src/chunkserver/chunk_replicator.cc
@@ -149,7 +149,11 @@ void ChunkReplicator::replicate(ChunkFileCreator& fileCreator,
 	SliceRecoveryPlanner planner;
 	ReadPlanExecutor::ChunkTypeLocations locations;
 	SliceRecoveryPlanner::PartsContainer available_parts;
+#ifdef SAUNAFS_HAVE_THREAD_LOCAL
+	static thread_local std::vector<uint8_t, AlignedAllocator<uint8_t, disk::kIoBlockSize>> buffer;
+#else
 	std::vector<uint8_t, AlignedAllocator<uint8_t, disk::kIoBlockSize>> buffer;
+#endif
 
 	for (const auto& source : sources) {
 		available_parts.push_back(source.chunk_type);
--- a/src/chunkserver/chunk_replicator.cc
+++ b/src/chunkserver/chunk_replicator.cc
@@ -141,7 +141,7 @@ void ChunkReplicator::replicate(ChunkFileCreator& fileCreator,
 		const std::vector<ChunkTypeWithAddress>& sources) {
 	// Get number of blocks to replicate
 	int blocks = getChunkBlocks(fileCreator.chunkId(), fileCreator.chunkVersion(), sources);
-	int batchSize = 50;
+	int batchSize = SFSBLOCKSINCHUNK;
 	int data_part_count = slice_traits::getNumberOfDataParts(fileCreator.chunkType());
 	blocks = slice_traits::getNumberOfBlocks(fileCreator.chunkType(), blocks);
 	batchSize = data_part_count * ((batchSize + data_part_count - 1) / data_part_count);
--- a/src/common/slice_read_planner.cc
+++ b/src/common/slice_read_planner.cc
@@ -165,10 +165,13 @@ bool SliceReadPlanner::shouldReadPartsRequiredForRecovery() const {
  * are queued for first wave and additional parts are added to consecutive waves.
  * \param first_block first block to be read
  * \param block_count number of blocks to be read
+ * \param addExtraPartsWhenRequiredPartsAvailable whether to add extra parts when required parts are
+ * available
  * \return read plan ready to be executed
  */
-std::unique_ptr<ReadPlan> SliceReadPlanner::buildPlanFor(uint32_t first_block,
-		uint32_t block_count) {
+std::unique_ptr<ReadPlan> SliceReadPlanner::buildPlanFor(
+    uint32_t first_block, uint32_t block_count,
+    bool addExtraPartsWhenRequiredPartsAvailable) {
 	std::unique_ptr<SliceReadPlan> plan = getPlan();
 	plan->buffer_part_size = block_count * SFSBLOCKSIZE;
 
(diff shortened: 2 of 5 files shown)


src/master/chunks.cc
```diff
@@ -82,6 +83,9 @@
 #define MINCHUNKSLOOPCPU    10
 #define MAXCHUNKSLOOPCPU    90
 
+std::random_device rd;
+std::default_random_engine rng(rd());
```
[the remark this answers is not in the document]
dmga44: It is obly being used from one thread.

src/chunkserver/chunk_replicator.cc
```diff
@@ -141,15 +141,19 @@ void ChunkReplicator::replicate(ChunkFileCreator& fileCreator,
...
 	batchSize = data_part_count * ((batchSize + data_part_count - 1) / data_part_count);
 
 	SliceRecoveryPlanner planner;
 	ReadPlanExecutor::ChunkTypeLocations locations;
 	SliceRecoveryPlanner::PartsContainer available_parts;
+#ifdef SAUNAFS_HAVE_THREAD_LOCAL
+	static thread_local std::vector<uint8_t, AlignedAllocator<uint8_t, disk::kIoBlockSize>> buffer;
+#else
 	std::vector<uint8_t, AlignedAllocator<uint8_t, disk::kIoBlockSize>> buffer;
+#endif
```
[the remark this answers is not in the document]
dmga44: Thank you.

src/common/slice_read_planner.cc
```diff
@@ -167,8 +167,9 @@ bool SliceReadPlanner::shouldReadPartsRequiredForRecovery() const {
  * \param block_count number of blocks to be read
  * \return read plan ready to be executed
  */
-std::unique_ptr<ReadPlan> SliceReadPlanner::buildPlanFor(uint32_t first_block,
-		uint32_t block_count) {
+std::unique_ptr<ReadPlan> SliceReadPlanner::buildPlanFor(
+    uint32_t first_block, uint32_t block_count,
+    bool addExtraPartsWhenRequiredPartsAvailable) {
```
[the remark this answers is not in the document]
dmga44: Indeed.

src/chunkserver/chunk_replicator.cc
```diff
@@ -141,15 +141,19 @@ void ChunkReplicator::replicate(ChunkFileCreator& fileCreator,
 		const std::vector<ChunkTypeWithAddress>& sources) {
 	// Get number of blocks to replicate
 	int blocks = getChunkBlocks(fileCreator.chunkId(), fileCreator.chunkVersion(), sources);
-	int batchSize = 50;
+	int batchSize = SFSBLOCKSINCHUNK;
```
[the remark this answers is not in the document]
dmga44: It is because the chunk size is always `SFSBLOCKSINCHUNK`.

src/master/chunks.cc
```diff
@@ -2250,7 +2254,12 @@ bool ChunkWorker::rebalanceChunkParts(Chunk *c, ChunkCopiesCalculator &calc, boo
 	// Consider each copy to be moved to a server with disk usage much less than actual.
 	// There are at least two servers with a disk usage difference grater than
 	// gAcceptableDifference, so it's worth checking.
-	for (const auto &part : c->parts) {
+	int partsSize = c->parts.size();
+	std::vector<int> part_indices(partsSize);
+	std::iota(part_indices.begin(), part_indices.end(), 0);
+	std::shuffle(part_indices.begin(), part_indices.end(), rng);
+	for (int i = 0; i < partsSize; ++i) {
+		const auto &part = c->parts[part_indices[i]];
```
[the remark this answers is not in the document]
dmga44: That was the idea.

src/master/chunks.cc
lgsilva3087: Consider using the global `kRandomEngine`, defined in `common/random.h` and initialized in `master/init.h`.

src/master/chunks.cc
```diff
@@ -2250,7 +2254,12 @@ bool ChunkWorker::rebalanceChunkParts(Chunk *c, ChunkCopiesCalculator &calc, boo
 	// Consider each copy to be moved to a server with disk usage much less than actual.
 	// There are at least two servers with a disk usage difference grater than
 	// gAcceptableDifference, so it's worth checking.
-	for (const auto &part : c->parts) {
+	int partsSize = c->parts.size();
+	std::vector<int> part_indices(partsSize);
+	std::iota(part_indices.begin(), part_indices.end(), 0);
+	std::shuffle(part_indices.begin(), part_indices.end(), rng);
```
lgsilva3087: Consider:
`std::shuffle(part_indices.begin(), part_indices.end(), kRandomEngine);`

lgsilva3087: Consider to comment why is it useful to randomize the parts.
Also to separate it in a different block with its own comment, like:

// Avoid same part types landing in the same chunkservers by randomizing...
Randomization code here

// Then the old comment
// Consider each copy to be moved to a server with disk usage much less than actual.
// There are at least two servers with a disk usage difference grater than
// gAcceptableDifference, so it's worth checking.
And the for loop

src/common/slice_read_planner.cc
```diff
@@ -167,8 +167,9 @@ bool SliceReadPlanner::shouldReadPartsRequiredForRecovery() const {
  * \param block_count number of blocks to be read
  * \return read plan ready to be executed
```
lgsilva3087: Add the new parameter to the documentation.

src/chunkserver/chunk_replicator.cc
```diff
@@ -141,7 +141,7 @@ void ChunkReplicator::replicate(ChunkFileCreator& fileCreator,
...
```
lgsilva3087, line 144: Replications are done in the workers assigned to the masterconn's JobPool, by default 10 workers. Please be aware that, for our usual 64 MiB chunks, it could hold in memory 64 MiB * 10 workers = 640 MiB per chunkserver, without releasing it (thread local storage)?
If the number of workers or the chunk size is increased, the memory will grow accordingly.

lgsilva3087 (commented):
Please see my comments.

src/chunkserver/chunk_replicator.cc
dmga44: That is only true if some of the replicated parts are goal1.

Please note that the replicate only replicates a chunk part, and saves space only for that part which is the objective of one of the commits in this PR.

And, yes, we should be aware that there could be an increase in memory similar to what you described.

src/common/slice_read_planner.cc
dmga44: Done.

src/master/chunks.cc
dmga44: Applied.

dmga44: Applied.

src/master/chunks.cc
dmga44: Done, removed this lines.

lgsilva3087 (approved):
LGTM

ralcolea (approved):
Great work @dmga44!!! :+1: :fire: :muscle:

PR #437, opened 2025-06-09
feat: Readcache buffers pool

dmga44:
This PR targets reducing the allocation and deallocation of memory
when actively reading. Should improve performance on heavy read
workloads.

Changes:
- implement `ReadCacheEntriesPool` to hold the read cache entries for a
while (1s by default, configurable via parameters) after supposedly
being deleted from cache.
- when requesting a new entry in the ReadCache we'd look at the global
`ReadCacheEntriesPool` singleton and take one from there if available.
- a new thread for periodically cleaning the `ReadCacheEntriesPool` is
created, it checks every 0.1 seconds.
- the operations on the `ReadCacheEntriesPool` are thread safe. A mutex
ensures protection.

As side changes:
- most of the readdata_cache.h function implementations were moved to
readdata_cache.cc altogether with the new `ReadCacheEntriesPool`
functions.
- various clang tidy warnings were addressed like removing unused
includes.

The client read algorithm params for LongSystemTests suite has been
decreased to save some memory.


src/mount/readdata_cache.h
```diff
@@ -338,191 +204,73 @@ class ReadCache {
...
+};
 
-	void clear() {
-		auto it = entries_.begin();
-		while (it != entries_.end()) {
-			it = erase(it);
-		}
+class ReadCacheEntriesPool {
+public:
+	ReadCacheEntriesPool() : entriesMap_(), mutex_() {
```
lgsilva3087: These initializers are redundant, please consider to remove the explicit initialization.

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,446 @@
...
+			} else {
+				++it;
+			}
+		}
+		lock.unlock();
+
+		// Delete buffers outside the lock to avoid contention overhead.
+		for (auto &entry : entriesToDelete) { delete entry; }
+	}
+}
```
lgsilva3087: Add empty line at the end of the file.

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,446 @@
...
+
+void ReadCacheEntriesPool::putEntry(ReadCache::Entry *entry) {
+	assert(entry != nullptr);
+	std::unique_lock<std::mutex> lock(mutex_);
+	entriesMap_[entry->buffer.size()].emplace_front(entry);
+}
+
+void ReadCacheEntriesPool::cleanerThreadFunc_(std::stop_token &stopToken) {
+	pthread_setname_np(pthread_self(), "ReadCacheEntriesPool's cleanerThread");
```
lgsilva3087: I think thread names can be up to 16 characters to be shown, please check it.

src/mount/readdata_cache.h
```diff
@@ -338,191 +204,73 @@ class ReadCache {
...
+	uint32_t expiration_time_;
+	static constexpr size_t kCollectGarbageCountDefault_ = 100000;
+};
 
-	void clear() {
-		auto it = entries_.begin();
-		while (it != entries_.end()) {
-			it = erase(it);
-		}
+class ReadCacheEntriesPool {
```
lgsilva3087, line 236: As this class manage its own resources and looks like is intended to be used as a singleton. Consider to remove some non-expected constructors/operators to explicitely avoid misuse, like:

```cpp
ReadCacheEntriesPool(const ReadCacheEntriesPool&) = delete;
ReadCacheEntriesPool& operator=(const ReadCacheEntriesPool&) = delete;
ReadCacheEntriesPool(ReadCacheEntriesPool&&) = delete;
ReadCacheEntriesPool& operator=(ReadCacheEntriesPool&&) = delete;
```

src/mount/readdata_cache.h
```diff
@@ -338,191 +204,73 @@ class ReadCache {
...
-			assert(e->refcount == 0);
-			std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
-			decreaseUsedReadCacheMemory(e->buffer.size());
-			usedMemoryLock.unlock();
-			delete e;
-		}
-		return ret;
-	}
+		BufferEntry(BufferEntry &&other) noexcept
+		    : entry_(other.entry_), timer_(other.timer_) {}
```
lgsilva3087: Consider:
```cpp
BufferEntry(BufferEntry &&other) noexcept
    : entry_(other.entry_), timer_(std::move(other.timer_)) {
    other.entry_ = nullptr;
}
```
The moved object should be put into a valid but unspecified stated after move. As entry_ is a raw pointer, the common practice is to set it to nullptr.

src/mount/readdata_cache.h
```diff
@@ -338,191 +204,73 @@ class ReadCache {
...
-	}
+	void putEntry(ReadCache::Entry *entry);
 
-	Entry *find(uint64_t offset) {
-		auto it = entries_.upper_bound(offset, Entry::OffsetComp());
-		if (it != entries_.begin()) {
-			--it;
-		}
+private:
+	void cleanerThreadFunc_(std::stop_token &stopToken);
```
lgsilva3087: Consider to pass the stop token by value, which matches how std::jthread passes the token from the lambda to the thread function.
`void cleanerThreadFunc_(std::stop_token stopToken);`

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,446 @@
...
+			return MISSING_OFFSET_PTR;
+		}
+		return std::addressof(*it);
+	}
+	return MISSING_OFFSET_PTR;
+}
+
+ReadCache::EntrySet::iterator ReadCache::insert(EntrySet::iterator it, Offset offset, Size size) {
+	it = clearCollisions(it, offset + size);
+	// Entry *e = new Entry(offset, size);
```
lgsilva3087: Remove the comment for the final version.

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,446 @@
...
+	auto ret = entries_.erase(it);
+	lru_.erase(lru_.iterator_to(*e));
+	if (e->refcount > 0) {
+		reserved_entries_.push_back(*e);
+	} else {
+		assert(e->refcount == 0);
+		std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
+		decreaseUsedReadCacheMemory(e->buffer.size());
+		usedMemoryLock.unlock();
+		// delete e;
```
lgsilva3087: Remove the comment for the final version.

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,446 @@
...
+	buffers.pop_back();
+	lock.unlock();
+
+	// Reinitialize the entry
+	entry->offset = offset;
+	entry->buffer.clear();
+	entry->timer = Timer();
+	entry->refcount = 0;
+	entry->requested_size = size;
+	entry->done = false;
```
lgsilva3087, line 422: Consider a function, member of ReadCache::Entry to reset the values, taking as arguments offset and size. Something like:
`entry->resetValues(offset, size);`
Consider it only if this will be used in other places.

wg1k (approved):

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,444 @@
...
+	auto it = entries_.begin();
+	while (it != entries_.end()) { it = erase(it); }
+}
+
+void ReadCache::selective_clear(uint32_t chunkIndex) {
+	// The least starting offset such that the end of the entry could overlap with the chunk
+	uint64_t leastOverlappingOffset = 0;
+	if (chunkIndex > 0) { leastOverlappingOffset = (chunkIndex - 1) * 1024 * 1024 * 64; }
+	uint64_t chunkStart = chunkIndex * 1024 * 1024 * 64;
+	uint64_t chunkEnd = chunkStart + 1024 * 1024 * 64;
```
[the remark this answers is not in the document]
lgsilva3087: Please check the comment from Gemini

src/mount/readdata_cache.cc
```diff
@@ -0,0 +1,446 @@
...
+	lru_.erase(lru_.iterator_to(*e));
+	if (e->refcount > 0) {
+		reserved_entries_.push_back(*e);
+	} else {
+		assert(e->refcount == 0);
+		std::unique_lock usedMemoryLock(gReadCacheMemoryMutex);
+		decreaseUsedReadCacheMemory(e->buffer.size());
+		usedMemoryLock.unlock();
+		// delete e;
+		gReadCacheEntriesPool->putEntry(e);
```
lgsilva3087: Consider to check gReadCacheMemoryAlmostExceeded before this call. If there is too much pressure in the memory, maybe we can just delete the entry.

lgsilva3087 (commented):
Please see my comments.

src/mount/readdata_cache.cc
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/readdata.cc
```diff
@@ -607,6 +600,7 @@ void read_data_init(uint32_t retries,
 
 	readDataTerminate = false;
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/readdata_cache.h
dmga44: Applied.

src/mount/readdata_cache.cc
dmga44: Applied.

src/mount/readdata_cache.cc
dmga44: Changed to "RCEP cleaner".

src/mount/readdata_cache.h
dmga44: Applied.

src/mount/readdata_cache.h
dmga44: Applied.

src/mount/readdata_cache.h
dmga44: Applied.

src/mount/readdata_cache.cc
dmga44: Removed.

src/mount/readdata_cache.cc
dmga44: Done.

src/mount/readdata_cache.cc
dmga44: It is not used anywhere else, so I did not add it.

src/mount/readdata_cache.cc
dmga44: The `putEntry` call was substituted by the following in `erase` and `clearReserved` functions:
```cpp
// If we are almost out of memory, delete the entry instead of putting it back to the pool
if (gReadCacheMemoryAlmostExceeded) {
	delete e;
} else {
	gReadCacheEntriesPool->putEntry(e);
}
```

dmga44: As you may notice, the memory held in the `ReadCacheEntriesPool` is not being computed in the `gUsedReadCacheMemory`, this is mainly because of two reasons:
- it is actually not in cache, as it does not hold any valuable data, just the entries.
- it would be a complete mess to keep track of it (mainly because the resizes of the buffers in the read algorithm won't be actually allocating data)

lgsilva3087 (approved):
LGTM

PR #444, opened 2025-06-12
feat(chunkserver): Implement listeners for JobPool

dmga44:
This PR targets changing the current behavior of the JobPool class,
in the way that it can serve different threads at the same time.
Current implementation does not allow a single JobPool instance to
serve different threads because it would mix up the wake up calls and
it would cause dangerous inconsistencies.

The proposed implementation partitions the JobPool inner structures for
each of the sources of jobs (which could be master or client network
workers) and creates specific channels of comunication to ensure that
only the source of a job is awaken.

This PR also includes changes in the JobPool's unittests to check the new
behavior.

The idea is to squash these commits.

### DIFF
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -47,17 +47,24 @@
 
 constexpr auto kInvalidJob = nullptr;
 
-JobPool::JobPool(const std::string &name, uint8_t workers, uint32_t maxJobs, int nrListeners,
-                 std::vector<int> &wakeupDesc)
+JobPool::JobPool(const std::string &name, uint8_t workers, uint32_t maxJobs, uint32_t nrListeners,
+                 std::vector<int> &wakeupFDs)
     // EFD_NONBLOCK to prevent blocking reads/writes
     : listenerInfos_(nrListeners), name_(name), workers(workers) {
-	for (int i = 0; i < nrListeners; ++i) {
+	if (wakeupFDs.size() != nrListeners) {
+		safs::log_warn(
+		    "JobPool: wakeupFDs size {} does not match nrListeners {}, resizing to match.",
+		    wakeupFDs.size(), nrListeners);
+		wakeupFDs.resize(nrListeners);
+	}
+
+	for (uint32_t i = 0; i < nrListeners; ++i) {
 		listenerInfos_[i].notifierFD = ::eventfd(0, EFD_NONBLOCK);
 		if (listenerInfos_[i].notifierFD < 0) {
 			throw std::runtime_error("JobPool: eventfd() failed for listener" +
 			                         std::to_string(i) + ": " + std::string(strerror(errno)));
 		}
-		wakeupDesc[i] = listenerInfos_[i].notifierFD;
+		wakeupFDs[i] = listenerInfos_[i].notifierFD;
 
 		listenerInfos_[i].statusQueue = std::make_unique<ProducerConsumerQueue>();
 		listenerInfos_[i].nextJobId = 1;
--- a/src/chunkserver/bgjobs.cc
+++ b/src/chunkserver/bgjobs.cc
@@ -51,6 +51,8 @@ JobPool::JobPool(const std::string &name, uint8_t workers, uint32_t maxJobs, uin
                  std::vector<int> &wakeupFDs)
     // EFD_NONBLOCK to prevent blocking reads/writes
     : listenerInfos_(nrListeners), name_(name), workers(workers) {
+	nrListeners = std::max(nrListeners, 1u);  // Ensure at least one listener
+
 	if (wakeupFDs.size() != nrListeners) {
 		safs::log_warn(
 		    "JobPool: wakeupFDs size {} does not match nrListeners {}, resizing to match.",
@@ -91,17 +93,22 @@ JobPool::~JobPool() {
 	}
 
 	jobsQueue.reset();
-	for (size_t i = 0; i < listenerInfos_.size(); ++i) { listenerInfos_[i].statusQueue.reset(); }
+	for (auto &listenerInfo : listenerInfos_) { listenerInfo.statusQueue.reset(); }
 
 	workerThreads.clear();
 
-	for (auto &listenerInfo : listenerInfos_) {
-		close(listenerInfo.notifierFD);
-	}
+	for (auto &listenerInfo : listenerInfos_) { close(listenerInfo.notifierFD); }
 }
 
 uint32_t JobPool::addJob(ChunkOperation operation, JobCallback callback, void *extra,
                          ProcessJobCallback processJob, uint32_t listenerId) {
+	// Check if the listenerId is valid
+	if (listenerId >= listenerInfos_.size()) {
+		safs::log_warn("JobPool: addJob: Invalid listenerId {} for operation {}, resetting to 0",
+		               listenerId, static_cast<int>(operation));
+		listenerId = 0;  // Reset to the first listener
+	}
+
 	auto &listenerInfo = listenerInfos_[listenerId];
 	std::unique_lock lock(listenerInfo.jobsMutex);
 	uint32_t jobId = listenerInfo.nextJobId++;
@@ -124,6 +131,13 @@ uint32_t JobPool::getJobCount() const {
 }
 
 void JobPool::disableAndChangeCallbackAll(const JobCallback &callback, uint32_t listenerId) {
+	// Check if the listenerId is valid
+	if (listenerId >= listenerInfos_.size()) {
+		safs::log_warn("JobPool: disableAndChangeCallbackAll: Invalid listenerId {}, returning",
+		               listenerId);
+		return;
+	}
+
 	auto &listenerInfo = listenerInfos_[listenerId];
 	std::lock_guard jobsLockGuard(listenerInfo.jobsMutex);
 	for (auto &[jobId, job] : listenerInfo.jobHash) {
@@ -133,6 +147,13 @@ void JobPool::disableAndChangeCallbackAll(const JobCallback &callback, uint32_t 
 }
 
 void JobPool::disableJob(uint32_t jobId, uint32_t listenerId) {
+	// Check if the listenerId is valid
+	if (listenerId >= listenerInfos_.size()) {
+		safs::log_warn("JobPool: disableJob: Invalid listenerId {} for jobId {}, returning",
+		               listenerId, jobId);
+		return;
+	}
+
 	auto &listenerInfo = listenerInfos_[listenerId];
 	std::unique_lock jobsUniqueLock(listenerInfo.jobsMutex, std::defer_lock);
 	auto jobIterator = listenerInfo.jobHash.find(jobId);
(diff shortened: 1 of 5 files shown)


src/chunkserver/bgjobs.cc
```diff
@@ -346,80 +379,86 @@ uint32_t job_replicate(JobPool &jobPool, JobPool::JobCallback callback, void *ex
...
 	if (newChunkVersion > 0) {
 		JobPool::ProcessJobCallback processJob = [=]() -> uint8_t {
 			return hddDuplicateTruncate(chunkId, chunkVersion, newChunkVersion, chunkType,
 			                            chunkIdCopy, chunkVersionCopy, length);
 		};
 		return jobPool.addJob(JobPool::ChunkOperation::DuplicateTruncate, callback, extra,
-		                      processJob);
+		                      processJob, listenerId);
 	}
 	return job_invalid(jobPool, callback, extra);
```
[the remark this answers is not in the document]
dmga44: Indeed, thank you for pointing that out.

src/chunkserver/bgjobs.cc
```diff
@@ -346,80 +379,86 @@ uint32_t job_replicate(JobPool &jobPool, JobPool::JobCallback callback, void *ex
...
+                     uint32_t newChunkVersion, uint32_t listenerId) {
 	if (newChunkVersion > 0) {
 		JobPool::ProcessJobCallback processJob = [=]() -> uint8_t {
 			return hddInternalUpdateVersion(chunkId, chunkVersion, newChunkVersion, chunkType);
 		};
-		return jobPool.addJob(JobPool::ChunkOperation::ChangeVersion, callback, extra, processJob);
+		return jobPool.addJob(JobPool::ChunkOperation::ChangeVersion, callback, extra, processJob,
+		                      listenerId);
 	}
 	return job_invalid(jobPool, callback, extra);
```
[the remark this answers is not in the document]
dmga44: Indeed, thank you for pointing that out.

src/chunkserver/masterconn.cc
```diff
@@ -950,8 +950,12 @@ int masterconn_init_threads(void) {
...
 
 	try {
-		gJobPool =
-		    std::make_shared<JobPool>("ma", gNumberOfWorkers, kMaxBackgroundJobsCount, &gJobFD);
+		// Create the JobPool instance with the specified number of workers, it would be serving
+		// only this master network thread, thus the number of listeners is 1.
+		std::vector<int> bgJobPoolFDs(1);
+		gJobPool = std::make_unique<JobPool>("ma", gNumberOfWorkers, kMaxBackgroundJobsCount, 1,
+		                                     bgJobPoolFDs);
+		gJobFD = bgJobPoolFDs[0];
```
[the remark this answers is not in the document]
dmga44: You're right, thank you.

src/chunkserver/bgjobs_unittest.cc
```diff
@@ -62,45 +68,71 @@ class JobPoolTest : public ::testing::Test {
...
-	std::atomic<int> counter;
-	std::atomic<int> counter2;
-	struct pollfd wakeupDescPollFd;
-	std::thread servePollThread;
+	std::vector<int> wakeupDescVec;
+	std::vector<std::thread> servePollThreads;
 	bool terminate;
 	std::mutex mutex_;
+	const static uint32_t kNrListeners = 4;
+	const static uint32_t kNrOperationTypes = 10;
```
lgsilva3087: These lines break building under my Clang, because the static variables are not defined outside the class. As they are compile time constants, please use:
```cpp
static constexpr uint32_t kNrListeners = 4;
static constexpr uint32_t kNrOperationTypes = 10;
```

src/chunkserver/bgjobs.cc
```diff
@@ -213,32 +241,30 @@ void JobPool::workerThread(const std::string &poolName, uint8_t workerId) {
...
-void JobPool::sendStatus(uint32_t jobId, uint8_t status) {
-	std::lock_guard statusLock(statusMutex_);
-
-	if (statusQueue->isEmpty()) {
-		static constexpr eventfd_t dummyValue = 1;  // Dummy value just to wake up the eventfd
-		eassert(::eventfd_write(notifierFD_, dummyValue) == 0 &&
-		        "JobPool: SendStatus: Failed to write to eventfd");
+void JobPool::sendStatus(uint32_t jobId, uint8_t status, uint32_t listenerId) {
+	auto &listenerInfo = listenerInfos_[listenerId];
+	std::unique_lock pipeLockGuard(listenerInfo.notifierMutex);
```
lgsilva3087: Please notice that there is no pipe anymore (`pipeLockGuard`).

lgsilva3087 (commented):
So far looks great. Please see my comments.

src/chunkserver/bgjobs_unittest.cc
dmga44: Applied.

src/chunkserver/bgjobs.cc
dmga44: Indeed, I've renamed it to `statusLock`.

lgsilva3087 (approved):
Great.

src/chunkserver/bgjobs.h
```diff
@@ -166,20 +175,23 @@ class JobPool {
...
 	bool receiveStatus(uint32_t &jobId, uint8_t &status, uint32_t listenerId = 0);
 
+	/// @brief Structure to hold information about a listener.
 	struct ListenerInfo {
-		int notifierFD;           /// File descriptor for notifications.
-		std::mutex notifierMutex;  /// Mutex for pipe operations.
-		std::mutex jobsMutex;      /// Mutex for job operations.
-		std::unique_ptr<ProducerConsumerQueue> statusQueue;          /// Queue for job statuses.
+		int notifierFD;                                      /// File descriptor for notifications.
+		std::mutex notifierMutex;                            /// Mutex for pipe operations.
```
ralcolea: I think `notifierMutex` is not related with pipe operations.
Consider updating the documentation.

src/chunkserver/bgjobs.cc
```diff
@@ -73,28 +86,35 @@ JobPool::~JobPool() {
...
 	}
 
-	if (!statusQueue->isEmpty()) { processCompletedJobs(); }
+	for (size_t i = 0; i < listenerInfos_.size(); ++i) {
+		if (!listenerInfos_[i].statusQueue->isEmpty()) { processCompletedJobs(i); }
+	}
 
 	jobsQueue.reset();
-	statusQueue.reset();
+	for (size_t i = 0; i < listenerInfos_.size(); ++i) { listenerInfos_[i].statusQueue.reset(); }
```
ralcolea: Consider using a range-based for loop as well as at the end of the function for closing the file descriptors.

src/chunkserver/bgjobs.cc
```diff
@@ -213,32 +241,32 @@ void JobPool::workerThread(const std::string &poolName, uint8_t workerId) {
...
+	std::lock_guard statusLock(listenerInfo.notifierMutex);
 
-	if (statusQueue->isEmpty()) {
-		static constexpr eventfd_t dummyValue = 1;  // Dummy value just to wake up the eventfd
-		eassert(::eventfd_write(notifierFD_, dummyValue) == 0 &&
-		        "JobPool: SendStatus: Failed to write to eventfd");
+	if (listenerInfo.statusQueue->isEmpty()) {
+		static constexpr eventfd_t dummyValue = 1;  // Dummy value to wake up the eventfd
+		eassert(::eventfd_write(listenerInfo.notifierFD, dummyValue) == 0 &&
+		        "JobPool: SendStatus: Failed to write status to pipe");
```
ralcolea: I don't think this is a pipe but `eventfd` mechanism. It works similar to pipe, but it's more lightweight than pipes.
I think the previous message is correct, only the specific listener should be added to the message.

src/chunkserver/bgjobs_unittest.cc
```diff
@@ -62,45 +68,71 @@ class JobPoolTest : public ::testing::Test {
...
 TEST_F(JobPoolTest, ProcessJob) {
-	jobPool->addJob(JobPool::ChunkOperation::Read, mockJobCallback, &counter, mockProcessJob);
+	const int nrJobsToProcess = 10;
+	std::vector<int> expectedCounters(kNrOperationTypes, 0);
+	std::vector<int> expectedProcessingCount(kNrListeners, 0);
 
-	// Wait for the job to be processed
-	std::this_thread::sleep_for(std::chrono::milliseconds(kWaitTimeMs));
+	for (int i = 0; i < nrJobsToProcess; ++i) {
+		auto opType = std::rand() % kNrOperationTypes;
```
ralcolea, line 118: The steps of selecting the operation and the target listener are not deterministic, which makes the unit test not deterministic and in case of failures, it makes more difficult to reproduce the test to identify the cause of the failure. I don't think this should remain in the final version.

I suggest you two different options:
- Use a deterministic random number generator with a fixed seed to make the test deterministic.
- As the number of jobs match the number of operation types, make sure to test always a different operation by assigning `opType = i`.

src/chunkserver/bgjobs.cc
```diff
@@ -73,28 +86,35 @@ JobPool::~JobPool() {
...
+		close(listenerInfo.notifierFD);
+	}
 }
 
 uint32_t JobPool::addJob(ChunkOperation operation, JobCallback callback, void *extra,
-                         ProcessJobCallback processJob) {
-	std::unique_lock lock(jobsMutex);
-	uint32_t jobId = nextJobId++;
+                         ProcessJobCallback processJob, uint32_t listenerId) {
+	auto &listenerInfo = listenerInfos_[listenerId];
```
ralcolea, line 112: Many `JobPool` functions rely on having a valid `listenerId`.
Maybe, it's worth to validate `listenerId` is valid with an assertion, `std::exception` or any other mechanism to avoid possible runtime errors or crashes.

ralcolea (commented):
Great job @dmga44!!!
Please, see my comments.

src/chunkserver/bgjobs.h
dmga44: Updated, now it says `/// Mutex for event notifications.`.

src/chunkserver/bgjobs.cc
dmga44: Applied, I used the analogous one to the end of the funcion for closing the file descriptors.

src/chunkserver/bgjobs.cc
dmga44: Indeed, it was my mistake while rebasing some old WIP changes, thank you.

src/chunkserver/bgjobs.cc
dmga44: I handled this case in most of the JobPool functions by logging the issue with `safs::log_warn` and directly returning, without doing the invalid operation.

For the `JobPool::addJob` function, besides the logging, I resetted the `listenerId` to 0 to make it valid and being able to perform the operation (an exception there could possibly crash the whole chunkserver).

src/chunkserver/bgjobs_unittest.cc
dmga44: I went with the first of your proposals:
- `std::srand(42);  // Seed for random number generation`

I don't like the second approach because it won't check the case when several jobs of the same type (or end up doing similar things) are being processed altogether, seemed too simple to me 😅 .

The idea of having it always random was to check with all the executions a broader set of scenarios (a new one every time), which will be fixed now. If the test were to have some probability `p` of failure, the expected time to reproduce (by modifying the test and logging the seed) is `1/p` seconds.

ralcolea (approved):
Great job @dmga44!!! :clap: :clap: :clap:

PR #452, opened 2025-06-17
feat(mds,chunkserver): Add CLUSTER_ID config var

lgsilva3087:
This change adds the CLUSTER_ID configuration variable for metadata (master, shadow) and chunkservers. The new variable should be set at the beginning and it is not intended to be changed frequently.

Metadata servers only accept connections from chunkservers with the same CLUSTER_ID (sent during registration), this prevents connections from undesired chunkservers.

To avoid breaking existent installations, the new variable defaults to 'default'. It is advisable for new installations to provide a meaningfull CLUSTER_ID.

### DIFF
--- a/src/chunkserver/master_connection.cc
+++ b/src/chunkserver/master_connection.cc
@@ -77,6 +77,14 @@ void MasterConn::reloadConfig() {
 	auto newMasterHostStr_ = cfg_getstring("MASTER_HOST", "sfsmaster");
 	auto newMasterPortStr_ = cfg_getstring("MASTER_PORT", "9420");
 
+	auto newClusterId = cfg_getstring("CLUSTER_ID", "default");
+
+	if (newClusterId != clusterId_) {
+		safs::log_warn(
+		    "MasterConn: Non-reloadable CLUSTER_ID changed from {} to {} during reload. Using the original value until restart.",
+		    clusterId_, newClusterId);
+	}
+
 	if (newMasterHostStr_ == masterHostStr_ && newMasterPortStr_ == masterPortStr_) {
 		return;  // no change
 	}
@@ -98,7 +106,7 @@ void MasterConn::reloadConfig() {
 			setMode(ConnectionMode::KILL);
 		}
 	} else {
-		safs::log_warn("master connection module: can't resolve master host/port ({}:{})",
+		safs::log_warn("MasterConn: can't resolve master host/port ({}:{})",
 		               masterHostStr_, masterPortStr_);
 	}
 }
@@ -126,6 +134,8 @@ void MasterConn::sendRegister() {
 	createAttachedPacket(
 	    cstoma::registerHost::build(myip, myport, gTimeout_ms, SAUNAFS_VERSHEX, clusterId_));
 
+	safs::log_info("MasterConn: registering to MDS: {}", address_.toString());
+
 	registrationStatus_ = RegistrationStatus::kRegistrationRequested;
 }
 
@@ -134,13 +144,13 @@ void MasterConn::onRegistered(const std::vector<uint8_t> &data) {
 	matocs::registerHost::deserialize(data, status);
 
 	if (status != SAUNAFS_STATUS_OK) {
-		safs::log_err("Master connection registration failed with status: {}",
+		safs::log_err("MasterConn: registration to {} failed with status: {}", address_.toString(),
 		              saunafs_error_string(status));
 		setMode(ConnectionMode::KILL);
 		return;
 	}
 
-	safs::log_info("Registered to MDS at {}", address_.toString());
+	safs::log_info("MasterConn: registered to MDS: {}", address_.toString());
 
 	registrationStatus_ = RegistrationStatus::kHostRegistered;
 
@@ -187,7 +197,7 @@ int MasterConn::initConnect() {
 			address_.port = mport;
 			isMasterAddressValid_ = true;
 		} else {
-			safs::log_warn("master connection module: can't resolve master host/port ({}:{})",
+			safs::log_warn("MasterConn: can't resolve master host/port ({}:{})",
 			               masterHostStr_, masterPortStr_);
 			return -1;
 		}
@@ -196,12 +206,12 @@ int MasterConn::initConnect() {
 	socketFD_ = tcpsocket();
 
 	if (socketFD_ < 0) {
-		safs::log_error_code(errno, "master connection module: create socket error");
+		safs::log_error_code(errno, "MasterConn: create socket error");
 		return -1;
 	}
 
 	if (tcpnonblock(socketFD_) < 0) {
-		safs::log_error_code(errno, "master connection module: set nonblock error");
+		safs::log_error_code(errno, "MasterConn: set nonblock error");
 		tcpclose(socketFD_);
 		socketFD_ = -1;
 		return -1;
@@ -209,7 +219,7 @@ int MasterConn::initConnect() {
 
 	if (bindHostAddress_.ip > 0) {
 		if (tcpnumbind(socketFD_, bindHostAddress_.ip, 0) < 0) {
-			safs::log_error_code(errno, "master connection module: can't bind socket to given ip");
+			safs::log_error_code(errno, "MasterConn: can't bind socket to given ip");
 			tcpclose(socketFD_);
 			socketFD_ = -1;
 			return -1;
@@ -219,7 +229,7 @@ int MasterConn::initConnect() {
 	int status = tcpnumconnect(socketFD_, address_.ip, address_.port);
 
 	if (status < 0) {
-		safs::log_error_code(errno, "master connection module: connect failed");
+		safs::log_error_code(errno, "MasterConn: connect failed");
 		tcpclose(socketFD_);
 		socketFD_ = -1;
 		isMasterAddressValid_ = false;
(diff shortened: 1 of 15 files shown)


src/master/matocsserv.cc
```diff
@@ -964,35 +964,58 @@ void matocsserv_got_duptruncchunk_status(matocsserventry* eptr, const std::vecto
...
 	eptr->servip   = servip;
 	eptr->servport = servport;
 	eptr->timeout  = timeout;
+
+	if (eptr->version >= kFirstVersionWithClusterId) {
+		if (clusterId.empty() || clusterId != gClusterId) {
+			safs::log_err("SAU_CSTOMA_REGISTER_HOST received empty cluster ID");
+			eptr->mode = KILL;
+			return;
+		}
```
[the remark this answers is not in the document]
lgsilva3087: Done.

src/master/matocsserv.cc
```diff
@@ -964,35 +964,58 @@ void matocsserv_got_duptruncchunk_status(matocsserventry* eptr, const std::vecto
...
 	eptr->version  = version;
 	eptr->servip   = servip;
 	eptr->servport = servport;
 	eptr->timeout  = timeout;
+
+	if (eptr->version >= kFirstVersionWithClusterId) {
+		if (clusterId.empty() || clusterId != gClusterId) {
+			safs::log_err("SAU_CSTOMA_REGISTER_HOST received empty cluster ID");
+			eptr->mode = KILL;
+			return;
```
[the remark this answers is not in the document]
lgsilva3087: Done.

src/chunkserver/master_connection.cc
```diff
@@ -43,6 +43,7 @@
 #include "common/output_packet.h"
 #include "common/sockets.h"
 #include "config/cfg.h"
+#include "errors/saunafs_error_codes.h"
```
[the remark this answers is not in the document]
lgsilva3087: Removed.

src/chunkserver/masterconn.cc
```diff
@@ -262,14 +262,21 @@ int masterconn_init(void) {
...
 	gMasterPort = cfg_getstring("MASTER_PORT", "9420");
+	std::string clusterId = cfg_getstring("CLUSTER_ID", "");
 	gBindHostStr = cfg_getstring("BIND_HOST", "*");
 	gTimeout_ms = get_cfg_timeout();
 	gEnableLoadFactor = static_cast<bool>(cfg_getuint32("ENABLE_LOAD_FACTOR", 0));
 
+	if (clusterId.empty()) {
+		safs::log_err("masterconn_init: CLUSTER_ID is not set or empty in the configuration");
+		return -1;
+	}
```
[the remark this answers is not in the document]
lgsilva3087: The specific error is logged.

src/master/filesystem.cc
```diff
@@ -338,6 +338,12 @@ static void fs_read_goal_config_file() {
 }
 
 static void fs_read_config_file() {
+	gClusterId = cfg_getstring("CLUSTER_ID", "");
+
+	if (gClusterId.empty()) {
+		throw ConfigurationException("CLUSTER_ID is not defined in configuration file");
+	}
```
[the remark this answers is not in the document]
lgsilva3087: The text of the exception, which is printed, is clarifying the error.

src/master/matocsserv.cc
```diff
@@ -1024,13 +1047,23 @@ void matocsserv_space(matocsserventry *eptr,const uint8_t *data,uint32_t length)
...
 	uint32_t timeout;
-	cstoma::registerHost::deserialize(data, servip, servport, timeout, version);
-	matocsserv_register_host(eptr, version, servip, servport, timeout);
+	std::string clusterId;
+
+	PacketVersion packetVersion;
+	deserializePacketVersionNoHeader(data, packetVersion);
+
+	if (packetVersion == cstoma::registerHost::kWithClusterId) {
+		cstoma::registerHost::deserialize(data, servip, servport, timeout, version, clusterId);
```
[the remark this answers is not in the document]
lgsilva3087: Older chunkservers can not use this version of the protocol packet.

tests/tools/saunafs.sh
```diff
@@ -332,6 +332,7 @@ create_bdb_name_storage_entry_() {
 create_sfsmaster_master_cfg_() {
 	local this_module_cfg_variable="MASTER_${masterserver_id}_EXTRA_CONFIG"
 	echo "PERSONALITY = master"
+	echo "CLUSTER_ID = cluster_01"
```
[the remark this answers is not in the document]
lgsilva3087: This is a default value, it can be overwritten by using the EXTRA_CONFIG family of variables.

src/master/filesystem.cc
```diff
@@ -338,6 +338,12 @@ static void fs_read_goal_config_file() {
...
+
```
uristdwarf: Main problem I see here is that existing setups could stop working unexpectedly, thus making this a breaking change

uristdwarf: Maybe it's better that the program itself generates a UUID and then stores it in /var/lib/saunafs (or whatever the data directory is) in a file? Because this also potentially introduces user error with defining the config variables.

src/master/matocsserv.cc
```diff
@@ -964,35 +964,66 @@ void matocsserv_got_duptruncchunk_status(matocsserventry* eptr, const std::vecto
...
+
 	eptr->csdb = csdb_find(eptr->servip, eptr->servport);
+
 	safs_pretty_syslog(LOG_NOTICE, "chunkserver register begin (packet version: 5) - ip: %s, port: %"
 			PRIu16, eptr->servstrip, eptr->servport);
+
+	// Send the answer with the status
+	if (eptr->version >= kFirstVersionWithClusterId) {
+		OutputPacket outPacket;
+		matocs::registerHost::serialize(outPacket.packet, 0);
```
ralcolea: If **0** represents a successful registration, please consider using `SAUNAFS_STATUS_OK`.

ralcolea (approved):
Great job @lgsilva3087!!! :+1: :fire: :muscle:

src/chunkserver/master_connection.cc
```diff
@@ -318,15 +347,16 @@ void MasterConn::readFromSocket() {
...
 			return;
 		}
 
 		if (ret < 0) {
 			if (errno != EAGAIN) {
-				safs::log_error_code(errno, "read from Master error");
-				mode_ = ConnectionMode::KILL;
+				safs::log_error_code(errno, "MasterConn: read from Master error: {}",
+				                     address_.toString());
+				setMode(ConnectionMode::KILL);
```
dmga44: Please note here that it will print something like `MasterConn: read from Master error: 192.168.50.206 (Read error)`.

src/protocol/cstoma.h
```diff
@@ -50,13 +50,24 @@ SAUNAFS_DEFINE_PACKET_SERIALIZATION(
 		cstoma, chunkNew, SAU_CSTOMA_CHUNK_NEW, kECChunks,
 		std::vector<ChunkWithVersionAndType>, chunks)
 
+SAUNAFS_DEFINE_PACKET_VERSION(cstoma, registerHost, kBeforeClusterId, 0)
+SAUNAFS_DEFINE_PACKET_VERSION(cstoma, registerHost, kWithClusterId, 1)
+
 SAUNAFS_DEFINE_PACKET_SERIALIZATION(
-		cstoma, registerHost, SAU_CSTOMA_REGISTER_HOST, 0,
+		cstoma, registerHost, SAU_CSTOMA_REGISTER_HOST, kBeforeClusterId,
```
dmga44: Maybe `kWithoutClusterId` would look better?

src/chunkserver/master_connection.cc
lgsilva3087: Done.

src/protocol/cstoma.h
lgsilva3087: Done.

src/master/matocsserv.cc
lgsilva3087: Done.

src/master/filesystem.cc
lgsilva3087: Nice catchs. I added a default value for this option, which will not break existing installations. Anyway, new installations could start using it to protect from connections from non intended chunkservers.
I agree, this function is expected to evolve in the future to be more robust (when multiple masters will be allowed), this change just adds it to the protocol.

uristdwarf: Just be sure to remove the breaking ! symbol from the commit

uristdwarf: As that can cause confusion when the changelog is generated

uristdwarf: I guess my only final concern is that are you sure you don't want to change this configuration option later (e.g not have it as a configuration option or something similar)? If not, you might consider making the option experimental (even though it's already used). Just a thought.

Anyway, feel free to close this thread after you've read my messages

rolysr (approved):
LGTM.

src/chunkserver/master_connection.cc
```diff
@@ -98,7 +98,7 @@ void MasterConn::reloadConfig() {
 			setMode(ConnectionMode::KILL);
 		}
 	} else {
-		safs::log_warn("master connection module: can't resolve master host/port ({}:{})",
+		safs::log_warn("MasterConn: can't resolve master host/port ({}:{})",
```
dmga44, line 110: On the `MasterConn::reloadConfig` function: would it make sense to also reload the `CLUSTER_ID` parameter? Please consider it.

If not, it would make sense to add a message that it is not being reloaded.

dmga44 (approved):
LGTM.

Taking into consideration this is a breaking change, that our major release is almost done, does it make sense to merge it to dev?

src/chunkserver/master_connection.cc
lgsilva3087: The CLUSTER_ID should be set at the beginning of the installation and it is not expected to change frequently, just in very specific rare situations. I added the message to warn if the user tries to reload it.

PR #458, opened 2025-06-19
Deprecate LizardFS code equal or older than v3.12.0

rolysr:
The idea of this PR is to deprecate all LizardFS related code for versions equal or older than v3.12.0, because code from these version will not be anymore supported from SaunaFS v5.0.0.

When reviewing this PR and in case you have some regular expression finder on you IDE, you can check all the occurrences that I used for finding potential code to deprecate: `masterversion|saunafsVersion|->version|HEX|>=  0x|<= 0x|->ver|>=  \(|<= \)|\.ver|kStdVersion|kFirstXorVersion|kFirstECVersion|kACL11Version|kRichACLVersion|kEC2Version|kDisconnectedChunkserverVersion|saunafsVersionToString|kECChunks|kStandardAndXorChunks|server_version|legacy`


src/master/chunks.cc
```diff
@@ -1997,16 +1988,13 @@ bool ChunkWorker::tryReplication(Chunk *c, ChunkPartType part_to_recover,
...
-		c->needverincrease = 1;
-		return true;
-	}
+	matocsserv_send_sau_replicatechunk(destination_server, c->chunkid, c->version,
+										part_to_recover, all_servers,
+										all_parts);
+	stats_replications++;
+	metrics::Counter::increment(metrics::Counter::Master::CHUNK_REPLICATE);
+	c->needverincrease = 1;
+	return true;
```
uristdwarf: This doesn't seem correct

src/master/matoclserv.cc
```diff
@@ -516,18 +516,14 @@ class SaunaFsStdXorPacketSerializer : public SaunaFsPacketSerializer {
 };
 
 
-const PacketSerializer* PacketSerializer::getSerializer(PacketHeader::Type type, uint32_t version) {
+const PacketSerializer* PacketSerializer::getSerializer(PacketHeader::Type type, [[maybe_unused]] uint32_t version) {
```
uristdwarf, line 519: If it's not unused, why not remove it?

src/master/matocsserv.cc
```diff
@@ -378,12 +378,7 @@ std::vector<std::pair<matocsserventry *, ChunkPartType>> matocsserv_getservers_f
 			             kRandomEngine);
 		}
 
-		uint32_t min_version = std::max({
-			slice_traits::isXor(slice) ? kFirstXorVersion : 0,
-			slice_traits::isEC(slice) ? kFirstECVersion : 0,
-			slice_traits::isEC(slice) && slice_traits::ec::isEC2(slice) ? kEC2Version : 0,
-			min_server_version
-		});
+		uint32_t min_version = std::max({kEC2Version, min_server_version});
```
uristdwarf: Doesn't this break XOR?

uristdwarf (changes requested):
Initial first impressions. I will talk to you about this PR in more detail in a call.

src/master/chunks.cc
rolysr: Fixed.

src/master/matoclserv.cc
rolysr: Because this is part of the modern PackageSerializer that we use, where each packet contains also the packet version, which is requered for a complete modern packet during communication, we could leave this param here as it could be reused for other logics in the future, based on the corresponding version.

src/master/matocsserv.cc
rolysr: Fixed.

PR #460, opened 2025-06-23
fix: Fix CS restart while replicating

dmga44:
The version of a chunk being replicated is set to 0 from the begining.
The changes introduced in 11e46f1 set the metadata filename to empty if
that was the case. In some scenarios it may lead to unsuccessful
chunkserver restart while replicating.


rolysr (approved):
Good catch.

lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM!!! :+1:

PR #461, opened 2025-06-23
feat: Avoid calculate CRC twice for replications

dmga44:
While doing a replication, the current behavior calculates the CRC of
the blocks to be written and then, if the chunkserver checks also CRC
while writing (HDD_CHECK_CRC_WHEN_WRITING=1 option, which is default)
it would have calculated twice the CRC for the same blocks. From one
calculation to the other the 4 byte value was being copied as a
function parameter, therefore should be safe to not check it again.


lgsilva3087 (approved):
LGTM

ralcolea (approved):
Nice catch @dmga44!!! :fire: :muscle:

PR #462, opened 2025-06-25
ci: Fix Ubuntu 24.04 images not being deleted

uristdwarf:
Due to a mistake, the tag for the image was being added twice, first in the Registry image name, and then again later when cleaning up, causing it to fail silently.

This should hopefully fix the no space problems for now.


ralcolea (approved):
Nice catch @uristdwarf!!! :fire: :muscle: :+1:

lgsilva3087 (approved):
LGTM

PR #464, opened 2025-06-25
fix(main): Fix test starting master when it's not running.

uristdwarf:
The LockStatus was returning the correct kTest when it was running, but returning LockStatus::kSuccess when not. This caused the master/chunkserver/metalogger to start up when running with test subcommand.
This was probably due to recent changes how test works, where it now checks for kTest when determining whether to start master or not. The commit makes sure that it correctly returns LockStatus::kTest when no lock is found.


rolysr (approved):

lgsilva3087 (approved):
LGTM

PR #467, opened 2025-06-26
feat(master): Make MetadataDumper extensible

lgsilva3087:
With new metadata backends comming soon, the former MetadataDumper needs to be easily extensible. This commit:

- Adds the new IMetadataDumper interface to allow extensibility.
- Renames MetadataDumper to be MetadataDumperFile, which is consistent with the implementation.
- Uses the new interface in IMetadataBackend and the concrete implementation where needed.
- Ensures internal linking (static) for the functions in MetadataBackendFile, to avoid collisions with other implementations (testing).

### DIFF
--- a/src/master/filesystem.cc
+++ b/src/master/filesystem.cc
@@ -44,7 +44,6 @@
 #include "master/metadata_backend_common.h"
 #include "master/metadata_backend_file.h"
 #include "master/metadata_backend_interface.h"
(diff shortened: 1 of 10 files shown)


antuan96314 (approved):

rolysr (approved):

ralcolea (approved):
Good job @lgsilva3087!!! :+1: :muscle: :fire:

PR #474, opened 2025-07-02
Make password digest size flexible

rolysr:
This PR is a step on the process of new password algorithms integration on our system. The goal of the PR is to add more flexibility for password digest handling, which will make it easier to implement newer password algorithms with bigger digest sizes.

PR #472, opened 2025-07-02
Refactor fs_connect function on client

rolysr:
These PR ensures that the `fs_connect` readability is improved by changing some variables names to more descriptive ones and also to set some magic numbers on the function as constants. The idea is to make it easier for developers to make some changes on the registration protocol function from client to master without expending much time on understanding `fs_connect`  implemention.

PR #480, opened 2025-07-04
feat: Add Argon2 password authentication

rolysr:
This commit integrates the Argon2 password-hash algorithm into the client-to-master password-based authentication flow. Clients can now choose between MD5 or Argon2 authentication; these methods are mutually exclusive. In the master’s sfsexports.cfg file you can specify which mount points require which algorithm.

NOTE!: This change should be moved from draft after merging expected changes from: 
https://github.com/leil-io/saunafs/pull/472
https://github.com/leil-io/saunafs/pull/474

The reason of waiting for those PRs is that the final password sent from client to master will be bigger, as it is one of the security requirements for ensuring the Argon2 password is safe enough.


rolysr:
Closing this for now due to shifting priorities. Will reopen if/when I resume work on it.

PR #482, opened 2025-07-09
test: Remove strict GID assertion in oplog test

wg1k:
When running under sudo, multiple legitimate GIDs may appear in the oplog (such as root's GID and the original user's GID). The primary purpose of this test is to ensure that encoded flag values don't appear in the oplog output, which is already validated by the previous assertion.


lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM

PR #473, opened 2025-07-02
Refactor filesystem metadata

ralcolea:
This PR refactors filesystem metadata files with the following changes:
- Replace raw uint8_t pointers used to manage extended attributes.
- Replace custom C-linked list used to store extended attributes (xattr_inode_entry_list)
  with an array of std::list of smart pointers.
- Replace custom C-linked lists used in FilesystemMetadata to store extended attributes.
- Remove unused code left over after the refactoring.
- Rename variables to use more meaningful names.
- Apply code style formatting to modified functions.

Signed-off-by: Crash <crash@leil.io>

PR #487, opened 2025-07-11
refactor(master): Replace std::list in hash buckets

ralcolea:
Since PR #473, some hash buckets in FilesystemMetadata, use `std::list` to store their elements. This creates a bigger memory overhead than other STL containers like `std::vector`.

This commit replaces `std::list` with `std::vector` in the affected buckets, aiming to reduce the memory overhead.

The screenshots below illustrate the memory consumption of different container types used to store filesystem metadata attributes. All tests were run using a 32 GiB metadata file containing over 194 million chunks, representing approximately 150 million files.

**1. Before PR #473 — Custom C-style linked lists with raw pointers.**
05-sfsmaster-before-PR-473

This version has the lowest memory footprint, as it relies on raw pointers and custom linked lists.
However, the implementation is more error-prone and harder to maintain, requiring manual pointer manipulation and explicit list traversal.

### &nbsp;
**2. After PR #473 — std::list with smart pointers (current dev)**
05-sfsmaster-after-PR-473

This version adopts modern C++ features like smart pointers and `std::list` for storing metadata entries in hash buckets.
While improving code safety and maintainability, it introduces **higher memory overhead**, mainly due to the allocation cost of every node in the `std::list` (~16–24 bytes per entry).

### &nbsp;
**3. This PR — Replace std::list with std::vector**
05-sfsmaster-update-nodeHash-with-std-vector

This version replaces `std::list` with `std::vector` in the relevant hash buckets of FilesystemMetadata.
The change reduces memory consumption of sfsmaster by eliminating the per-node overhead of `std::list`, while preserving safety and simplifying the data structure.

### DIFF
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -1328,10 +1328,9 @@ static inline void fsnodes_remove_node(uint32_t ts, FSNode *node) {
 			uint64_t chunkid = static_cast<FSNodeFile*>(node)->chunks[i];
 			if (chunkid > 0) {
 				if (chunk_delete_file(chunkid, node->goal) != SAUNAFS_STATUS_OK) {
-					safs_pretty_syslog(LOG_ERR,
-					                   "structure error - chunk %016" PRIX64
-					                   " not found (inode: %" PRIiNode " ; index: %" PRIu32 ")",
-					                   chunkid, node->id, i);
+					safs::log_err(
+					    "structure error - chunk {:#016x} not found (inode: {} ; index: {})",
+					    chunkid, node->id, i);
 				}
 			}
 		}
@@ -1358,7 +1357,9 @@ static inline void fsnodes_remove_node(uint32_t ts, FSNode *node) {
 	FSNode::destroy(node);
 
 	if (nodeIterator != gMetadata->nodeHash[nodeHashIndex].end()) {
-		gMetadata->nodeHash[nodeHashIndex].erase(nodeIterator);
+		auto lastElement = gMetadata->nodeHash[nodeHashIndex].end() - 1;
+		std::iter_swap(nodeIterator, lastElement); // Swap with last element to avoid erase: O(1)
+		gMetadata->nodeHash[nodeHashIndex].pop_back(); // Remove the last element: O(1)
 	}
 }
 
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -639,8 +639,8 @@ FSNode *fsnodes_create_node(uint32_t ts, FSNodeDirectory *parent, const HString 
 		node->gid = gid;
 	}
 
-	uint32_t nodepos = NODEHASHPOS(node->id);
-	gMetadata->nodeHash[nodepos].push_front(node);
+	uint32_t nodeHashIndex = NODEHASHPOS(node->id);
+	gMetadata->nodeHash[nodeHashIndex].push_back(node);
 
 	fsnodes_update_checksum(node);
 	fsnodes_link(ts, parent, node, name);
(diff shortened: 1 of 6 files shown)


src/master/metadata_backend_file.cc
```diff
@@ -913,13 +915,17 @@ void fs_new(void) {
 	gMetadata->root->mode = 0777;
 	gMetadata->root->uid = 0;
 	gMetadata->root->gid = 0;
-	nodepos = NODEHASHPOS(gMetadata->root->id);
-	gMetadata->nodeHash[nodepos].push_front(gMetadata->root);
+
+	auto hashRootIndex = NODEHASHPOS(gMetadata->root->id);
```
rolysr: Consider here to keep `hashRootIndex` as `uint32_t` in case you see that this `fs_new` is very frequently used. Also if you conclude using auto here makes not difference or is better, consider to unify it on the other places on the code that are calling `NODEHASHPOS(...)`.

rolysr (approved):
LGTM. Please consider my minor suggestion.

src/master/filesystem_node.cc
```diff
@@ -1301,59 +1301,64 @@ void fsnodes_change_uid_gid(FSNode *p, uint32_t uid, uint32_t gid) {
...
-				if (chunk_delete_file(chunkid, toremove->goal) != SAUNAFS_STATUS_OK) {
-					safs_pretty_syslog(LOG_ERR, "structure error - chunk %016" PRIX64
-					                " not found (inode: %" PRIiNode
-					                " ; index: %" PRIu32 ")",
-					       chunkid, toremove->id, i);
+				if (chunk_delete_file(chunkid, node->goal) != SAUNAFS_STATUS_OK) {
+					safs_pretty_syslog(LOG_ERR,
+					                   "structure error - chunk %016" PRIX64
+					                   " not found (inode: %" PRIiNode " ; index: %" PRIu32 ")",
+					                   chunkid, node->id, i);
```
dmga44: Consider to use `safs::log_err()`.

src/master/filesystem_node.cc
```diff
@@ -1301,59 +1301,64 @@ void fsnodes_change_uid_gid(FSNode *p, uint32_t uid, uint32_t gid) {
...
+
+	// remove node from nodeHash
+	uint32_t nodeHashIndex = NODEHASHPOS(node->id);
+	auto nodeIterator = std::find(gMetadata->nodeHash[nodeHashIndex].begin(),
+	                              gMetadata->nodeHash[nodeHashIndex].end(), node);
+
+	FSNode::destroy(node);
+
+	if (nodeIterator != gMetadata->nodeHash[nodeHashIndex].end()) {
+		gMetadata->nodeHash[nodeHashIndex].erase(nodeIterator);
```
dmga44: Consider to use `std::remove_if` and `resize`.

dmga44 (approved):
Nice change @ralcolea ! 👏 🔥 

Consider to mention in the commit description that you also did a small refactor of the related code (variable name changes, comments, spaces added for readability).

src/master/filesystem_node.cc
```diff
@@ -1301,59 +1301,64 @@ void fsnodes_change_uid_gid(FSNode *p, uint32_t uid, uint32_t gid) {
...
-	});
 
-	if (gChecksumBackgroundUpdater.isNodeIncluded(toremove)) {
-		removeFromChecksum(gChecksumBackgroundUpdater.fsNodesChecksum, toremove->checksum);
+	if (gChecksumBackgroundUpdater.isNodeIncluded(node)) {
+		removeFromChecksum(gChecksumBackgroundUpdater.fsNodesChecksum, node->checksum);
 	}
-	removeFromChecksum(gMetadata->fsNodesChecksum, toremove->checksum);
+
+	removeFromChecksum(gMetadata->fsNodesChecksum, node->checksum);
```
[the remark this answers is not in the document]
ralcolea: Your suggestion is incorrect. It's the other way around.
The `gChecksumBackgroundUpdater.isNodeIncluded(node)` check cannot be done after the node is destroyed.
Otherwise, it will be triggered a segfault due to use-after-free.

src/master/filesystem_xattr.cc
```diff
@@ -60,7 +62,9 @@ static void xattr_update_checksum(XAttributeDataEntry *xattrDataEntry) {
 static inline void xattr_removeentry(XAttributeInodeEntry *entry,
                                      XAttributeDataEntry *xattrDataEntry) {
 	// Remove the data from the xattr_inode_entry
-	entry->xattrDataList.remove(xattrDataEntry);
+	entry->xattrDataEntries.erase(
+	    std::remove(entry->xattrDataEntries.begin(), entry->xattrDataEntries.end(), xattrDataEntry),
+	    entry->xattrDataEntries.end());
```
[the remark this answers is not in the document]
ralcolea: Nice catch :) Applied.

src/master/filesystem_xattr.cc
```diff
@@ -114,8 +118,8 @@ void xattr_removeinode(inode_t inode) {
 	for (auto attributeIterator = start; attributeIterator != end;) {
 		xattrInodeEntry = attributeIterator->get();
 		if (xattrInodeEntry->inode == inode) {
-			while (!xattrInodeEntry->xattrDataList.empty()) {
-				xattr_removeentry(xattrInodeEntry, xattrInodeEntry->xattrDataList.front());
+			while (!xattrInodeEntry->xattrDataEntries.empty()) {
+				xattr_removeentry(xattrInodeEntry, xattrInodeEntry->xattrDataEntries.front());
```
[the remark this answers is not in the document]
ralcolea: This is already validated in `xattr_removeentry()` function.

src/master/metadata_backend_file.cc
ralcolea: Considering this was the only reference of `NODEHASHPOS()` where it was not used `uint32_t`, I reverted the line to preserve the `uint32_t` type.

src/master/filesystem_node.cc
ralcolea: Applied.

src/master/filesystem_node.cc
ralcolea: In this case, considering that nodes are not duplicated, `std::find` + `erase` should be a little bit more efficient than `std::remove_if`, which is more appropriate to erase multiple elements.

ralcolea: I updated the removal with a more efficient approach:
```cpp
auto nodeIterator = std::find(gMetadata->nodeHash[nodeHashIndex].begin(),
                              gMetadata->nodeHash[nodeHashIndex].end(), node);

FSNode::destroy(node);

if (nodeIterator != gMetadata->nodeHash[nodeHashIndex].end()) {
	auto lastElement = gMetadata->nodeHash[nodeHashIndex].end() - 1;
	std::iter_swap(nodeIterator, lastElement); // Swap with last element to avoid erase: O(1)
	gMetadata->nodeHash[nodeHashIndex].pop_back(); // Remove the last element: O(1)
}
```

wg1k (approved):

src/master/filesystem_node.cc
dmga44: As you are not enforcing any order of the elements in each bucket, I think the proposal is very good :+1:

PR #488, opened 2025-07-11
fix: Check master version for write algorithm

dmga44:
Previous version of the code was connecting to old master and using
chunk based write algorithm, which is very unsafe. The proposed fix is
to forcefully set the write algorithm to inode based one and log the
change.

### DIFF
--- a/src/common/saunafs_version.h
+++ b/src/common/saunafs_version.h
@@ -50,4 +50,5 @@ constexpr uint32_t kEC2Version = saunafsVersion(3, 13, 0);
 constexpr uint32_t kFirstVersionWithPathByInodeHiddenFile = saunafsVersion(4, 8, 0);
 constexpr uint32_t kFirstVersionWithUseQuotaInVolumeSize = saunafsVersion(4, 9, 0);
 constexpr uint32_t kFirstVersionWithMountInfoOnMonitoring = saunafsVersion(5, 0, 0);
+constexpr uint32_t kFirstVersionWithChunkBasedWriteAlgorithm = saunafsVersion(5, 0, 0);
 constexpr uint32_t kFirstVersionWithClusterId = saunafsVersion(5, 0, 0);
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -49,6 +49,7 @@
 #include "common/datapack.h"
 #include "common/errno_defs.h"
 #include "common/lru_cache.h"
+#include "common/saunafs_version.h"
 #include "errors/sfserr.h"
 #include "common/richacl_converter.h"
 #include "slogger/slogger.h"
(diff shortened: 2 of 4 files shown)


src/common/saunafs_version.h
```diff
@@ -50,4 +50,5 @@ constexpr uint32_t kEC2Version = saunafsVersion(3, 13, 0);
 constexpr uint32_t kFirstVersionWithPathByInodeHiddenFile = saunafsVersion(4, 8, 0);
 constexpr uint32_t kFirstVersionWithUseQuotaInVolumeSize = saunafsVersion(4, 9, 0);
 constexpr uint32_t kFirstVersionWithMountInfoOnMonitoring = saunafsVersion(5, 0, 0);
+constexpr uint32_t kFirstVersionWithChunkBasedAlgorithm = saunafsVersion(5, 0, 0);
```
aNeutrino:
```suggestion
constexpr uint32_t kFirstVersionWithChunkBasedWriteAlgorithm = saunafsVersion(5, 0, 0);
```

rolysr (approved):
LGTM.

antuan96314 (approved):

src/mount/sauna_client.cc
```diff
@@ -3671,6 +3672,11 @@ void fs_init(FsInitParams &params) {
 			params.prefetch_xor_stripes,
 			std::max(params.bandwidth_overuse, 1.));
 
+	if (masterversion < kFirstVersionWithChunkBasedWriteAlgorithm) {
+		fprintf(stderr, "Master version is too old, using inode-based write algorithm");
```
aNeutrino:
```suggestion
		fprintf(stderr, "Metadata server version %s is too old,\nthis is why we force using inode-based write algorithm instead of chunk-based write algorithm.\nPlease upgrade your metadata server to minimum version %s", saunafsVersionToString
 , kFirstVersionWithChunkBasedWriteAlgorithm.toString());
```

src/mount/sauna_client.cc
```diff
@@ -3671,6 +3672,11 @@ void fs_init(FsInitParams &params) {
...
+		params.use_inode_based_write_algorithm = true;
```
aNeutrino, line 3683: is this AFTER we got parameters from the user already parsed and applied... so we do not risk that user will override this ?

src/mount/sauna_client.cc
```diff
@@ -3671,6 +3672,17 @@ void fs_init(FsInitParams &params) {
 			params.prefetch_xor_stripes,
 			std::max(params.bandwidth_overuse, 1.));
 
+	if (!params.use_inode_based_write_algorithm &&
+	    masterversion < kFirstVersionWithChunkBasedWriteAlgorithm) {
+		fprintf(stderr,
+		        "Metadata server version v%s is too old, using inode-based write algorithm"
+		        "(sfsuseinodebasedwritealgorithm=1). "
+		        "Required version for chunk based algorithm is v%s.",
```
aNeutrino:
```suggestion
		        "Required minimum version for chunk based algorithm is v%s.",
```

src/mount/sauna_client.cc
dmga44: This is after we got parameters, so user won't be able to override it.

aNeutrino (approved):

uristdwarf (approved):

PR #489, opened 2025-07-12
fix(chunkserver): Fallback to old registration if needed

lgsilva3087:
Fixes new Chunkservers connecting to Master in versions 4.x.

Starting from v5.0.0, chunkservers register with the Master and expect a response containing status and version. Old Masters (v4.x) simply disconnect on the unknown registration message. 

Previously, chunkservers would just retry the connection. Now we log a warning when no registration response is received, so users see the likely cause of the failure.

This change also introduces a retry and fallback mechanism to detect possible registration attempts to old Master and then uses the old registration procedure.

A test was added to cover the initial old Master rejection, the succesfull Chunkserver reconnection using the old procedure and a reconnection after Master restarts.

### DIFF
--- a/src/chunkserver/master_connection.cc
+++ b/src/chunkserver/master_connection.cc
@@ -132,35 +132,55 @@ void MasterConn::sendRegister() {
 	uint32_t myip = mainNetworkThreadGetListenIp();
 	uint16_t myport = mainNetworkThreadGetListenPort();
 
-	createAttachedPacket(
-	    cstoma::registerHost::build(myip, myport, gTimeout_ms, SAUNAFS_VERSHEX, clusterId_));
+	if (isVersionLessThan5_) {  // Preserve compatibility with masters < 5.0
+		createAttachedPacket(
+		    cstoma::registerHost::build(myip, myport, gTimeout_ms, SAUNAFS_VERSHEX));
 
-	safs::log_info("MasterConn: registering to MDS: {}", address_.toString());
+		safs::log_warn("MasterConn: using old master registration to {}", address_.toString());
 
-	registrationStatus_ = RegistrationStatus::kRegistrationRequested;
+		registrationStatus_ = RegistrationStatus::kHostRegistered;
+
+		onRegistered({});
+	} else {
+		createAttachedPacket(
+		    cstoma::registerHost::build(myip, myport, gTimeout_ms, SAUNAFS_VERSHEX, clusterId_));
+
+		safs::log_info("MasterConn: registering to MDS: {}", address_.toString());
+
+		registrationStatus_ = RegistrationStatus::kRegistrationRequested;
+	}
 }
 
 void MasterConn::onRegistered(const std::vector<uint8_t> &data) {
-	uint8_t status{};
-	uint32_t version{};
-	std::string clusterId;
-	matocs::registerHost::deserialize(data, status, version, clusterId);
+	if (isVersionLessThan5_) {
+		(void)data;  // Not used
+		version_ = saunafsVersion(0, 0, 0);  // The version is unknown at this point
+	} else {
+		uint8_t status{};
+		uint32_t version{};
+		std::string clusterId;
+		matocs::registerHost::deserialize(data, status, version, clusterId);
 
-	if (status != SAUNAFS_STATUS_OK) {
-		safs::log_err(
-		    "MasterConn: registration to {} failed with status: {}, version: {}, clusterId: {}",
-		    address_.toString(), saunafs_error_string(status), saunafsVersionToString(version),
-		    clusterId);
-		setMode(ConnectionMode::KILL);
-		return;
+		if (status != SAUNAFS_STATUS_OK) {
+			safs::log_err(
+			    "MasterConn: registration to {} failed with status: {}, version: {}, clusterId: {}",
+			    address_.toString(), saunafs_error_string(status), saunafsVersionToString(version),
+			    clusterId);
+			setMode(ConnectionMode::KILL);
+			return;
+		}
+
+		version_ = version;
+
+		safs::log_info("MasterConn: registered to MDS: {}, version: {}, clusterId: {}",
+		               address_.toString(), saunafsVersionToString(version), clusterId);
+
+		registrationStatus_ = RegistrationStatus::kHostRegistered;
 	}
 
-	version_ = version;
-
-	safs::log_info("MasterConn: registered to MDS: {}, version: {}, clusterId: {}",
-	               address_.toString(), saunafsVersionToString(version), clusterId);
-
-	registrationStatus_ = RegistrationStatus::kHostRegistered;
+	// Reset registration parameters for future reconnections to use the new protocol first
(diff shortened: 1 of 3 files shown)


tests/test_suites/ShortSystemTests/test_upgrade_new_chunkserver_with_old_saunafs.sh
```diff
@@ -0,0 +1,35 @@
...
+# Stop old chunkserver and start the new one
+saunafsXX_chunkserver_daemon 0 stop
+saunafs_chunkserver_daemon 0 start
+
+# Give some time for the new chunkserver to log connection issues
+sleep $(timeout_rescale_seconds 5)
+
+# Ensure that versions are switched
+saunafs_admin_master info | head -1
+assert_success saunafs_admin_master list-chunkservers | grep "disconnected"
```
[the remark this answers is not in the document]
lgsilva3087: Applied.

tests/test_suites/ShortSystemTests/test_upgrade_new_chunkserver_with_old_saunafs.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Check if chunkserver log contains the expected message
+assert_success $(grep -qi "Master server did not answer the registration request" "${TEMP_DIR}/log")
```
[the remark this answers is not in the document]
lgsilva3087: Applied.

src/chunkserver/master_connection.cc
```diff
@@ -364,6 +364,13 @@ void MasterConn::readFromSocket() {
 
 		if (ret == 0) {
 			safs::log_info("MasterConn: connection reset by Master: {}", address_.toString());
+
+			if (registrationStatus_ == RegistrationStatus::kRegistrationRequested) {
+				safs::log_warn(
+				    "MasterConn: Master server did not answer the registration request; make sure Master is at least version {}",
+				    saunafsVersionToString(kFirstVersionWithClusterId));
+			}
+
```
uristdwarf: How are we sure that the problem is the master version? Could be a false-positive message

lgsilva3087: Could be, that's why it is just a warning hinting a possible cause.

uristdwarf (approved):

aNeutrino (approved):

wg1k (commented):
The retry and fallback mechanism to support old Masters (v4.x) makes sense and improves visibility by logging warnings when no response is received.

That said, I want to flag a real-world scenario that might still cause issues, especially in unstable network conditions:

1. Chunkserver tries to register.
2. Network flaps: retries (say, 3) fail.
3. Chunkserver assumes it’s dealing with an _old_ master (due to no reply).
4. It downgrades to the old registration protocol (no cluster ID).
5. When connectivity is restored, if the Master is actually v5.x (new), the chunkserver still uses the legacy protocol.
   
What would happen then?

- If the new Master rejects legacy registrations (which is likely the safer behavior), the chunkserver doesn’t automatically retry the new registration flow, potentially requiring manual intervention or a restart.
- If the Master accepts the old registration despite being new, it might cause subtle misconfigurations or unexpected behavior (e.g., incorrect cluster association that the cluster ID addition was designed to prevent).

Please correct me if I missed anything.

lgsilva3087:
> The retry and fallback mechanism to support old Masters (v4.x) makes sense and improves visibility by logging warnings when no response is received.
> 
> That said, I want to flag a real-world scenario that might still cause issues, especially in unstable network conditions:
> 
> 1. Chunkserver tries to register.
> 2. Network flaps: retries (say, 3) fail.
> 3. Chunkserver assumes it’s dealing with an _old_ master (due to no reply).
> 4. It downgrades to the old registration protocol (no cluster ID).
> 5. When connectivity is restored, if the Master is actually v5.x (new), the chunkserver still uses the legacy protocol.
> 
> What would happen then?
> 
> * If the new Master rejects legacy registrations (which is likely the safer behavior), the chunkserver doesn’t automatically retry the new registration flow, potentially requiring manual intervention or a restart.
> * If the Master accepts the old registration despite being new, it might cause subtle misconfigurations or unexpected behavior (e.g., incorrect cluster association that the cluster ID addition was designed to prevent).
> 
> Please correct me if I missed anything.

How did you tested it?

dmga44:
> What would happen then?
> If the new Master rejects legacy registrations (which is likely the safer behavior), the chunkserver doesn’t automatically retry the new registration flow, potentially requiring manual intervention or a restart.
> If the Master accepts the old registration despite being new, it might cause subtle misconfigurations or unexpected behavior (e.g., incorrect cluster association that the cluster ID addition was designed to prevent).
> Please correct me if I missed anything.

From what I see in the code, new master won't accept legacy connections from new chunkservers. So, if connection flaps the legacy attempts won't succeed and some of the new protocol ones should succeed eventually. If the connection issues persist, then issue is connection not the registering scheme.

New master accepts legacy connections from old chunkservers, so it is consistent.

That is my understanding @ictus4u .

dmga44 (approved):
Nice job, @lgsilva3087 .

ralcolea (approved):
Great job @lgsilva3087. Very nice test 👍

wg1k (approved):

uristdwarf (approved):

src/chunkserver/master_connection.cc
```diff
@@ -132,35 +132,55 @@ void MasterConn::sendRegister() {
...
-		safs::log_err(
-		    "MasterConn: registration to {} failed with status: {}, version: {}, clusterId: {}",
-		    address_.toString(), saunafs_error_string(status), saunafsVersionToString(version),
-		    clusterId);
-		setMode(ConnectionMode::KILL);
-		return;
-	}
+	if (isVersionLessThan5_) {
+		(void)data;  // Not used
+		version_ = saunafsVersion(0, 0, 0);  // The version is unknown at this point
```
rolysr, line 157: Consider naming this as a conventional `kUnknownSaunaFSVersion ` or similar name to make it clear is a default invalid one.

src/chunkserver/master_connection.h
```diff
@@ -209,6 +209,12 @@ class MasterConn {
 	std::string clusterId_;                      ///< Cluster ID for this connection.
 	std::shared_ptr<JobPool> jobPool_;           ///< Shared reference to the JobPool.
 
+	// For compatibility with old masters (version < 5.0)
```
rolysr, line 212: It would be good to explicitly show 5.0.0 in this comment to be as clear as possible.

rolysr (approved):
LGTM. Please consider my minor suggestions.

src/chunkserver/master_connection.cc
lgsilva3087: We can do it in next changes without loosing the merge window.

src/chunkserver/master_connection.h
lgsilva3087: I think the comment matches all previous versions.

PR #496, opened 2025-07-17
fix: Fix write algorithm in delayed init mode

dmga44:
This PR targets fixing the inconsistent selection of write
algorithm when using the `sfsdelayedinit` option in client.

Main feature of sfsdelayedinit option is to be able to start the mount
without an available master, which means it is not expected after short
notice to know the master version. This means client is not supposed to
display any warning message right after mounting and select the
appropiate write algorithm after knowing the current master version.

Implements the support for the client to switch the
write algorithm used when needed depending on master version.

An integration was added and another one was enhanced in order
to test the proposed changes. Also, a specific test for the
`sfsdelayedinit` option was added in order to check its expected
behavior.

### DIFF
--- a/src/mount/mastercomm.cc
+++ b/src/mount/mastercomm.cc
@@ -60,6 +60,7 @@
 #include "mount/exports.h"
 #include "mount/notification_area_logging.h"
 #include "mount/stats.h"
+#include "mount/writedata.h"
 #include "protocol/SFSCommunication.h"
 #include "protocol/cltoma.h"
 #include "protocol/matocl.h"
@@ -131,6 +132,7 @@ static pthread_t rpthid,npthid;
 static std::mutex fdMutex, recMutex;
 
 static uint32_t sessionid;
+static uint32_t masterversion;
 
 static char masterstrip[17];
 static uint32_t masterip=0;
@@ -140,6 +142,8 @@ static uint32_t srcip=0;
 
 static uint8_t fterm;
 static std::atomic<bool> gIsKilled(false);
+
+static SaunaClient::FsInitParams gSaunaFSInitParams;
 
 typedef std::unordered_map<PacketHeader::Type, PacketHandler*> PerTypePacketHandlers;
 static PerTypePacketHandlers perTypePacketHandlers;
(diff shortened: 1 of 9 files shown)


src/mount/mastercomm.cc
```diff
@@ -196,6 +200,70 @@ struct InitParams {
...
+			safs::log_warn(
+			    "New metadata server supports chunk based write algorithm "
+			    "(current master version: {}, required: {}).",
+			    saunafsVersionToString(masterversion),
+			    saunafsVersionToString(kFirstVersionWithChunkBasedWriteAlgorithm));
+		}
+
+		assert(gUseInodeBasedWriteAlgorithm && gIsInodeBasedWriteAlgorithmInitialized);
+		write_data_term();
+		gIsChunkBasedWriteAlgorithmInitialized = false;
```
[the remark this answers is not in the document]
dmga44: Nice catch.

src/mount/mastercomm.cc
```diff
@@ -196,6 +200,70 @@ struct InitParams {
...
+			        saunafsVersionToString(masterversion).c_str(),
+			        saunafsVersionToString(kFirstVersionWithChunkBasedWriteAlgorithm).c_str());
+		} else {
+			safs::log_warn(
+			    "Metadata server version v{} is too old, using inode based write algorithm"
+			    "(sfsuseinodebasedwritealgorithm=1). "
+			    "Required minimum version for chunk based write algorithm is v{}.",
+			    saunafsVersionToString(masterversion),
+			    saunafsVersionToString(kFirstVersionWithChunkBasedWriteAlgorithm));
+		}
```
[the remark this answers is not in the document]
dmga44: This function is only going to be called after a connection with a master has been stablished. Therefore, the master version should not be never 0.

src/mount/mastercomm.cc
```diff
@@ -196,6 +200,68 @@ struct InitParams {
...
+	// algorithm and it is intended by params, then we should use it.
+	if (gIsInodeBasedWriteAlgorithmInitialized && gUseInodeBasedWriteAlgorithm &&
+	    !gSaunaFSInitParams.use_inode_based_write_algorithm &&
+	    masterversion >= kFirstVersionWithChunkBasedWriteAlgorithm) {
+		if (isFromMainThread) {
+			fprintf(stderr,
+			        "New metadata server supports chunk based write algorithm "
+			        "(current master version: %s, required: %s).\n",
+			        saunafsVersionToString(masterversion).c_str(),
+			        saunafsVersionToString(kFirstVersionWithChunkBasedWriteAlgorithm).c_str());
```
uristdwarf, line 243: Maybe use fmt instead of fprintf? fmt at least does compile time type checking.

src/mount/mastercomm.cc
```diff
@@ -196,6 +200,68 @@ struct InitParams {
...
+		} else {
+			safs::log_warn(
+			    "New metadata server supports chunk based write algorithm "
+			    "(current master version: {}, required: {}).",
+			    saunafsVersionToString(masterversion),
+			    saunafsVersionToString(kFirstVersionWithChunkBasedWriteAlgorithm));
+		}
+
+		assert(gUseInodeBasedWriteAlgorithm && gIsInodeBasedWriteAlgorithmInitialized);
```
uristdwarf: Rather than just assert (because of ndebug), I would add passert or something to explicitly crash this even in release builds.

src/mount/writedata.cc
```diff
@@ -683,6 +683,10 @@ void *write_worker(void *) {
 void write_data_init(uint32_t cachesize, uint32_t retries, uint32_t workers,
                      uint32_t writewindowsize, uint32_t chunkserverTimeout_ms,
                      uint32_t cachePerInodePercentage, uint32_t waveTimeout) {
+	if (gIsInodeBasedWriteAlgorithmInitialized) {
+		return;
```
rolysr: As a doubt, why is this kind of check needed on the `write_data_init`? On which scenario will the corresponding write algorithm be initialized before the `write_data_init`?

tests/test_suites/ShortSystemTests/test_upgrade_master_switch_write_algo.sh
```diff
@@ -0,0 +1,85 @@
...
+# Test if all files produced so far are readable on legacy mount:
+cd "${info[mount0]}/from_new_mount"
+assert_success file-validate new_mount_old_master_file
+
+saunafsXX_master_daemon stop
+
+# Force the master to create new sessions (and clients to connect instead of reconnect)
+rm $(get_current_master_sessions_file)
+
+# Don't know why, but we to bypass the saunafs_master_daemon function
```
rolysr: Check this comment. Should it be `but we have to bypass`?

tests/test_suites/ShortSystemTests/test_upgrade_master_switch_write_algo.sh
```diff
@@ -0,0 +1,85 @@
...
+assert_success file-validate old_mount_old_master_file
+
+cd ..
+mkdir from_new_mount
+cd from_new_mount
+# Test if reading and writing on new SaunaFS mount works:
+assert_success generate_file new_mount_old_master_file
+assert_success file-validate new_mount_old_master_file
+
+# Test if all files produced so far are readable on legacy mount:
```
rolysr, line 51: Should this comment here refer to new mount instead of legacy? As I see from this next lines, you are reading legacy-created files on new mount.

rolysr (commented):
Consider checking my suggestions.

tests/test_suites/ShortSystemTests/test_upgrade_master_switch_write_algo.sh
```diff
@@ -0,0 +1,85 @@
...
+# Test checks if new SaunaFS mount works with legacy versions of master and chunkservers
+# and if it can read and write files created on legacy SaunaFS mount.
+
+export SAFS_MOUNT_COMMAND="sfsmount"
+
+CHUNKSERVERS=2 \
+  MOUNTS=2 \
+  START_WITH_LEGACY_SAUNAFS=YES \
+  USE_RAMDISK=YES \
+  MOUNT_EXTRA_CONFIG="sfscachemode=NEVER" \
```
uristdwarf, line 12: What about `sfsdelayedinit` option?

uristdwarf (approved):
Looks fine.

src/mount/writedata.cc
dmga44: The proposed solution to the issue calls `write_data_init` if connected to a new master, so a previous master connection to a master supporting the previously used write algorithm could have called `write_data_init` before and have the structures initialized before the next call.

tests/test_suites/ShortSystemTests/test_upgrade_master_switch_write_algo.sh
dmga44: Nope.

I'm reading files from legacy mount (`"${info[mount0]}"` implies it) that were created from the new mount.

tests/test_suites/ShortSystemTests/test_upgrade_master_switch_write_algo.sh
dmga44: True, thank you.

src/mount/mastercomm.cc
dmga44: Well, I did it just like the rest of the logging of client initialization, the way that it appears like all the other messages.

src/mount/mastercomm.cc
dmga44: Applied.

tests/test_suites/ShortSystemTests/test_upgrade_master_switch_write_algo.sh
dmga44: Well, this specific test checks that the change of the write algorithm depending on the connected master is working.

The `sfsdelayedinit` option is kind of a subset of this type of behavior (is only to skip the wait at begining when no master is available), while the test checks that new client can connect to old master and use inode based wirte algorithm and then connect to new master and use chunk based write algorithm. I could add a test for the `sfsdelayedinit` option, but it would look very similar to this one.

rolysr (approved):
Nice catch. LGTM.

ralcolea (approved):
Great job @dmga44! 👍 
Nice job fixing the Ganesha multi export test!

PR #497, opened 2025-07-17
feat(common): Introduce a thread pool class

wg1k:
This commit adds a `ThreadPool` class for executing tasks in parallel, along with a set of unit tests to verify its functionality.

The `ThreadPool` class manages a pool of worker threads to execute tasks concurrently.
It provides methods to enqueue tasks and execute them in parallel, with proper synchronization to ensure thread safety.
The class is designed as a singleton, providing a global instance that can be accessed from anywhere in the application.

The unit tests cover various scenarios, including:
- Enqueuing tasks and verifying their results.
- Executing tasks in parallel using the `parallelFor` method.
- Testing the singleton instance and its lifecycle.
- Verifying thread safety and proper synchronization.

### DIFF
--- a/src/common/thread_pool.h
+++ b/src/common/thread_pool.h
@@ -1,0 +1,247 @@
+/*
+   Copyright 2025      Leil Storage OÜ
+
+   This file is part of SaunaFS
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


src/common/thread_pool.h
```diff
@@ -0,0 +1,233 @@
...
+	};
+
+	for (std::size_t workerIndex = 0; workerIndex < workerCount; ++workerIndex)
+		enqueue([wrapper, workerIndex] { wrapper(workerIndex); });
+
+	done.wait();
+	if (exception) std::rethrow_exception(exception);
+}
+
+}  // namespace safs
```
rolysr: Remember to leave an empty line at the end of the file.

wg1k: tx

wg1k:
The immediate purpose for which this class was introduced has shifted; therefore, I am closing this PR until an obvious need arises.

PR #510, opened 2025-07-29
Refactor meta fuse mount

rolysr:
### Refactor meta-mounts and mountpoint logic

This PR improves the maintainability and clarity of the code related to meta-mounts and general mountpoint handling.

**Key changes:**

* Replaced magic numbers in meta-mount code with named constants.
* Refactored directory and path logic to use modern C++ constructs.
* Introduced `DirectoryBuffer` and `PathBuffer` structs to simplify and clarify directory/path handling.
* Centralized global variables used by both regular and metadata FUSE mountpoints.
* Moved the duplicated `PKGVERSION` function to a shared location for reuse.

These changes reduce duplication, improve readability, and make the code easier to extend and debug.


src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -546,83 +557,90 @@ void replyDirReadRoot(fuse_req_t request, off_t offset, size_t maxsize) {
...
-	dirinfo->lock.lock();
-	dirinfo->lock.unlock();
-	free(dirinfo->p);
-	delete dirinfo;
-	fi->fh = 0;
-	fuse_reply_err(req,0);
+	auto dirinfo = std::unique_ptr<DirectoryBuffer>(reinterpret_cast<DirectoryBuffer *>(fi->fh));
+	if (dirinfo) {
+		dirinfo->lock.lock();
+		dirinfo->lock.unlock();
```
antuan96314: Why is this need it???

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
-		pathinfo->size = strlen((char*)path) + 1;
-		if (!(pathinfo->p = (char*)malloc(pathinfo->size))) {
-			safs_pretty_syslog(LOG_EMERG, "out of memory");
-			delete pathinfo;
-			return;
-		}
-		memcpy(pathinfo->p, path, pathinfo->size - 1);
-		pathinfo->p[pathinfo->size - 1] = '\n';
+		pathinfo->changed = false;
+		pathinfo->size = strlen((char *)path) + 1;
```
antuan96314: Avoid old C-Style cast operations.
```suggestion
		pathinfo->size = strlen(reinterpret_cast<const char *>(path)) + 1;
```

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
+		pathinfo->path = std::string((char *)path);
```
antuan96314: Avoid old C-Style cast operations.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
+	constexpr off_t kMasterInfoSize =
+	    14;  // size of the master info buffer with version information.
 #else
-		if (off>=10) {
-			fuse_reply_buf(req,NULL,0);
-		} else if (off+size>10) {
-			fuse_reply_buf(req,(char*)(masterinfo+off),10-off);
+	constexpr off_t kMasterInfoSize =
+	    10;  // size of the master info buffer without version information.
 #endif
```
antuan96314: Consider deprecating old masterinfo size 10.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
+	auto pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
+
+	if (ino == SPECIAL_INODE_MASTERINFO) {
+		uint8_t masterinfo[14];
```
antuan96314, line 712: If only one size is supported (see comment above), then,
```suggestion
		uint8_t masterinfo[kMasterInfoSize];
```

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
+		fs_getmasterlocation(masterinfo);
+		masterproxy_getlocation(masterinfo);
+
+		if (off >= kMasterInfoSize) {
+			fuse_reply_buf(req, nullptr, 0);
+		} else if (off + size > kMasterInfoSize) {
+			fuse_reply_buf(req, (char *)(masterinfo + off), kMasterInfoSize - off);
```
antuan96314: Avoid old C-Style cast operations.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
 		} else {
-			fuse_reply_buf(req,(char*)(masterinfo+off),size);
+			fuse_reply_buf(req, (char *)(masterinfo + off), size);
```
antuan96314: Avoid old C-Style cast operations.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
+	if (pathinfo->changed == 0) { pathinfo->size = 0; }
+
+	if (off + size > pathinfo->size) {
+		pathinfo->size = off + size;
+		pathinfo->path.resize(off + size, '\0');
 	}
-	memcpy((pathinfo->p)+off,buf,size);
+
+	std::memcpy(pathinfo->path.data() + off, buf, size);
 	pathinfo->changed = 1;
```
antuan96314:
```suggestion
	pathinfo->changed = true;
```

antuan96314 (approved):
Consider my comments, but not as stoppers to merge.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +323,204 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
-			name = (const char*)dbuff;
-			dbuff+=nleng;
-			getINode(&dbuff, inode);
-			sprintf((char*)buff,"%08" PRIXiNode "|",inode);
-			if (nleng>255-9) {
-				memcpy(buff+9,name,255-9);
-				buff+=255;
+
+			put8bit(&buff, inodeLength);
+			name = (const char *)dataBuffer;
```
ralcolea: You can avoid the cast declaring `name` to be the same type as dataBuffer, because name is only used to copy the pointer.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +323,204 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
+	dirbuf->size = 0;
+	metaEntriesSize = getDirMetaEntriesSize(ino);
 
 	if (ino == SPECIAL_INODE_META_TRASH) {
-		status = fs_gettrash(&dbuff, &dsize);
-		if (status != SAUNAFS_STATUS_OK)
-			return;
-		dcsize = dir_dataentries_size(dbuff,dsize);
+		status = fs_gettrash(&bufData, &entriesSize);
+		if (status != SAUNAFS_STATUS_OK) return;
```
ralcolea: The `return` statement is not inside braces, while other in the same function are.
Please, unify.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +323,204 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
 	}
 
-	if (msize > 0)
-		dir_metaentries_fill(b->p, ino);
+	if (metaEntriesSize > 0) { fillDirMetaEntries(dirbuf->buffer.data(), ino); }
 
-	if (dcsize > 0)
-		dir_dataentries_convert(b->p + msize, dbuff, dsize);
+	if (dataEntriesSize > 0) {
+		convertDirDataEntries(dirbuf->buffer.data() + metaEntriesSize, bufData,
```
ralcolea: This function fits in a single line.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +323,204 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
 void sfs_meta_opendir(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
-	dirbuf *dirinfo;
-
-	if (ino == SPECIAL_INODE_ROOT || ino == SPECIAL_INODE_META_TRASH ||
-	    ino == SPECIAL_INODE_META_UNDEL || ino == SPECIAL_INODE_META_RESERVED) {
-
-		dirinfo = new dirbuf();
-
-		dirinfo->p = NULL;
+	constexpr auto isValidMetaMountInode = [](fuse_ino_t i) {
```
ralcolea: Consider using `inode`.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +323,204 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
-		fi->fh = (unsigned long)dirinfo;
+		dirinfo->wasRead = false;
+		fi->fh = reinterpret_cast<uintptr_t>(dirinfo.release());
 
-		if (fuse_reply_open(req,fi) == -ENOENT) {
+		if (fuse_reply_open(req, fi) == -ENOENT) {
 			fi->fh = 0;
-			free(dirinfo->p);
-			delete dirinfo;
 		}
```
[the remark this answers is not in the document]
ralcolea: Please check Gemini suggestion, it makes sense and it could be a real leak.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -47,32 +47,37 @@
...
+// Number of bytes used for metadata fields (inode, type, etc.) after the name in a directory
+// entry
+constexpr size_t kDirEntryMetaSize = 5;
+
+// Total stride (name length + metadata) to advance to the next directory entry
+constexpr size_t kDirEntryStride = 6;
+
+// Number of bytes reserved for extra fields in a directory entry,
+// such as inode, type, and any additional metadata.
+constexpr size_t kDirEntryExtraFieldsSize = 9;
```
dmga44: Please substitute these constants by the sum of the lengths of the fields they represent. Keep in mind that if the inode values were 64 bits long, these sizes are wrong.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -546,83 +557,90 @@ void replyDirReadRoot(fuse_req_t request, off_t offset, size_t maxsize) {
...
+			}
+		}
+	}
+}
 
-	if (off<0) {
-		fuse_reply_err(req,EINVAL);
+void sfs_meta_readdir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
+                      struct fuse_file_info *fi) {
+	auto dirinfo = reinterpret_cast<DirectoryBuffer *>(fi->fh);
```
ralcolea: Consider using `auto *dirinfo = ...`.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,119 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
 		fi->direct_io = 1;
-		fi->fh = (unsigned long)pathinfo;
+		fi->fh = reinterpret_cast<uintptr_t>(pathinfo.release());
 
-		if (fuse_reply_open(req, fi) == -ENOENT) {
-			fi->fh = 0;
-			free(pathinfo->p);
-			delete pathinfo;
-		}
+		if (fuse_reply_open(req, fi) == -ENOENT) { fi->fh = 0; }
```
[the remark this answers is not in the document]
ralcolea: Please, consider the Gemini suggestion about the leak. It makes sense.

src/mount/fuse/sfs_meta_fuse.cc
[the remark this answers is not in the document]
dmga44: @rolysr There is a lot of truth in these words, please check suggestion. Another idea is to not using the `unique_ptr`, but a bare one.

`unique_ptr` is suitable for cases in which we move the pointer ownership to some other structure. In this case we need to cast it somehow to `uintptr_t`, which counters most of its use.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
+	    14;  // size of the master info buffer with version information.
 #else
-		if (off>=10) {
-			fuse_reply_buf(req,NULL,0);
-		} else if (off+size>10) {
-			fuse_reply_buf(req,(char*)(masterinfo+off),10-off);
+	constexpr off_t kMasterInfoSize =
+	    10;  // size of the master info buffer without version information.
 #endif
+	auto pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
```
ralcolea:
```suggestion
	auto *pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
```

src/mount/fuse/sfs_meta_fuse.cc
dmga44: Probably (not sure), is to make sure the `dirinfo` was not being used somewhere else (where the mutex is locked). So, we'd need to wait for the other operation to finish and then perform this one. Other idea is that it could be there to create less helgrind warnings.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +657,124 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
 	}
 }
 
-void sfs_meta_write(fuse_req_t req, fuse_ino_t ino, const char *buf, size_t size, off_t off, struct fuse_file_info *fi) {
-	pathbuf *pathinfo = (pathbuf *)((unsigned long)(fi->fh));
-	if (ino==SPECIAL_INODE_MASTERINFO) {
-		fuse_reply_err(req,EACCES);
+void sfs_meta_write(fuse_req_t req, fuse_ino_t ino, const char *buf, size_t size, off_t off,
+                    struct fuse_file_info *fi) {
+	auto pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
```
ralcolea:
```suggestion
	auto *pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
```

src/mount/fuse/sfs_meta_fuse.cc
dmga44: A question, can you call `std::unique_ptr` on `nullptr`? Makes sense that it should not possible, so the next if would be not necessary.

src/mount/fuse/sfs_fuse.h
```diff
@@ -25,6 +25,7 @@
 #include <fuse.h>
 #include <fuse_lowlevel.h>
 
+#include "mount/fuse/sfs_fuselib/metadata.h"
```
ralcolea: `metadata.h` is not used in the header, but in the source.
It can be moved to the source to avoid including it where is not needed.

src/mount/fuse/sfs_meta_fuse.cc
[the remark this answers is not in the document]
dmga44: @rolysr Please take this suggestion in consideration as well.

ralcolea (commented):
Nice job @rolysr
Please, see my suggestions.

dmga44 (commented):
Not related to the PR, but please consider changing the return value of `metadataNameToInode` to `inode_t`.

That said, please check my comments. Good job.

src/mount/fuse/sfs_meta_fuse.cc
uristdwarf: In any case, I'd invoke Chesterton's fence. Unless we know why it exists, we should keep it as it was like that before, since this is only a refactor.

uristdwarf: Wouldn't it lead to UB? @rolysr definitely check what happens if you call std::unique_ptr on nullptr.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +323,204 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
-	if (msize + dcsize == 0)
-		return;
+	if (metaEntriesSize + dataEntriesSize == 0) { return; }
 
-	if (!(b->p = (uint8_t*)malloc(msize + dcsize))) {
-		safs_pretty_syslog(LOG_EMERG, "out of memory");
+	try {
+		dirbuf->buffer.resize(metaEntriesSize + dataEntriesSize);
+	} catch (const std::bad_alloc &) {
+		safs::log_critical("out of memory");
```
uristdwarf: Bit generic log message. Try to use a more descriptive error message (what function?) and include the error message.

src/mount/fuse/sfs_meta_fuse.cc
uristdwarf: Why not use std::string for path?

src/mount/fuse/sfs_meta_fuse.cc
uristdwarf: You mean removing? For me deprecation means marking something as something should not be used anymore :sweat_smile:

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Good, I will take this as a consideration to remove the few references to all master info size in another PR. To make it clearer when deprecating it.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Considered to explicitly put the expected pointer type.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: I could consider to move this removal idea to another PR scope to better handle all cases on the mount module.

src/mount/fuse/sfs_meta_fuse.cc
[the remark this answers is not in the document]
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: the thing is that: `fs_gettrashpath` expects a `const uint8_t **` and returns a pointer to a C-style, null-terminated string buffer managed by the underlying code.
We cannot use std::string directly as an output parameter for this API.
Instead, we must first receive the result as a const uint8_t *, then convert it to std::string if needed. This ensures compatibility with the API and avoids unnecessary copies or type mismatches.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: In this case I decided to keep it as it was to keep the PR scope as intended, but as I see from the code, it was decided to make this function the only place releasing memory associated with this pointer. So according to what was designed, it is not intended to free this memory somewhere else.

rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Considered.

src/mount/fuse/sfs_meta_fuse.cc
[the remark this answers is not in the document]
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -111,25 +119,25 @@ static void sfs_attr_to_stat(inode_t inode, const Attributes &attr, struct stat
...
-
-	fuse_reply_statfs(req,&stfsbuf);
+	stfsbuf.f_frsize = kBlockSize;
+	stfsbuf.f_bsize = kBlockSize;
+	stfsbuf.f_blocks = trashspace / kBlockSize + reservedspace / kBlockSize;
+	stfsbuf.f_bfree = reservedspace / kBlockSize;
+	stfsbuf.f_bavail = reservedspace / kBlockSize;
+	stfsbuf.f_files = 1000000000 + PKGVERSION;
+	stfsbuf.f_ffree = 1000000000 + PKGVERSION;
+	stfsbuf.f_favail = 1000000000 + PKGVERSION;
```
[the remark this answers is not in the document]
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_fuse.h
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -47,32 +47,53 @@
...
+// Number of bytes used for metadata fields (inode + file type) after the name in a directory
+// entry. When using 32-bit inodes, the inode occupies 4 bytes and the file type occupies 1 byte.
+constexpr size_t kDirEntryMetaSize = 4 + 1;
+
+// Total number of bytes reserved at the start of a directory entry for extra fields,
+// such as a formatted inode string and separator before the name. This includes:
+//  - 8 characters for the hexadecimal inode string
+//  - 1 character for the '|' separator
+constexpr size_t kDirEntryExtraFieldsSize = 9;
+#endif
```
lgsilva3087: Consider to remove the pre-procesor conditionals. It could be, for instance:
```cpp
// kinode_t_size contains the correct size for 32 or 64 bits.
constexpr size_t kDirEntryMetaSize = kinode_t_size + 1;
```

Or

`constexpr size_t kDirEntryMetaSize = sizeof(inode_t) + 1;`

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -47,32 +47,53 @@
...
+
+// Total stride (name length + metadata) to advance to the next directory entry
+constexpr size_t kDirEntryStride = kDirEntryMetaSize + 1;
+
+// Block size used for filesystem statistics and block calculations (in bytes)
+constexpr size_t kBlockSize = 512;
```
lgsilva3087, line 63: Consider kSectorSize if you consider it more accurate.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -546,83 +573,89 @@ void replyDirReadRoot(fuse_req_t request, off_t offset, size_t maxsize) {
...
-	free(dirinfo->p);
 	delete dirinfo;
 	fi->fh = 0;
-	fuse_reply_err(req,0);
+	fuse_reply_err(req, 0);
 }
 
 void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
-	pathbuf *pathinfo;
+	PathBuffer *pathinfo = new PathBuffer();
```
lgsilva3087: Looks like a good place for a smart pointer. If something happens after the declaration and before the deletion (unusual exit point), the pointer will be leaked.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -47,32 +47,53 @@
...
 	size_t size;
 	std::mutex lock;
 };
 
-struct pathbuf {
-	int changed;
-	char *p;
+struct PathBuffer {
+	bool changed;
+	std::string path;
```
lgsilva3087, line 78: If you replaced the char * by a std::string, maybe size is not needed anymore, the string already contains it.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -47,32 +47,53 @@
...
+
+// Maximum file permission bits (all permission bits, sticky/setuid/setgid)
+constexpr mode_t kMaxFilePermissions = 07777;
+
+constexpr uint64_t kStatfsFilesBase = 1000'000'000 + PKGVERSION;
+
+struct DirectoryBuffer {
+	bool wasRead = false;
+	std::vector<uint8_t> buffer;
 	size_t size;
```
lgsilva3087: If this is the size of the vector, consider to remove the extra variable.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +672,127 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
-		masterproxy_getlocation(masterinfo);
+void sfs_meta_read(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
+                   struct fuse_file_info *fi) {
 #ifdef MASTERINFO_WITH_VERSION
-		if (off>=14) {
-			fuse_reply_buf(req,NULL,0);
-		} else if (off+size>14) {
-			fuse_reply_buf(req,(char*)(masterinfo+off),14-off);
+	constexpr off_t kMasterInfoSize =
+	    14;  // size of the master info buffer with version information.
```
lgsilva3087: Consider:
```cpp
// size of the master info buffer with version information.
constexpr off_t kMasterInfoSize = 14;
```
To avoid the strange formatting.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -639,126 +672,127 @@ void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
 #else
-		if (off>=10) {
-			fuse_reply_buf(req,NULL,0);
-		} else if (off+size>10) {
-			fuse_reply_buf(req,(char*)(masterinfo+off),10-off);
+	constexpr off_t kMasterInfoSize =
+	    10;  // size of the master info buffer without version information.
```
lgsilva3087: Same here.

lgsilva3087 (commented):
Nice hard job. Please consider my suggestions.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -47,23 +47,39 @@
...
+// Number of bytes used for metadata fields (inode + file type) after the name in a directory
+// entry. When using 32-bit inodes, the inode occupies 4 bytes and the file type occupies 1 byte.
+constexpr size_t kDirEntryMetaSize = 4 + 1;
+
+// Total number of bytes reserved at the start of a directory entry for extra fields,
+// such as a formatted inode string and separator before the name. This includes:
+//  - 8 characters for the hexadecimal inode string
+//  - 1 character for the '|' separator
+constexpr size_t kDirEntryExtraFieldsSize = 9;
+#endif
```
dmga44: Please consider to substitute these lines by:
```cpp
constexpr size_t kDirEntryMetaSize = sizeof(inode_t) + sizeof(uint8_t);
constexpr size_t kDirEntryExtraFieldsSize = 9;
```
plus explanation. Please notice that the `kDirEntryExtraFieldsSize` is effectively 9 right now because the `sprintf` is always printing 9 bytes regardless of the of the size of the inode variable. Maybe it could be renamed to `kDirEntryHexPreffixSize`?

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -747,7 +765,7 @@ void sfs_meta_read(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
 
 void sfs_meta_write(fuse_req_t req, fuse_ino_t ino, const char *buf, size_t size, off_t off,
                     struct fuse_file_info *fi) {
-	auto pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
+	PathBuffer * pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
```
dmga44: @ralcolea meant (and suggested :) ):
```cpp
auto *pathinfo = reinterpret_cast<PathBuffer *>(fi->fh);
```

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -630,17 +646,16 @@ void sfs_meta_readdir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
...
+	dirinfo->lock.lock();
+	dirinfo->lock.unlock();
+	delete dirinfo;
+	fi->fh = 0;
 	fuse_reply_err(req, 0);
 }
 
 void sfs_meta_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
-	auto pathinfo = std::make_unique<PathBuffer>();
+	PathBuffer *pathinfo = new PathBuffer();
```
dmga44: Consider to apply similar suggestion here as well.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -630,17 +646,16 @@ void sfs_meta_readdir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
 
 void sfs_meta_releasedir(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
 	(void)ino;
-	auto dirinfo = std::unique_ptr<DirectoryBuffer>(reinterpret_cast<DirectoryBuffer *>(fi->fh));
-	if (dirinfo) {
-		dirinfo->lock.lock();
-		dirinfo->lock.unlock();
-		fi->fh = 0;
-	}
+	DirectoryBuffer *dirinfo = reinterpret_cast<DirectoryBuffer *>(fi->fh);
```
dmga44: And here.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -598,7 +614,7 @@ void fillDirEntryBuffer(fuse_req_t req, DirectoryBuffer *dirinfo, off_t &off, ch
 
 void sfs_meta_readdir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
                       struct fuse_file_info *fi) {
-	auto dirinfo = reinterpret_cast<DirectoryBuffer *>(fi->fh);
+	DirectoryBuffer *dirinfo = reinterpret_cast<DirectoryBuffer *>(fi->fh);
```
dmga44: And here.

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -491,35 +507,35 @@ static void fillDirectoryBufferMeta(DirectoryBuffer *dirbuf, inode_t ino) {
...
-		return i == SPECIAL_INODE_ROOT || i == SPECIAL_INODE_META_TRASH ||
-		       i == SPECIAL_INODE_META_UNDEL || i == SPECIAL_INODE_META_RESERVED;
+	constexpr auto isValidMetaMountInode = [](inode_t inode) {
+		return inode == SPECIAL_INODE_ROOT || inode == SPECIAL_INODE_META_TRASH ||
+		       inode == SPECIAL_INODE_META_UNDEL || inode == SPECIAL_INODE_META_RESERVED;
 	};
 
 	if (isValidMetaMountInode(ino)) {
-		auto dirinfo = std::make_unique<DirectoryBuffer>();
+		DirectoryBuffer *dirinfo = new DirectoryBuffer();
```
dmga44: And here.

dmga44 (approved):
Please check my comments.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Considered, but due to previous design of `FileInfo` struct, which contains a field related to a file handler (`fh`) and this one expects to store the address this pointer is referring to, then for this specific case, changing the `fh` type for better integration with smart pointers usage would required another specific PR for it.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: I decided to leave it as `kBlockSize` because we're describing the filesystem's logical block size for allocation and statistics, as used in POSIX and `statvfs`.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done. Even improved the corresponding `sprintf` statement.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Done.

lgsilva3087 (approved):
LGTM

src/mount/fuse/sfs_meta_fuse.cc
```diff
@@ -315,201 +324,200 @@ static uint32_t dir_metaentries_size(inode_t ino) {
...
-	if (msize + dcsize == 0)
-		return;
+	if (metaEntriesSize + dataEntriesSize == 0) { return; }
 
-	if (!(b->p = (uint8_t*)malloc(msize + dcsize))) {
-		safs_pretty_syslog(LOG_EMERG, "out of memory");
+	try {
+		dirbuf->buffer.resize(metaEntriesSize + dataEntriesSize);
+	} catch (const std::bad_alloc &e) {
+		safs::log_critical("fillDirectoryBufferMeta: Memory allocation failed: {}", e.what());
```
ralcolea, line 495: In the future, consider using macro `__func__` if you want to log the name of the function.
Like this, the log line is shorter and if in the future, the function is renamed, you don't
have to worry about updating the log :)
```suggestion
		safs::log_critical("{}: Memory allocation failed: {}", __func__, e.what());
```

ralcolea (approved):

ralcolea (approved):
Great job!! @rolysr

src/mount/fuse/sfs_meta_fuse.cc
rolysr: Considered.

PR #514, opened 2025-08-01
fix(mount): Fix readahead under high latency

dmga44:
This change targets fixing some crashes of the clients while using the
readahead algorithm on some slow-to-respond clusters. These crashes
were caused in older versions due to accessing invalid cache entries
and in the current state of the code due to reuse of some cache entries
that make its state inconsistent.

The main case where this happened was the following:
- some ReadRecord scheduled several read requests in advance.
- the first read requests to be satisfied is not the one with least
offset (and first in the order of schedule), so it needs to wait until
all the previous ones get satisfied.
- that first resolved request waits too much and gets expired, and the
released (or sent to the entries pool), but not taken away from the
ReadRecords pending requests container.
- it is then added to some read reply (from the pending requests
container) while being invalid.

The proposed solution is to remove this kind of entries from the cache
(such that they won't appear for queries) but keeping them alive for as
long as they are not removed from the pending requests container.

Some consistency checks were also added.

Co-authored-by: Rolando Sánchez Ramos <rolysr@leil.io>

### DIFF
--- a/src/mount/readdata.cc
+++ b/src/mount/readdata.cc
@@ -125,6 +125,7 @@ uint32_t getBytesToBeReadFromCS(uint32_t index, uint32_t offset, uint32_t size,
 
 RequestConditionVariablePair *ReadaheadRequests::append(
     ReadaheadRequestPtr readaheadRequestPtr) {
+	readaheadRequestPtr->entry->isPendingNotify = true;
 	pendingRequests_.emplace_back(readaheadRequestPtr);
 	return &pendingRequests_.back();
 }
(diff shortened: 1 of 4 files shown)


lgsilva3087 (approved):
LGTM. Consider to change the title of this commit:

`fix(mount): Fix high latency reads`
It could be:
`fix(mount): Fix readahead under high latency`

The same with the PR title.

ralcolea (approved):
Great job @dmga44! 👍
Please consider suggestions from @lgsilva3087

PR #520, opened 2025-08-05
fix(charts): Fix wrong range value for CSV

uristdwarf:
Using modulo with the size of RANGE did not work correctly. It would give incorrect chart ranges for CSV (for example, chart id 90031 (medium range) would give 90033 (very long) instead).

The previous commit to this line explained that this fixed a compiler warning regarding array subscripting, however no such compiler warning appears for GCC 15.1.1. So it may have been a false positive.

### DIFF
--- a/src/common/charts.cc
+++ b/src/common/charts.cc
@@ -1852,6 +1852,10 @@ uint32_t charts_make_csv(uint32_t number) {
 
 	type = number / 10;
 	range = number % kNumberRangeModulus;
+	if (range >= RANGES) {
+		safs::log_warn("wrong range for chart {}", number);
+		return 0;
+	}
 	charts_filltab(c1dispdata, range, type, 1);
 	charts_filltab(c2dispdata, range, type, 2);
 	charts_filltab(c3dispdata, range, type, 3);
--- a/src/common/charts.cc
+++ b/src/common/charts.cc
@@ -1837,6 +1837,7 @@ int charts_fake_compress(uint8_t *src,uint32_t srcsize,uint8_t *dst,uint32_t *ds
 
 
 uint32_t charts_make_csv(uint32_t number) {
+	const int kNumberRangeModulus = 10;
 	uint32_t type, range;
 	uint32_t tm_year, tm_mon, tm_day, tm_hour, tm_min, tm_sec;
 	uint64_t c1dispdata[LENG];
@@ -1850,7 +1851,7 @@ uint32_t charts_make_csv(uint32_t number) {
 	tm_year = tm_mon = tm_day = tm_hour = tm_min = tm_sec = 0;
 
 	type = number / 10;
-	range = number % RANGES;
+	range = number % kNumberRangeModulus;
 	charts_filltab(c1dispdata, range, type, 1);
 	charts_filltab(c2dispdata, range, type, 2);
 	charts_filltab(c3dispdata, range, type, 3);


uristdwarf:
Tests have not been added, however there is already a integration test for the new monitoring, found here:

https://github.com/leil-io/cgi-rework/blob/dev/tests/test_main.py#L146

src/common/charts.cc
```diff
@@ -1850,7 +1850,7 @@ uint32_t charts_make_csv(uint32_t number) {
 	tm_year = tm_mon = tm_day = tm_hour = tm_min = tm_sec = 0;
 
 	type = number / 10;
-	range = number % RANGES;
+	range = number % 10;
```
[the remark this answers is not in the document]
uristdwarf: @gemini-code-assist

The practical idea of modulo with 10 is that you get the last number. Adding a named constant here would make it more confusing to reason with as it's not immediately then clear what the purpose is. Likewise, division is to remove the last number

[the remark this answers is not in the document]
antuan96314: Please avoid MAGIC constant as suggested by @gemini-code-assist.

rolysr: If not adding these constants as named values, then consider adding here a line commenting the purpose of this calculation in this part.

rolysr (approved):
LGTM. Consider my suggestion

src/common/charts.cc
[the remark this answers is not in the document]
uristdwarf: _\*sigh\*_. Fine

src/common/charts.cc
```diff
@@ -1850,7 +1851,7 @@ uint32_t charts_make_csv(uint32_t number) {
 	tm_year = tm_mon = tm_day = tm_hour = tm_min = tm_sec = 0;
 
 	type = number / 10;
```
ralcolea, line 1853: Please, verify the implications of your change for the rest of the function.
The previous versions expected `range` to be in the range `[0..3]` for the logic of lines 1859-1885.

After this change, unless `number` will always has the last digit in the range `[0..3]` the logic will change and the behavior of the function will be different, for instance to compute the values of `tm_day`, `tm_mon` and similars. Please, verify because the new change could assign incorrect values to other attributes.

ralcolea (commented):
Please, consider my comment to confirm the rest of the function works as expected.

src/common/charts.cc
uristdwarf: @ralcolea I think you indirectly spotted a very good issue. The previous behavior was incorrect, however it safely made sure that it would always be 3 or less. This is important because the pointers array size is tied to RANGES. I was testing to make sure behavior is as I expected it (that it would always pick the 0 range, i.e a minute), but I got a nonsense response. The png function actually checks this and returns if less than RANGES.

uristdwarf: As for what would happen in the previous behaviour, it would wrap around after 4 and start returning 0 range for 91005 (CPU), 1 range for 91006, etc. This is safe, but incorrect, as for chart 91030 it would return range 2 even though it should return 0.

ralcolea: Exactly this was my concern and that's why I asked :)

src/common/charts.cc
```diff
@@ -1850,7 +1851,11 @@ uint32_t charts_make_csv(uint32_t number) {
...
-	range = number % RANGES;
+	range = number % kNumberRangeModulus;
+	if (range >= RANGES) {
```
ralcolea, line 1855: Good solution @uristdwarf.
I think this solution is consistent with the logic of `charts_makedata()` and `charts_datasize()` functions.

ralcolea (approved):
Great job @uristdwarf!! 👍

PR #523, opened 2025-08-07
fix(cs): Fix release of read jobs and CS reloads

dmga44:
This change was motivated by a high memory consumption of the
chunkserver processes in some scenarios where some failed reads are
involved.

Some issues were fixed in the process:
- the CS is now able to release the structures related to failed reads.
- to test the motivating case, it was necessary to fix the case when
some disk is being removed from the CS hdd file and the CS is reloaded,
it was crashing from time to time.
- during that process, it was fixed the reload of the SYSLOG_IDENT
parameter and the thread id.

The changes were tested in a new test `test_concurrent_replications`.

### DIFF
--- a/src/chunkserver/chunkserver_entry.cc
+++ b/src/chunkserver/chunkserver_entry.cc
@@ -368,10 +368,15 @@ void ChunkserverEntry::prepareDiscardReadJobs() {
 	while (!pendingReadJobIds.empty()) {
 		// pendingReadJobIds and pendingReadDataPackets should have the related elements in the
 		// correct order
-		toDiscardReadJobIds.push_back(pendingReadJobIds.front());
-		pendingReadJobIds.pop_front();
+		if (pendingReadDataPackets.front()->outputBuffer->getStatus() == kNotSaunafsStatus) {
+			toDiscardReadJobIds.push_back(pendingReadJobIds.front());
+			toDiscardReadDataPackets.emplace_back(std::move(pendingReadDataPackets.front()));
+		} else {
+			// Already processed packets, can be moved to the pool
+			getReadOutputBufferPool().put(std::move(pendingReadDataPackets.front()->outputBuffer));
+		}
 
-		toDiscardReadDataPackets.emplace_back(std::move(pendingReadDataPackets.front()));
+		pendingReadJobIds.pop_front();
 		pendingReadDataPackets.pop_front();
 	}
 	assert(pendingReadJobIds.empty());
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -103,6 +103,20 @@ constexpr uint32_t startingOffsetOfBlock(uint32_t blocknum) { return blocknum * 
 
 inline std::atomic_bool gCheckCrcWhenReading{true};
 
+static std::mutex gDisksToBeDeletedWithPendingChunksMutex;
+static std::vector<std::pair<std::unique_ptr<IDisk>, std::vector<ChunkWithType>>>
+    gDisksToBeDeletedWithPendingChunks;
+static std::vector<std::pair<IDisk *, std::vector<ChunkWithType>>>
+    gNewDisksToBeDeletedWithPendingChunks;
+
+/// @enum SendDataToMasterMode
+/// @brief Represents the reason of sending data to the master server.
+enum class SendDataToMasterMode : uint8_t {
+	ForDiskRemoval,
+	ForDamagedDisk,
+	ForNewChunk
+};
+
 void hddGetDamagedChunks(std::vector<ChunkWithType>& chunks,
                          std::size_t limit) {
 	TRACETHIS();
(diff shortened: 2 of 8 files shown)


src/chunkserver/chunkserver_entry.cc
```diff
@@ -406,7 +411,7 @@ void ChunkserverEntry::readFinishedCallback(uint8_t status, void *entry) {
 
 		// after sending status even if there was an error it's possible to
 		// receive new requests on the same connection
-		eptr->state = State::Idle;
+		eptr->state = State::Close;
```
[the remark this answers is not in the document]
dmga44: Applied, more or less.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -368,10 +368,15 @@ void ChunkserverEntry::prepareDiscardReadJobs() {
 	while (!pendingReadJobIds.empty()) {
 		// pendingReadJobIds and pendingReadDataPackets should have the related elements in the
 		// correct order
-		toDiscardReadJobIds.push_back(pendingReadJobIds.front());
-		pendingReadJobIds.pop_front();
+		if (pendingReadDataPackets.front()->outputBuffer->getStatus() == kNotSaunafsStatus) {
+			toDiscardReadJobIds.push_back(pendingReadJobIds.front());
+			toDiscardReadDataPackets.emplace_back(std::move(pendingReadDataPackets.front()));
```
uristdwarf, line 373: Why push_back and emplace_back? Why not just use emplace_back?

uristdwarf (approved):

src/chunkserver/chunkserver_entry.cc
dmga44: The `push_back ` is because those are `uint32_t` values. If they were more complex structures I'd consider it.

The `emplace_back` is for the `std::unique_ptr<PacketStruct>` instances.

src/chunkserver/hddspacemgr.cc
```diff
@@ -317,22 +388,28 @@ void hddCheckDisks() {
...
+				}
 				gDisks.erase(it);
 				break;
 			}
 		}
 	}
+	diskToBeDeletedLock.unlock();
 
-	disksToRemove.clear();
+	gNewDisksToBeDeletedWithPendingChunks.clear();
```
rolysr, line 413: Why not to protect here the clear operation also on the lock?

src/slogger/slogger.cc
```diff
@@ -35,6 +35,12 @@
 
 #include "errors/sfserr.h"
 
+static std::mutex gSloggerMutex;
+
+std::mutex& safs::get_mutex() {
+	return gSloggerMutex;
+}
```
rolysr: I wonder how would this impact the system when many threads will be logging at the same time. If possible, consider if there is another workaround on the chunkserver side so that this issue could be solved without adding this mutex.

src/chunkserver/hddspacemgr.cc
dmga44: Because the `gNewDisksToBeDeletedWithPendingChunks` is only accessed by the mainloop thread (the one performing the reload), so it does not need protection.

The mutex protects the `gDisksToBeDeletedWithPendingChunks` because it will be modified by the reloading thread and the free resource thread.

src/slogger/slogger.cc
dmga44: I wonder the same, I don't know how this could impact the performance when many threads are logging. I expect it not to be too much because we are not logging much in every operation. As for workarounds, I tried one, but didn't succeed.

I think this commit should be considered as optional for the PR, as it did not solve the issue it was targetting though it made sure no helgrind warnings come up from chunkserver reload.

src/chunkserver/io_buffers.h
```diff
@@ -31,9 +31,14 @@
 #include "chunkserver-common/chunk_interface.h"
 #include "chunkserver/buffers_pool.h"
 #include "common/aligned_allocator.h"
+#include "slogger/slogger.h"
```
ralcolea: This header is pointed out by clangd as unused in this file.
Maybe, it can be moved to `io_buffers.cc` to avoid the warning and unneeded inclusion in other files.

src/chunkserver/hddspacemgr.cc
```diff
@@ -103,6 +103,19 @@ constexpr uint32_t startingOffsetOfBlock(uint32_t blocknum) { return blocknum *
...
 inline std::atomic_bool gCheckCrcWhenReading{true};
 
+using ChunkIdentifier = std::pair<uint64_t, ChunkPartType>;
+std::mutex gDisksToBeDeletedWithPendingChunksMutex;
+inline std::vector<std::pair<std::unique_ptr<IDisk>, std::vector<ChunkIdentifier>>>
+    gDisksToBeDeletedWithPendingChunks;
+inline std::vector<std::pair<IDisk *, std::vector<ChunkIdentifier>>>
+    gNewDisksToBeDeletedWithPendingChunks;
+
+enum class SendDataToMasterMode : uint8_t {
```
ralcolea, line 114: Consider adding documentation for possible values of `SendDataToMasterMode`.

src/chunkserver/hddspacemgr.cc
```diff
@@ -2539,6 +2617,9 @@ void hddTerminate(void) {
 		}
 	}
 
+	// Destroy the disks-to-be-deleted related structures
```
ralcolea: Is it always safe to delete the disks-to-be-deleted structures before `gChunksMap`?
As a precaution, I will move it to the end of the function, after `gDisks.clear()`.

src/chunkserver/hddspacemgr.cc
```diff
@@ -103,6 +103,19 @@ constexpr uint32_t startingOffsetOfBlock(uint32_t blocknum) { return blocknum *
 
 inline std::atomic_bool gCheckCrcWhenReading{true};
 
+using ChunkIdentifier = std::pair<uint64_t, ChunkPartType>;
```
lgsilva3087: Consider the existing ChunkWithType.

src/chunkserver/io_buffers.cc
```diff
@@ -436,3 +440,32 @@ void InputBuffer::setFinished() { state_.store(WriteState::Finished); }
...
+	// could be negative because some buffers could be added to the pools in the meantime.
+	int releasedOutputBuffers =
+	    initialTotalOutputBufferBlocks - gCurrentTotalOutputBufferBlocks.load();
+	int releasedInputBuffers =
+	    initialTotalInputBufferBlocks - gCurrentTotalInputBufferBlocks.load();
+	int releasedReplicatorBuffers =
+	    initialTotalReplicatorBufferBlocks - gCurrentTotalReplicatorBufferBlocks.load();
+	// Log the number of released buffers.
+	safs::log_debug(
+	    "(releaseOldIoBuffers) Released buffer blocks per operation: read {}, write {}, replicate {}",
```
ralcolea: Consider using `__func__` macro to avoid typing manually the name of the function.
```suggestion
	    "({}) Released buffer blocks per operation: read {}, write {}, replicate {}",
	    __func__, gCurrentTotalOutputBufferBlocks.load(), gCurrentTotalInputBufferBlocks.load(),
```

src/chunkserver/hddspacemgr.cc
lgsilva3087, line 114: Add documentation please.

src/chunkserver/hddspacemgr.cc
```diff
@@ -103,6 +103,19 @@ constexpr uint32_t startingOffsetOfBlock(uint32_t blocknum) { return blocknum *
...
+std::mutex gDisksToBeDeletedWithPendingChunksMutex;
+inline std::vector<std::pair<std::unique_ptr<IDisk>, std::vector<ChunkIdentifier>>>
```
lgsilva3087: Consider static instead of inline in .cc files.

src/slogger/slogger.cc
lgsilva3087, line 1: As we agreed, please consider to remove this commit from this PR.

lgsilva3087 (approved):
Great work. Please see my suggestions.

ralcolea (approved):
Great job @dmga44!! 👍
As we agreed, please remember to exclude the last commit with the changes in slogger and helgrind test.

src/chunkserver/hddspacemgr.cc

src/chunkserver/hddspacemgr.cc
dmga44: Added.

src/chunkserver/hddspacemgr.cc

src/slogger/slogger.cc
dmga44: Removed.

src/chunkserver/io_buffers.h
dmga44: Removed.

src/chunkserver/hddspacemgr.cc
dmga44: Added.

src/chunkserver/hddspacemgr.cc
dmga44: It should be always safe, check next comment:
```cpp
	// Delete chunks even not in AVAILABLE state here, as all threads using
	// chunk objects should already be joined (by this function and other
	// cleanup functions of other chunkserver modules that are registered on
	// eventloop termination) This function should always be executed after all
	// other chunkserver modules' (that use chunk objects) cleanup functions
	// were executed.
```

Anyway, I moved the clear down as suggested.

src/chunkserver/io_buffers.cc
dmga44: Applied here and in some other functions from the same PR.

PR #519, opened 2025-08-04
feat(cs): Release old iobuffers

dmga44:
The current behavior of the chunkserver keeps in memory at most 512MB
in the readOutputBuffersPool and the writeInputBuffersPool regardless
of whether the chunkserver is performing any operations. Plus, the
buffers used in replication (at most 64MB each of them) are thread
local, which means no releasing as well.

This change targets releasing old buffer from the IO buffers
(replicate/read/write ops). The release is performed by the
hddFreeResourcesThread and outside of the buffers pool mutexes
protection, so very low contention overhead is expected.

To make this possible, it was necessary to fit the buffer used in replications
into the same interface the other Output/InputBuffer has. This allows
get/put them to the generic buffers pool already defined.

Signed-off-by: Dave <dave@leil.io>

PR #529, opened 2025-08-14
chore: Update NEWS file for v5.1.0

uristdwarf:


rolysr (approved):
LGTM.

dmga44 (approved):
Consider to remove fixes of the new added code in this release. For instance:
- fix(cs): Fix BuffersPool release (#521) could be removed since the issue was introduced in the PR #519 (and just mention the #519 which actually did the meaningful change)
- fix: Fix Windows client build from refactor (#506) could be removed as well due to the same reason, it is a fix on top of one of the new changes, in this case a refactor.

PR #530, opened 2025-08-14
ci: Ignore non-numbered when updating NEWS file

uristdwarf:
Due to inconsistent use of merge commits, often there are duplicate
entries that say the same thing. This change will now ignore
non-numbered commits to stay consistent. However, an exception is made
if a commit is not part of a merge commit.

* Add alphabetical sorting to the entries
* Allow passing branch names to the script

### DIFF
--- a/ci/update-news.sh
+++ b/ci/update-news.sh
@@ -2,19 +2,49 @@
 
 set -eux
 
-gitChangelog=$(./ci/get-changelog.sh)
+headCommit="${2:-HEAD}"
+gitChangelog=$(./ci/get-changelog.sh "${1:-stable}" "${headCommit}")
 version=$(sed -nE 's/^set\(DEFAULT_MIN_VERSION "([^"]+)"\).*/\1/p' ./CMakeLists.txt)
 
 
 newsEntry="* SaunaFS (${version}) ($(date '+%Y-%m-%d' -u))\n"
 
-while read -r _ title message; do
+isPartOfMerge() {
+	commit="$1"
+	commit=$(git rev-parse "$commit")
+
+	if git rev-list --first-parent "$headCommit" | grep -qx "$commit"; then
+		return 1;
+	else
+		return 0;
+	fi
+}
+
+declare -a entries
+
+while read -r shortHash title message; do
 	case $title in
 		chore*|style*|refactor*|test*|"Merge"|ci*)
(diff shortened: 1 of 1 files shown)


ci/update-news.sh
```diff
@@ -2,19 +2,34 @@
...
+
+	entries+=(" - $title $message\n")
 done <<< "${gitChangelog}"
 
+IFS=$'\n' sortedEntries=($(sort <<<"${entries[*]}"))
+unset IFS
+
+for entry in "${sortedEntries[@]}"; do
+	newsEntry+="${entry}"
+done
```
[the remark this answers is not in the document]
uristdwarf: @gemini-code-assist This does not properly sort the entries

ci/update-news.sh
```diff
@@ -2,19 +2,34 @@
...
 	case $title in
 		chore*|style*|refactor*|test*|"Merge"|ci*)
 			continue
 		;;
 	esac
-	newsEntry+=" - $title $message\n"
+
+	# Only include commits that have a PR number like "(#504)"
+	if [[ ! "$title $message" =~ \(#[0-9]+\) ]]; then
+		continue
```
rolysr: I think it depends, because on the cases that someone does a `rebase and merge` and then, there could be `N` separated commits but that came from a given PR and none of them shows the PR number, they could be ignore, for example, on PR #449 and the commits 9422fb1, 0a0edc7 and 5c51a67 added to dev after merging. On the other hand there is the situation, which is probably being targeted by this commit, for example, on PR #519, which after merging to dev produced commits 29b413e (containing the `(PR #519)` reference on the commit title), plus fe1524a and 31c7443 (where this last two if combined are the same as 29b413e). Then my suggestion is to avoid ignoring commits from the first case.

uristdwarf: My general observation is that we seem to prefer merge when merging multiple commits, rather than rebase.

uristdwarf: I would like consistency is all :)

uristdwarf: I've added an exception to this

ralcolea (approved):
Nice solution @uristdwarf 
I liked the solution to exclude commits belonging to merge commits 👍

rolysr (approved):
LGTM.

PR #531, opened 2025-08-15
feat(master): Refactor the FSNode family

lgsilva3087:
The FSNode family (FSNode, FSNodeFile, FSNodeDirectory, FSNodeSymlink, and FSNodeDevice) plays a central role in the filesystem metadata. Even if they were a struct hierarchy, some of their responsibilities were implemented in unexpected or not centralized places (like concrete metadata backends).

This change aims to improve the hierarchy by:
- Introduce an ISerializable interface to unify serialization/deserialization. Previously it was done in different places (backend, metadump, etc.).
- Implement own serialization/deserialization (all data is known) and expose it, so any custom backend could use it without reimplementing.
- Add the serializedSize function, that will be useful for example for the FDB backend.
- Reduce the `switch` statements asking for custom behavior depending on the node type, which simplifies the code by using polymorphism.
- Move the constants related to header sizes to the closest class.
- Convert the structs into classes (other commits will improve the visibility).

Related changes:
- Rename struct `statsrecord` to be StatsRecord, for readability.
- Rename collections like `parent` to be `parents` for better consistency.
- Apply the code style in modified places.
- Use auto to not repeat the type in casts.

Next steps:
- Preffer dynamic_cast were considered better fit.
- Provide a toString function for better logging.

### DIFF
--- a/src/common/serializable_interface.h
+++ b/src/common/serializable_interface.h
@@ -1,0 +1,50 @@
+/*
+   Copyright 2023      Leil Storage OÜ
+
+   This file is part of SaunaFS.
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "common/platform.h"
+
+#include <cstddef>
+#include <cstdint>
+
+/// Interface for serializable objects.
+class ISerializable {
+public:
+	// Default constructor
+	ISerializable() = default;
+
+	// virtual destructor for polymorphic deletion
+	virtual ~ISerializable() = default;
+
+	ISerializable(const ISerializable &) = delete;
+	ISerializable &operator=(const ISerializable &) = delete;
+	ISerializable(ISerializable &&) = delete;
+	ISerializable &operator=(ISerializable &&) = delete;
+
+	/// Returns the size of the serialized object in bytes.
+	virtual size_t serializedSize() const = 0;
+
+	/// Serializes the object into the provided destination buffer.
+	/// The function assumes that the destination buffer has enough space allocated.
+	/// Use serializedSize in the caller to ensure sufficient space.
+	virtual void serialize(uint8_t **destination) const = 0;
+
+	/// Deserializes the object from the provided source buffer.
+	virtual void deserialize(const uint8_t **source) = 0;
+};
--- a/src/master/filesystem_node_types.h
+++ b/src/master/filesystem_node_types.h
@@ -23,11 +23,16 @@
 #include "common/platform.h"
 
 #include <array>
+#include <cstddef>
 #include <cstdint>
+#include <cstring>
+#include <string>
 #include <unordered_map>
 
-#include "common/goal.h"
 #include "common/compact_vector.h"
+#include "common/datapack.h"
+#include "common/goal.h"
+#include "common/serializable_interface.h"
 #include "common/type_defs.h"
 #include "protocol/SFSCommunication.h"
 
@@ -88,7 +93,7 @@ using TrashtimeMap = std::unordered_map<uint32_t, uint32_t>;
 using GoalStatistics = std::array<inode_t, GoalId::kMax + 1>;
 using ParentsCompactVector = compact_vector<std::pair<inode_t, const hstorage::Handle *>, uint32_t>;
 
-struct statsrecord {
+struct StatsRecord {
 	inode_t inodes;
 	inode_t dirs;
 	inode_t files;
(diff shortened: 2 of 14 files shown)


src/master/filesystem_periodic.cc
```diff
@@ -432,13 +430,13 @@ void fs_process_file_test() {
...
 						auto parentInChildPtr = std::find_if(
-						    childNode->parent.begin(), childNode->parent.end(),
+						    childNode->parents.begin(), childNode->parents.end(),
 						    [node](const std::pair<inode_t, const hstorage::Handle *> &p) {
 							    return p.first == node->id;
 						    });
 						// the node doesn't have a parent entry pointing to the
 						// current directory
-						if (parentInChildPtr == node->parent.end()) {
+						if (parentInChildPtr == node->parents.end()) {
```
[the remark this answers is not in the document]
lgsilva3087: @gemini-code-assist nice catch, fixed.

src/common/serializable_interface.h
```diff
@@ -0,0 +1,47 @@
...
+	ISerializable &operator=(ISerializable &&) = delete;
+
+	/// Returns the size of the serialized object in bytes.
+	virtual size_t serializedSize() const = 0;
+
+	/// Serializes the object into the provided destination buffer.
+	virtual void serialize(uint8_t **destination) const = 0;
+	/// Deserializes the object from the provided source buffer.
+	virtual void deserialize(const uint8_t **source) = 0;
+};
```
dmga44: Please add the trailing newline.

src/master/metadata_backend_file.cc
```diff
@@ -550,76 +551,41 @@ static int8_t fs_parseNode(const std::shared_ptr<MemoryMappedFile> &metadataFile
...
-	constexpr uint32_t kChunkSize = (1 << 16);
+	// Deserialize the node
+	node->deserialize(&pSrc);
+
+	// Update the offset for next nodes
+	sectionOffset = metadataFile->offset(pSrc);
+
+#ifndef METARESTORE
+	auto *nodeFile = static_cast<FSNodeFile *>(node);
+#endif
```
dmga44, line 566: Why not move this piece of code to the case where it is only used? (We could also join the #ifndef directives)

src/common/serializable_interface.h
uristdwarf: File ending

src/master/filesystem_node_types.h
```diff
@@ -123,29 +129,33 @@ enum class FSNodeType : std::uint8_t {
...
+public:
+	inode_t id{};       ///< Unique number identifying node.
+	uint32_t ctime{};   ///< Change time.
+	uint32_t mtime{};   ///< Modification time.
+	uint32_t atime{};   ///< Access time.
+	FSNodeType type{};  ///< Node type. (file, directory, symlink, ...)
+	uint8_t goal{};     ///< Goal id.
+
+	/// Only 12 lowest bits are used for mode, in unix standard upper 4 are used for object type,
+	/// but since there is field "type" this bits can be used as extra flags.
```
uristdwarf: nit: this > these

src/master/filesystem_node_types.h
```diff
@@ -159,55 +169,205 @@ struct FSNode {
...
+		put8bit(destination, static_cast<uint8_t>(type));
+		putINode(destination, id);
+		put8bit(destination, goal);
+		put16bit(destination, mode);
+		put32bit(destination, uid);
+		put32bit(destination, gid);
+		put32bit(destination, atime);
+		put32bit(destination, mtime);
+		put32bit(destination, ctime);
+		put32bit(destination, trashtime);
```
uristdwarf, line 195: Not necessarily related to this PR, but something to think about with the new master:

Using 32-bit integers means we (or someone) will run into the year 2038 problem when that happens. Maybe consider changing this potentially? Migrating should be easy, as it's a simple cast of 32 to 64 bit integers.

Also,  on Linux and probably other POSIX systems, the time can be negative, so is using unsigned a good idea here?

src/master/filesystem_node_types.h
```diff
@@ -159,55 +169,205 @@ struct FSNode {
...
+	void serialize(uint8_t **destination) const override {
+		FSNode::serialize(destination);
+
+		put64bit(destination, length);
+		auto chunkAmount = chunkCount();
+		put32bit(destination, chunkAmount);
+		auto sessionsCount = std::min<int>(sessionIds.size(), kMaxSessionSize);
+		put16bit(destination, sessionsCount);
+
+		for (size_t i = 0; i < chunkAmount; ++i) { put64bit(destination, chunks[i]); }
```
dmga44, line 282: Please note there is a significant difference here from the original dump code.

Previously, it was done in batches of at most 65536 chunks at a time, which is why it fit in the intermediate buffer (`gNodeStoreBuffer`). Here, you put everything at once (at most 2^32 chunks from my understanding).

Consider adding some kind of rework here, in `MetadataBackendFile::storenode` could be something like:
```cpp
while (f->pendingDataToSerialize > 0)
    serialize some data
```

src/master/filesystem_node_types.h
```diff
@@ -159,55 +169,205 @@ struct FSNode {
...
+	/// Returns the size of the serialized object in bytes.
+	size_t serializedSize() const override { return kSymlinkHeaderSize + path_length; }
+
+	/// Serializes the object into the provided destination buffer.
+	void serialize(uint8_t **destination) const override {
+		FSNode::serialize(destination);
+		put32bit(destination, path_length);
+		auto name = static_cast<std::string>(path);
+		::memcpy(*destination, name.data(), name.length());
+		std::advance(*destination, static_cast<std::ptrdiff_t>(name.length()));
```
uristdwarf: Why not create a helper function for this, like put32bit?

Of course, remember that in our network protocol we have two string types, one from MooseFS (no NULL character at end) and LizardFS (NULL character at end)

Speaking of which, does the metadata have null ptr or not at the end of string?

dmga44 (commented):
Great job, @lgsilva3087 , but please check one of my comments, I think it is important.

src/master/filesystem_node_types.h
```diff
@@ -159,55 +169,205 @@ struct FSNode {
...
+		FSNode::deserialize(source);
+		uint32_t pathLengthTemp{};  // For some reason it is stored as uint32_t
+		get32bit(source, pathLengthTemp);
+		path_length = static_cast<decltype(path_length)>(pathLengthTemp);
+
+		if (path_length > 0) {
+			const uint8_t* start = *source;
+			const uint8_t* end = std::next(start, static_cast<std::ptrdiff_t>(pathLengthTemp));
+			path = HString(start, end);
+			*source = end;
```
uristdwarf: Same here, why not a string deserializer helper?

uristdwarf (commented):
Do we have any tests for serializing and deserializing these types?

src/master/filesystem_node_types.h
dmga44: In that case, we need a `HString` which is a string with a precomputed hash.

src/master/filesystem_node_types.h
uristdwarf: Ah nevermind, I forgot that unsigned is a year 2106 problem: https://en.wikipedia.org/wiki/Time_formatting_and_storage_bugs#Year_2106

So maybe not that big of a deal, but I wonder how we are converting to Linux time?

src/master/filesystem_node_types.h
uristdwarf: Maybe a serializer/deserializer for HString?

src/master/metadata_backend_file.cc
lgsilva3087: It gives build errors when I tried to move it inside the switch statements.

src/common/serializable_interface.h
lgsilva3087: Sure, a classic one.

lgsilva3087: Sure.

src/master/filesystem_node_types.h
lgsilva3087: Please remind me in 2105 ;).

src/master/filesystem_node_types.h
lgsilva3087: Applied.

src/master/filesystem_node_types.h
lgsilva3087: Because the serialization functions are in common and HString in master. Same with data pack.

It doesn't use null terminators for the metadata.

HStrings should not be needed in the distributed master.

I tried different variants, some of them really ugly. The less ugly to me was to add static `serialize` and `deserialize` to the HString class. Then the calls are like this:
```cpp
HString::serialize(destination, path.get());
HString::deserialize(source, pathString);
```

src/master/filesystem_node_types.h
lgsilva3087: Please see the other related comment.
The hash does not need to be serialized.

src/master/filesystem_node_types.h
lgsilva3087: Nice catch, I will work on this one. Thanks.

lgsilva3087: Done.

src/master/metadata_backend_file.cc
```diff
@@ -1004,121 +970,95 @@ void MetadataBackendFile::loadall(int ignoreflag) {
...
+		uint32_t includedSessions = 0;
+
+		for (const auto &sid : fileNode->sessionIds) {
+			if (includedSessions >= FSNodeFile::kMaxSessionSize) {
 				break;
 			}
 			put32bit(&chptr, sid);
-			sessionids++;
+			includedSessions++;
 		}
```
dmga44, line 1052: Consider, maybe for the next iteration, to store the sessions in a similar fashion the chunks are being stored.

Though this is a small improvement/change could be useful when many client access the same file.

dmga44 (approved):
LGTM, great job.

uristdwarf (approved):

PR #538, opened 2025-08-19
fix(chunkserver): Don't clear chunkserver output packets

uristdwarf:
When the master connection was closed (for example, when it's demoting), it could cause inconsistent state with the master. If a chunkserver received a packet, to create a chunk for example, it will finish it but not reply to master. This could lead to lost data if there were active writes happening at the same time.

This fix ensures that when chunkserver reconnects with master, the chunkserver will always send the register packet first, and then send the remaining output packets still remained in the output packets when the disconnect happened.

### DIFF
--- a/src/chunkserver/master_connection.cc
+++ b/src/chunkserver/master_connection.cc
@@ -688,7 +688,6 @@ void MasterConn::releaseResources() {
 	}
 }
 
-void MasterConn::resetPackets() {
+void MasterConn::resetInputPackets() {
 	inputPacket_.reset();
-	outputPackets_.clear();
 }
--- a/src/chunkserver/master_connection.h
+++ b/src/chunkserver/master_connection.h
@@ -158,7 +158,7 @@ public:
 
 	void releaseResources();
 
-	void resetPackets();
+	void resetInputPackets();
 
 	// Inline getters and setters
 
--- a/src/chunkserver/masterconn.cc
+++ b/src/chunkserver/masterconn.cc
@@ -189,11 +189,11 @@ void masterconn_send_status() {
 
 void masterconn_serve(const std::vector<pollfd> &pdesc) {
 	LOG_AVG_TILL_END_OF_SCOPE0("master_serve");
-	
+
 	MasterConn *eptr = gMasterConnSingleton.get();
 
 	eptr->handlePollErrors(pdesc);
-	
+
 	// Check if there are any background jobs to process.
 	if (eptr->mode() == ConnectionMode::CONNECTED) {
 		if (gJobFDpDescPos >= 0 && (pdesc[gJobFDpDescPos].revents & POLLIN)) {
@@ -219,7 +219,7 @@ void masterconn_serve(const std::vector<pollfd> &pdesc) {
 		gJobPool->disableAndChangeCallbackAll(masterconn_unwantedjobfinished);
 		gReplicationJobPool->disableAndChangeCallbackAll(masterconn_unwantedjobfinished);
 		tcpclose(eptr->socketFD());
-		eptr->resetPackets();
+		eptr->resetInputPackets();
 		eptr->setMode(ConnectionMode::FREE);
 	}
 }


src/chunkserver/master_connection.cc
```diff
@@ -690,5 +690,4 @@ void MasterConn::releaseResources() {
 
 void MasterConn::resetPackets() {
 	inputPacket_.reset();
-	outputPackets_.clear();
 }
```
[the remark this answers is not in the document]
uristdwarf: Done

src/chunkserver/master_connection.cc
```diff
@@ -688,7 +688,6 @@ void MasterConn::releaseResources() {
 	}
 }
 
-void MasterConn::resetPackets() {
+void MasterConn::resetInputPackets() {
 	inputPacket_.reset();
-	outputPackets_.clear();
```
lgsilva3087, line 693: My only concern is this case for example:
- The Chunkserver gets disconnected (connection is set to KILL), but we don't clear the existing output packets.
- After some time, let's say 1h, connection is restored and the chunkserver sends the old output packets from 1h ago.

Maybe we could expire them after some time, let's say 10 seconds to avoid those situations.

lgsilva3087 (commented):
Please see my comment.

src/chunkserver/master_connection.cc
uristdwarf: I think we discussed that in this situation it should be fine, the chunkserver will report the updated state it's in and that master could handle them clearly. It could be 1 day and I don't think it would be a problem.

uristdwarf: Master doesn't wait for a response usually, it sends the packets off and then later reads the response from chunkserver. Look at the src/master/matocsserv.cc send functions for examples.

uristdwarf: At a minimum, it should be whatever the timeout for the clients have for writing/reading files, but since this can vary between clients, it isn't a reliable indicator

lgsilva3087: We discussed it on Monday, and we will discuss it as many times as needed.

For example, in the replication workflow there are two main functions:
matocsserv_send_sau_replicatechunk
  calls matocsserv_replication_begin

matocsserv_got_replicatechunk_status
  calls matocsserv_replication_end

I understand Master is not waiting explicitely for the answer, but there are actions on packet received (like matocsserv_replication_end) that should be idempotent. All these types of functions should be checked (looking into the code or better with tests) to check that the new behavior does not introduce problems.

Another scenario is at failover with uRaft, then the promoted shadow will receive the pending packets, maybe it is OK, but we need to test it.

How are you reproducing these disconnection scenarios?

My other concern is:
On disconnection, Chunkservers needs to be registered again to the MDSs, and this is done using a packet in the outputPackets (which should be the first one).

How do you ensure that the registration packet is processed before the pending outputPackets?

uristdwarf: > Another scenario is at failover with uRaft, then the promoted shadow will receive the pending packets, maybe it is OK, but we need to test it.

This was the main reason for the change. It's precisely because of shadow not being able to complete which what led to chunk loss and chunks not being able to be written to chunkserver by client. I've tested it manually. There's no automated test, because I can't reproduce this in our automated tests, But both Valga clusters exhibit this behavior when active writes are occurring.

> On disconnection, Chunkservers needs to be registered again to the MDSs, and this is done using a packet in the outputPackets (which should be the first one).

> How do you ensure that the registration packet is processed before the pending outputPackets?

Good point. I need to correct myself (and the PR). It doesn't seem to be the case that it's sending the registration packet. But master is still accepting the older packets. I'll have a look why exactly that's the case. And maybe ensure it's the first output packet.

> matocsserv_send_sau_replicatechunk
calls matocsserv_replication_begin

> matocsserv_got_replicatechunk_status
calls matocsserv_replication_end

OK, I'll check these.

PR #546, opened 2025-08-25
fix(cs): Fix release reads

dmga44:
It has being noticed that in some read workloads the memory used by the
chunkserver can increase steadily over time. The issue is related to
the proper release of the structures related to the reads in the
chunkserver side. This change targets fixing that issue and some other
inconsistencies found while studying the issue.

The main change that fixed the issue was removing the actually disabled
jobs in the jobs pool from the ones put in the discarded read jobs
container. Those jobs were not setting up the status of the related
OutputBuffer and thus, were getting stuck the releasing mechanism of
that structure and the whole csentry.

There were a couple of fixes applied along the way:
- in the function changeCallback, while called upon a list of jobs, the
std::move of the callback provided was substituted by a direct copy.
- in the function readDiscardCallback make sure there are no jobs with
status after the cycle.

Beside these fixes, the other important change is to simplify a bit the
closing scheme of the csentry.

This change extends the sfschunkserver_check_no_buffer_in_use over many
tests which did not have it. All of those tests belong to
SingleMachineTests suite. Also extends the test_concurrent_replications
test to make it fail in some cases without the proposed fixes.


src/chunkserver/chunkserver_entry.cc
```diff
@@ -349,12 +337,17 @@ void ChunkserverEntry::readDiscardCallback(uint8_t status, void *entry) {
 	(void) status;
 	auto *eptr = static_cast<ChunkserverEntry *>(entry);
 
-	while (!eptr->toDiscardReadDataPackets.empty() &&
-	       eptr->toDiscardReadDataPackets.front()->outputBuffer->getStatus() != kNotSaunafsStatus) {
-		getReadOutputBufferPool().put(
-		    std::move(eptr->toDiscardReadDataPackets.front()->outputBuffer));
-		eptr->toDiscardReadDataPackets.pop_front();
-		eptr->toDiscardReadJobIds.pop_front();
+	auto jobsIt = eptr->toDiscardReadJobIds.begin();
```
lgsilva3087, line 343: Consider:
assert(toDiscardReadJobIds.size() == toDiscardReadDataPackets.size());

And maybe a comment reminding that both collections are lockstep.

src/chunkserver/hddspacemgr.cc
```diff
@@ -735,10 +735,11 @@ static void hddReadAheadAndBehind(IChunk *chunk, uint16_t block,
 		chunk->owner()->prefetchChunkBlocks(
 		    *chunk, firstBlockToRead,
 		    blocksToBeReadAhead + block - firstBlockToRead);
-		OutputBuffer buffer = OutputBuffer(kIgnoreHeaderSize, block - firstBlockToRead);
+		auto buffer = getReadOutputBufferPool().get(kIgnoreHeaderSize, block - firstBlockToRead);
 		for (uint16_t b = firstBlockToRead; b < block; ++b) {
-			hddReadCrcAndBlock(chunk, b, &buffer, false);
+			hddReadCrcAndBlock(chunk, b, buffer.get(), false);
 		}
+		getReadOutputBufferPool().put(std::move(buffer));
```
lgsilva3087, line 742: Nice touch. I like that you are making it consistent with the pool.

tests/tools/saunafs.sh
```diff
@@ -28,6 +28,7 @@ setup_local_empty_saunafs() {
 	local group_id=${GROUP_ID:=$(id -g)}
 	declare -gA saunafs_info_
 	saunafs_info_[chunkserver_count]=$number_of_chunkservers
+	saunafs_info_[mount_count]=$number_of_mounts
```
lgsilva3087: Remember to remove this if unused.

lgsilva3087 (approved):
Great complex job. Please see my minor comments.

src/chunkserver/chunkserver_entry.cc
dmga44: Added, plus comment.

tests/tools/saunafs.sh
dmga44: Removed.

tests/test_suites/SingleMachineTests/test_pread_eio_in_chunkserver_single_disk.sh
```diff
@@ -48,3 +48,5 @@ for f in test/*; do
 	assert_eventually_prints "" "saunafs fileinfo '$f' | grep ':${info[chunkserver0_port]}'"
 	assert_eventually_prints 2 "saunafs fileinfo '$f' | grep copy | wc -l"
 done
+
+sfschunkserver_check_no_buffer_in_use
```
ralcolea, line 51: After this change, this test fails by timeout in my laptop.
I suggest you set the same time for Linux and Windows of 1 minute.

tests/test_suites/SingleMachineTests/test_pread_eio_nonheader_in_chunkserver.sh
```diff
@@ -55,3 +55,5 @@ for f in goal2/*; do
 	assert_eventually_prints "" "saunafs fileinfo '$f' | grep ':${info[chunkserver0_port]}'"
 	assert_eventually_prints 2 "saunafs fileinfo '$f' | grep copy | wc -l"
 done
+
+sfschunkserver_check_no_buffer_in_use
```
ralcolea, line 61: Same comment for this test, despite the timeout is more tied. In my laptop passes in 32 seconds.
I think timeout can be set between 45 seconds to 1 minute.

ralcolea (approved):
Great job @dmga44 👍 🔥 💪 
I really appreciate this PR, not only for the solid fixes, but also for making several functions cleaner and reducing their responsibilities. Please, see some minor suggestions I shared for your consideration.

tests/test_suites/SingleMachineTests/test_pread_eio_in_chunkserver_single_disk.sh
dmga44: Thanks, set timeout to 1 minute.

tests/test_suites/SingleMachineTests/test_pread_eio_nonheader_in_chunkserver.sh
dmga44: Thanks, set timeout to 1 minute.

PR #567, opened 2025-09-04
fix(metarestore): Exclude unit tests from master

lgsilva3087:
Avoids including unit tests files from master that could match the criteria for the blobs. For instance:

filesystem_node_types_unittest.cc

Co-authored-by: Urmas Rist <urmas@leil.io>


uristdwarf (approved):
nit: Should be build(metarestore) instead of fix(metarestore)

ralcolea (approved):
It looks nice :)

PR #571, opened 2025-09-08
fix(vcpkg): Make baseline update optional

wg1k:
The vcpkg_install.sh script is intended to pin the package versions for achieving reproducible environments. It also sets the version baseline for packaging. However, updating the baseline each time could potentially break builds.

This commit allows safer execution of the script by making baseline updates optional and disabled by default.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -8,8 +8,8 @@
     "boost-system",
     "fmt",
     "gtest",
-    "spdlog",
-    "prometheus-cpp"
+    "prometheus-cpp",
+    "spdlog"
   ],
   "overrides": [
     {
@@ -285,6 +285,10 @@
       "version": "1.0.8"
     },
     {
+      "name": "civetweb",
+      "version": "1.16"
+    },
+    {
       "name": "fmt",
       "version": "11.0.2"
     },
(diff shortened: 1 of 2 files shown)


uristdwarf (approved):

vcpkg.json
```diff
@@ -284,6 +284,10 @@
       "name": "bzip2",
       "version": "1.0.8"
     },
+    {
```
ralcolea, line 287: This new dependence is not referenced in the commit message.
Was it intentionally added? Looks like not related with the goal of the PR.
What is it used for?

ralcolea (approved):
Good job @ictus4u 
Please check my question about the new dependence. 
If it's really needed, please consider explaining in the commit message the reason motivating its inclusion.

vcpkg.json
wg1k: This is a pre-existing dependency that I took the opportunity to formalize in this change. `civetweb` is a required dependency of `prometheus-cpp`; it is installed by vcpkg regardless. By adding it explicitly to the vcpkg.json file, we enable version pinning (and complete dependency list). I included it in this PR, as collateral to fixing the script that does the pinning, to prevent it from appearing as an unrelated change in a future commit.

wg1k: [Commit](https://github.com/leil-io/saunafs/commit/216c89ac37de615a878e2c7b2c5c27fdfccf929e) message updated.

PR #577, opened 2025-09-11
perf: Defer metadata dumping during failover

ralcolea:
This PR introduces deferred metadata dumping in SaunaFS master to improve failover recovery time and cluster availability.

Previously, when a shadow was promoted to master, it had to perform a full metadata dump before accepting connections from chunkservers. This introduced unnecessary delays, as the cluster could not serve I/O until the dump completed.

**Summary of changes:**

- **DeferredMetadataDumpTask**
  - A new task class that runs the metadata dump in the background.
  - Logs detailed outcomes and gracefully handles cases where a periodic dump is already in progress.

- **Master failover behavior**
  - Added -o defer-metadata-dump option to master startup.
  - When enabled, metadata dump is scheduled asynchronously via TaskManager (with a 30s delay) instead of being performed immediately.
  - This allows chunkservers to reconnect and I/O operations to resume sooner.

- **uraft helper update**
  - Failover path (saunafs-uraft-helper) now uses `-o defer-metadata-dump` by default when promoting a shadow to master.

- **New integration test** (test_deferred_metadata_dump.sh)
  - Validates cluster behavior during a deferred metadata dump.
  - Ensures logs reflect scheduling and successful completion.

Signed-off-by: Crash <<crash@leil.io>>

### DIFF
--- a/src/master/deferred_metadata_dump_task.cc
+++ b/src/master/deferred_metadata_dump_task.cc
@@ -51,5 +51,13 @@ int DeferredMetadataDumpTask::execute(uint32_t /*timeStamp*/, intrusive_list<Tas
 }
 
 bool DeferredMetadataDumpTask::isFinished() const {
+	// This task is executed only once and triggers a background dump that relies on fork.
+	// Unlike other tasks (e.g., SetGoalTask, SetTrashtimeTask) that iterate through collections
+	// and track progress via iterators, this task's responsibility ends immediately after
+	// calling fs_storeall(DumpType::kBackgroundDump).
+	//
+	// The actual metadata dumping happens asynchronously in a forked child process,
+	// which is monitored separately through the metadata dumper's polling mechanism.
+	// Therefore, the task is considered finished as soon as fs_storeall() returns.
 	return true;  // Single execution task
 }
--- a/src/master/deferred_metadata_dump_task.cc
+++ b/src/master/deferred_metadata_dump_task.cc
@@ -1,0 +1,58 @@
+/*
+   Copyright 2025      Leil Storage OÜ
+
+   This file is part of SaunaFS.
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "master/deferred_metadata_dump_task.h"
+
+#include "errors/saunafs_error_codes.h"
+#include "master/metadata_dumper_interface.h"
+#include "slogger/slogger.h"
+
+DeferredMetadataDumpTask::DeferredMetadataDumpTask(IMetadataBackend *backend) : backend_(backend) {}
+
+int DeferredMetadataDumpTask::execute(uint32_t timeStamp, intrusive_list<Task> &work_queue) {
+	(void)timeStamp;   // Suppress unused parameter warning
+	(void)work_queue;  // Task doesn't create subtasks
+
+	safs::log_info("Executing deferred metadata dump");
+
+	try {
+		uint8_t result = backend_->fs_storeall(DumpType::kBackgroundDump);
+
+		if (result == SAUNAFS_ERROR_TEMP_NOTPOSSIBLE) {
+			safs::log_info(
+			    "Deferred metadata dump skipped due to periodic dump in progress - "
+			    "metadata will be captured in next periodic dump");
+			return SAUNAFS_STATUS_OK;  // Treat as success
+		}
+
+		if (result == SAUNAFS_STATUS_OK) {
+			safs::log_info("Deferred metadata dump completed successfully");
+			return SAUNAFS_STATUS_OK;
+		}
+
+		safs::log_err("Deferred metadata dump failed with result: {}", result);
+		return result;
+	} catch (const std::exception &e) {
+		safs::log_err("Deferred metadata dump failed with exception: {}", e.what());
+		return SAUNAFS_ERROR_IO;
+	}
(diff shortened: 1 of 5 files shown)


src/master/deferred_metadata_dump_task.cc
```diff
@@ -0,0 +1,58 @@
...
+
+#include "errors/saunafs_error_codes.h"
+#include "master/metadata_dumper_interface.h"
+#include "slogger/slogger.h"
+
+DeferredMetadataDumpTask::DeferredMetadataDumpTask(IMetadataBackend *backend) : backend_(backend) {}
+
+int DeferredMetadataDumpTask::execute(uint32_t timeStamp, intrusive_list<Task> &work_queue) {
+	(void)timeStamp;   // Suppress unused parameter warning
+	(void)work_queue;  // Task doesn't create subtasks
```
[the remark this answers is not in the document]
ralcolea: Nice catch @copilot :+1: 
Applied.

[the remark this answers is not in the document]
ralcolea: I commented the parameter names, such that other developers can see the names of the expected parameters.

src/master/deferred_metadata_dump_task.cc
```diff
@@ -0,0 +1,58 @@
...
+		return result;
+	} catch (const std::exception &e) {
+		safs::log_err("Deferred metadata dump failed with exception: {}", e.what());
+		return SAUNAFS_ERROR_IO;
+	}
+}
+
+bool DeferredMetadataDumpTask::isFinished() const {
+	return true;  // Single execution task
+}
```
rolysr, line 63: Consider to use a private boolean variable from the `DeferredMetadataDumpTask` in order to denote that task is finished, so that it can be set when the task is really finished. In case is not needed, please explain the idea of why it is enough to always return true on this concrete implementation of `isFinished` method.

rolysr (approved):
Very good addition @ralcolea. LGTM.

tests/test_suites/ShortSystemTests/test_deferred_metadata_dump.sh
uristdwarf, line 1: To be honest, I don't like this "test by logging". It's not really testing behavior, but checking if a log exists, which can changed/be moved/removed, all of which can cause unexpected failures.

Of course, I can't think of a better alternative to test right now. But I would urge caution in using logs to test behavior. It's arguable if it's better than no test at all, but if it is, it's only marginal.

uristdwarf: What I would do perhaps is check if the metadata file is updated after 30 seconds instead of immediately. Of course, be careful with the timing. If the test runs around a full hour, it may cause a accidental dump

src/master/filesystem.cc
```diff
@@ -249,9 +251,29 @@ int fs_loadall(void) {
...
 		safs::log_info("all needed changelogs applied successfully");
 
-		// Dump the new metadata
-		gMetadataBackend->fs_storeall(DumpType::kForegroundDump);
-		safs::log_info("Metadata dumped successfully after applying changelogs");
+		if (deferDump) {
+			safs::log_info("Deferring metadata dump option detected");
+
+			// Schedule one-time deferred metadata dump task using TaskManager
+			uint32_t delayedTime = eventloop_time() + 30;  // 30 seconds delay
```
uristdwarf: eventloop_time is based on system clock. Strange things may happen if it jumps forward or back.

src/master/filesystem.cc
```diff
@@ -249,9 +251,29 @@ int fs_loadall(void) {
...
+
+			auto *metadataBackendPtr = gMetadataBackend.get();
```
uristdwarf: I don't think this is a problem, but double check that metadataBackenPtr is either a smart pointer or is valid when this function calls and remains valid until the async task finishes. Maybe cancel the task when master is shutting down?

uristdwarf (changes requested):
Main concern is the filesystem.cc changes. The test could be done better IMHO, but it's not a blocker.

src/master/deferred_metadata_dump_task.cc
ralcolea: I added an additional explanation in the function to clarify why returning `true` is sufficient, without introducing extra variables.

src/master/filesystem.cc
ralcolea: Actually, this was not affecting the behavior of the task because `delayedTime` value is ignored in the `execute()` function (`timestamp` attribute).

I removed the dependence on `eventloop_time()` function.
However, this is not the first task relying on `eventloop_time()` function. Most of the existing ones do :)

src/master/filesystem.cc
ralcolea: `metadataBackenPtr` is a raw pointer returned by `gMetadataBackend` (smart pointer).
The new logic is equivalent to the previous one, only the function is called in another class (`DeferredMetadataDumpTask::execute()`).

tests/test_suites/ShortSystemTests/test_deferred_metadata_dump.sh
ralcolea: Testing this feature with the restrictions of the testing environment can be tricky :)
However, I added an additional validation to confirm that the `metadata.sfs` file was updated by the deferred metadata dump.

src/master/filesystem.cc
```diff
@@ -249,9 +251,27 @@ int fs_loadall(void) {
...
+			// Schedule one-time deferred metadata dump task using TaskManager
+			gMetadata->taskManager.submitTask(
+			    0, 1, metadataDumpTask.release(), DeferredMetadataDumpTask::generateDescription(),
+			    [](int status) {
+				    if (status == SAUNAFS_STATUS_OK) {
+					    safs::log_info("Deferred metadata dump completed successfully");
+				    } else {
+					    safs::log_err("Deferred metadata dump failed with status: {}", status);
+				    }
+			    });
```
wg1k: Just a style sugetion: could this code be refactored to a separate function? Just for keeping the `fs_loadall` function readable.

wg1k (approved):

uristdwarf (approved):

src/master/filesystem.cc
ralcolea: Nice suggestion :)
Applied.

PR #578, opened 2025-09-12
fix: Corrected Windows specific error messages

rolysr:
This commit resolves two issues:
- Error codes returned by Windows libraries (e.g., winsock2.h) could have identifiers not present in the Linux error number set, resulting in incorrect error messages when handling Windows-only errors. The error message handling has been updated to properly display Windows-specific error codes.

- Fixed an issue in ServerConnection::connect where the last TCP error was requested after the socket was closed, causing incorrect error messages to be shown. The error retrieval now occurs before socket closure to ensure accurate reporting. This fixed problem was specially present when catching Windows socket library errors.

### DIFF
--- a/src/common/server_connection.cc
+++ b/src/common/server_connection.cc
@@ -140,18 +140,20 @@ MessageBuffer ServerConnection::sendAndReceive(
 	return receiveRequestGeneric(fd, expectedType, receiveMode, timeout);
 }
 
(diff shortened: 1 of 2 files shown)


ralcolea (approved):
Nice catch @rolysr! 👍

dmga44 (approved):

PR #580, opened 2025-09-16
fix: Fix test for client retry tuning

rolysr:
Previous version of test for retry time and sleep time divisor parametized options was flaky due to no catching correct output from ls -l command and not waiting as much as expected to make the command output the expected value. This commit fixes those issues.

### DIFF
--- a/tests/test_suites/ShortSystemTests/test_wait_retry_time_and_sleep_time_divisor.sh
+++ b/tests/test_suites/ShortSystemTests/test_wait_retry_time_and_sleep_time_divisor.sh
@@ -1,13 +1,41 @@
-timeout_set "5 minutes"
+timeout_set "4 minutes"
 
 run_ls_background() {
-    local suffix="$1"
+	local suffix="$1"
 
-    # start the ls command in the background and redirect
-    # its output to a file named ls_output_<suffix>.txt
-    (
-      ls -l "${info[mount0]}" > "/tmp/ls_output_${suffix}.txt" 2>&1
-    ) &
+	# start the ls command in the background and redirect
+	# its output to a file named ls_output_<suffix>.txt
+	(
+		ls -l "${info[mount0]}" > "$TEMP_DIR/ls_output_${suffix}.txt" 2>&1
+	) &
+}
+
+# This function tests client retry behavior when the master server is temporarily unavailable.
+# It works as follows:
+# 1. Stops the master server.
+# 2. Starts an 'ls' command in the background on the client, which will try to access the mount point.
+# 3. Waits for a specified sleep time, allowing the client to enter retry mode.
+# 4. Restarts the master server and waits for chunkservers to be ready.
+# 5. Checks the result:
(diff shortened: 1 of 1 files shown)


tests/test_suites/ShortSystemTests/test_wait_retry_time_and_sleep_time_divisor.sh
```diff
@@ -1,12 +1,12 @@
...
+timeout_set "4 minutes"
 
 run_ls_background() {
     local suffix="$1"
 
     # start the ls command in the background and redirect
     # its output to a file named ls_output_<suffix>.txt
     (
-      ls -l "${info[mount0]}" > "/tmp/ls_output_${suffix}.txt" 2>&1
+      ls -l "${info[mount0]}" > "$TEMP_DIR/ls_output_${suffix}.txt" 2>&1
```
ralcolea: Consider using tabs for indentation in `run_ls_background()` function.

tests/test_suites/ShortSystemTests/test_wait_retry_time_and_sleep_time_divisor.sh
```diff
@@ -40,7 +40,7 @@ saunafs_master_daemon start
 saunafs_wait_for_ready_chunkservers 1
 
 # at this time, the ls command should have failed
-assert_success grep -q "Input/output error" "/tmp/ls_output_1.txt"
+assert_failure grep -q "total" "$TEMP_DIR/ls_output_1.txt"
```
ralcolea: In the test there is a general pattern that can be refactored within a function receiving the parameters needed, i.e., `validate_background_ls X Y validate`.
```
run_ls_background "X"

sleep Y
saunafs_master_daemon start
saunafs_wait_for_ready_chunkservers 1

assert_failure grep -q "total" "$TEMP_DIR/ls_output_{X}.txt" or
assert_success grep -q "total" "$TEMP_DIR/ls_output_{X}.txt"
```

This will reduce the length of the test and will make it easier to maintain. Consider this suggestion as an improvement, maybe not related with the original goal of this PR.

tests/test_suites/ShortSystemTests/test_wait_retry_time_and_sleep_time_divisor.sh
```diff
@@ -93,12 +93,12 @@ run_ls_background "3"
 # each failing operation. Then, for this case, we need to wait
 # for more than 30 seconds to check that the ls command
 # could not finish before the master server is back online.
-sleep 32
+sleep 35
```
ralcolea: Is it possible to reduce the sleep times by reducing the number of retries in the client?
If it's possible, it will reduce the overall time for the test.

ralcolea (approved):
Nice work @rolysr.
I have shared a few minor suggestions for your consideration, but none of them are blocking. From my side, this is good to merge.

uristdwarf (approved):

PR #582, opened 2025-09-22
fix: Resolve compiler errors and improve code quality

wg1k:
This PR addresses several issues found during compilation and improves code reliability.

Changes:

fix(uraft): Resolve compiler error in jthread initialization

- Replaced member function pointer with lambda to capture 'this' correctly
- Properly passes std::stop_token to eventListenerThread method
- Fixes incompatible syntax with jthread's stop_token requirements

fix(memory_info): Standardize Linux platform detection

- Replaced non-standard `linux` macro with `__linux__`
- Ensures reliable platform detection across different compilers and environments
- Improves conditional compilation correctness

refactor(admin): Use shared pointers for command management

- Updated admin tool to manage commands using shared pointers
- Reduces risk of memory leaks by ensuring proper command object lifetime
- Improves memory safety and resource management

### DIFF
--- a/src/admin/main.cc
+++ b/src/admin/main.cc
@@ -25,15 +25,18 @@
 #include <ostream>
 
 #include "admin/chunk_health_command.h"
+#include "admin/delete_sessions_command.h"
+#include "admin/dump_config_command.h"
 #include "admin/info_command.h"
 #include "admin/io_limits_status_command.h"
 #include "admin/list_chunkservers_command.h"
 #include "admin/list_defective_files_command.h"
-#include "admin/list_disks_command.h"
 #include "admin/list_disk_groups_command.h"
+#include "admin/list_disks_command.h"
 #include "admin/list_goals_command.h"
 #include "admin/list_metadataservers_command.h"
 #include "admin/list_mounts_command.h"
+#include "admin/list_sessions_command.h"
 #include "admin/list_tasks_command.h"
 #include "admin/magic_recalculate_metadata_checksum_command.h"
 #include "admin/manage_locks_command.h"
@@ -42,17 +45,14 @@
 #include "admin/promote_shadow_command.h"
 #include "admin/ready_chunkservers_count_command.h"
 #include "admin/reload_config_command.h"
-#include "admin/dump_config_command.h"
 #include "admin/save_metadata_command.h"
 #include "admin/stop_master_without_saving_metadata.h"
 #include "admin/stop_task_command.h"
-#include "admin/delete_sessions_command.h"
-#include "admin/list_sessions_command.h"
-#include "common/human_readable_format.h"
-#include "protocol/SFSCommunication.h"
-#include "errors/sfserr.h"
 #include "common/sockets.h"
-#include <common/version.h>
+#include "common/version.h"
+
+using CommandPtr = std::shared_ptr<SaunaFsAdminCommand>;
+using CommandList = std::vector<CommandPtr>;
 
 int printVersion() {
 	std::cout << "SaunaFS Admin Tool\n\n";
@@ -60,7 +60,7 @@ int printVersion() {
 	return 0;
 }
 
-void printCommandHelp(const SaunaFsAdminCommand* command) {
+void printCommandHelp(const CommandPtr &command) {
 	command->usage();
 	if (!command->supportedOptions().empty()) {
 		std::cerr << "  Possible command-line options:\n";
@@ -73,13 +73,13 @@ void printCommandHelp(const SaunaFsAdminCommand* command) {
 }
 
 int printHelp(const std::string& programName,
-	const std::vector<const SaunaFsAdminCommand*>& commands, bool _printVersion = false) {
+	const CommandList& commands, bool _printVersion = false) {
 	if (_printVersion) { printVersion(); }
 	std::cerr << std::endl;
 	std::cerr << "Usage:\n";
 	std::cerr << "    " << programName << " COMMAND [OPTIONS...] [ARGUMENTS...]\n\n";
 	std::cerr << "Available COMMANDs:\n\n";
-	for (const auto command : commands) {
+	for (const auto &command : commands) {
 		if (command->name().substr(0, 6) == "magic-") {
 			// Treat magic-* commands as undocumented
 			continue;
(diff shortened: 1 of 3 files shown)


src/admin/main.cc
```diff
@@ -125,7 +125,7 @@ int main(int argc, const char** argv) {
 #ifdef _WIN32
 		socketinit();
 #endif
-		std::string command_name = argv[1];
+		command_name = argv[1];
```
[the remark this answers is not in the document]
wg1k: @copilot This is exactly what this line changed: removed the shadowed re-declaration of that variable.

src/uraft/floating-ip-manager.cc
```diff
@@ -74,7 +74,9 @@ bool HAFloatingIPManager::restoreFloatingIp() {
 void HAFloatingIPManager::startEventListener() {
 	if (!listenerThread.joinable()) {
 		listenerThread =
-		    std::jthread(&HAFloatingIPManager::eventListenerThread, this);
+		    std::jthread([this](std::stop_token stopToken) {
+			    eventListenerThread(stopToken);
+		    });
```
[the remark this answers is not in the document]
wg1k: @gemini-code-assist `std::bind_front` would only work if `eventListenerThread` already takes a `std::stop_token` as its first parameter. Since the lambda explicitly forwards the injected stop_token, it’s needed here to make the signature match what `std::jthread` expects.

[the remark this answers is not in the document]
wg1k: @gemini-code-assist That makes sense since `eventListenerThread` now takes a `std::stop_token`, `std::bind_front(&HAFloatingIPManager::eventListenerThread, this)` would indeed be equivalent to the lambda. I’ll switch to `bind_front` for conciseness.

src/uraft/floating-ip-manager.cc
```diff
@@ -74,7 +74,7 @@ bool HAFloatingIPManager::restoreFloatingIp() {
 void HAFloatingIPManager::startEventListener() {
 	if (!listenerThread.joinable()) {
 		listenerThread =
-		    std::jthread(&HAFloatingIPManager::eventListenerThread, this);
+		    std::jthread(std::bind_front(&HAFloatingIPManager::eventListenerThread, this));
```
rolysr: I see `std::bind_front` isn’t needed here because eventListenerThread already matches the std::jthread constructor (it just needs the object pointer and optionally the stop_token).
Using `std::jthread(&HAFloatingIPManager::eventListenerThread, this)` is simpler, avoids the extra wrapper, and is the more idiomatic form unless we need to pre-bind additional arguments.

dmga44 (approved):

src/uraft/floating-ip-manager.cc
ralcolea: I would like to know the compiler version reporting this error.
I tested gcc-14 and clang-19 and I could not reproduce it.

ralcolea (commented):
About the change in uraft, I could not reproduce it with the compilers I tested.
The other suggestion is to update the commit message for the change related with uraft, which references the lambda function initially used.

src/uraft/floating-ip-manager.cc
wg1k: Thanks for looking at this, @rolysr. I understand what you're suggesting, but that simpler form `std::jthread(&HAFloatingIPManager::eventListenerThread, this)` was actually the original code that caused the build failure.

The compiler error occurs because `eventListenerThread` takes a `std::stop_token` parameter, but when using the member function pointer approach, the `jthread` cannot automatically inject the `stop_token` it creates. This creates the signature mismatch shown in the error.

Here's the build failure that prompted my change:
```
In file included from /home/ictus/work/leil/saunafs/src/uraft/floating-ip-manager.h:27,
                 from /home/ictus/work/leil/saunafs/src/uraft/floating-ip-manager.cc:19:
/usr/include/c++/14/thread: In instantiation of ‘static std::thread std::jthread::_S_create(std::stop_source&, _Callable&&, _Args&& ...) [with _Callable = void (HAFloatingIPManager::*)(const std::stop_token&); _Args = {HAFloatingIPManager*}]’:
/usr/include/c++/14/thread:163:28:   required from ‘std::jthread::jthread(_Callable&&, _Args&& ...) [with _Callable = void (HAFloatingIPManager::*)(const std::stop_token&); _Args = {HAFloatingIPManager*}; <template-parameter-1-3> = void]’
  163 |       : _M_thread{_S_create(_M_stop_source, std::forward<_Callable>(__f),
      |                   ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  164 |                             std::forward<_Args>(__args)...)}
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ictus/work/leil/saunafs/src/uraft/floating-ip-manager.cc:77:67:   required from here
   77 |                     std::jthread(&HAFloatingIPManager::eventListenerThread, this);
      |                                                                                 ^
/usr/include/c++/14/thread:270:27: error: static assertion failed: std::jthread arguments must be invocable after conversion to rvalues
  270 |             static_assert(is_invocable_v<decay_t<_Callable>,
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  271 |                                          decay_t<_Args>...>,
      |                                          ~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/thread:270:27: note: ‘std::is_invocable_v<void (HAFloatingIPManager::*)(const std::stop_token&), HAFloatingIPManager*>’ evaluates to false
gmake[2]: *** [src/uraft/CMakeFiles/uraft.dir/build.make:79: src/uraft/CMakeFiles/uraft.dir/floating-ip-manager.cc.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:3200: src/uraft/CMakeFiles/uraft.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
```
The lambda wrapper (or `std::bind_front`) resolves this by explicitly showing how the `stop_token` parameter should be passed. My initial fix used a lambda, then I tried the `std::bind_front` alternative that `gemini` suggested, but I'm reverting to lambda per @ralcolea's suggestion since it aligns with my original approach and is clearer for this case.

@ralcolea above is how the compiler cries. Below the versions:

```
> g++ --version && clang --version
g++ (Ubuntu 14.2.0-19ubuntu2) 14.2.0
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Ubuntu clang version 20.1.2 (0ubuntu1)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-20/bin
```

wg1k: Thanks for checking, @ralcolea. That makes sense, the issue appears with newer compiler versions (I'm on Ubuntu 25.04 with GCC 14). This is actually why the fix is valuable: it's catching a future compatibility issue early.

The member function pointer approach with `std::jthread` and a `stop_token` parameter is being correctly flagged as invalid by the newer, more standards-compliant compiler. The lambda wrapper is the portable, future-proof solution that works across all compiler versions.

To help reproduce it, here are the stricter build flags I'm using that might be triggering the more pedantic error checking:

```
> cat CMakeUserPresets.json 
{
    "version": 6,
    "configurePresets": [
        {
            "name": "strict-flags",
            "hidden": true,
            "cacheVariables": {
            },
            "environment": {
                "CXXFLAGS": "-g -O1 -Wall -Wextra -Wpedantic -Wunused -Werror -fno-omit-frame-pointer"
            }
        },
        {
            "name": "stricter-flags",
            "hidden": true,
            "environment": {
                "CXXFLAGS": "-Wconversion -Wsign-conversion -Wshadow -fsanitize=address,undefined ",
                "LDFLAGS": "-fsanitize=address,undefined"
            }
        },
        {
            "name": "my-vcpkg",
            "toolchainFile": "${sourceDir}/vcpkg/scripts/buildsystems/vcpkg.cmake",
            "hidden": true
        },
        {
            "name": "local-test",
            "inherits": ["default", "my-vcpkg", "strict-flags"],
            "cacheVariables": {
                "CMAKE_CXX_EXTENSIONS": "OFF",
                "CMAKE_CXX_STANDARD_REQUIRED": "ON",
                "CMAKE_CXX_STANDARD": "23",
                "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
                "ENABLE_SYSLOG_FOR_TRACES": "ON",
                "ENABLE_TRACES": "OFF"
            }
        }
    ],
    "buildPresets": [
        {
            "name": "local-test",
            "configurePreset": "local-test",
            "inherits": ["default"]
        }
    ],
    "workflowPresets": [
        {
            "name": "local-test",
            "steps": [
                {
                    "type": "configure",
                    "name": "local-test"
                },
                {
                    "type": "build",
                    "name": "local-test"
                }
            ]
        }
    ]
}
```

Since stable environments will eventually catch up to these toolchains, let's fix this now with the lambda approach. This will prevent the exact same build failure for everyone else in the future. It's the standards-compliant way to handle this pattern.

rolysr: Good, now is much clearer to me on this context you provided.

rolysr (approved):

PR #583, opened 2025-09-22
refactor: Apply few improvements in metadata load

dmga44:
This commit encompasses some minor improvements in the metadata load.
The main goal of this changes is to boost the loading time. The changes
can be split into the following:
- do not log trace the recalculation of chunks checksum while loading.
- reuse the already found parent FSNode* structure while parsing edges.
- simplify the accessing to some data (hashes, id).

### DIFF
--- a/src/master/metadata_backend_file.cc
+++ b/src/master/metadata_backend_file.cc
@@ -382,8 +382,10 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
 	static const int8_t kSuccess = 0;
 	static const int8_t kLastEdge = 1;
 	static inode_t currentParentId;
+	static FSNodeDirectory* currentParentNode;
 	if (init) {
 		currentParentId = 0;
+		currentParentNode = nullptr;
 		return kSuccess;
 	}
 	const auto* pSrc = metadataFile->seek(sectionOffset);
@@ -434,7 +436,13 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
 			return kError;
 		}
 	} else {
-		FSNodeDirectory *parent = fsnodes_id_to_node<FSNodeDirectory>(parentId);
+		FSNodeDirectory *parent;
+		if (currentParentId != parentId){
+			parent = fsnodes_id_to_node<FSNodeDirectory>(parentId);
+			currentParentNode = parent;
+		} else {
+			parent = currentParentNode;
+		}
 		if (!parent) {
 			safs_pretty_syslog(LOG_ERR,
 			                   "loading edge: %" PRIiNode ",%s->%" PRIiNode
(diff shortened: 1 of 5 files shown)


antuan96314 (approved):

uristdwarf (approved):

src/master/metadata_backend_file.cc
```diff
@@ -506,20 +513,18 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
 			currentParentId = parentId;
 		}
 
-		hstorage::Handle *handlePtr = new hstorage::Handle(name);
-		auto it = parent->entries.insert({handlePtr, child}).first;
-		parent->entries_hash ^= (*it).first->hash();
+		auto *handlePtr = new hstorage::Handle(name);
+		parent->entries.insert({handlePtr, child});
+		parent->entries_hash ^= handlePtr->hash();
```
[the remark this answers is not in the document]
dmga44: Applied, in some way.

src/master/metadata_backend_file.cc
```diff
@@ -506,20 +513,18 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
...
 		if (parent->case_insensitive) {
 			HString lowerCaseName = HString::hstringToLowerCase(HString(name));
-			auto lowercaseHandlePtr = new hstorage::Handle(lowerCaseName);
-			auto it =
-			    parent->lowerCaseEntries.insert({lowercaseHandlePtr, child})
-			        .first;
-			parent->lowerCaseEntriesHash ^= (*it).first->hash();
+			auto *lowercaseHandlePtr = new hstorage::Handle(lowerCaseName);
+			parent->lowerCaseEntries.insert({lowercaseHandlePtr, child});
+			parent->lowerCaseEntriesHash ^= lowercaseHandlePtr->hash();
```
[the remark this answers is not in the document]
dmga44: Applied, similarly as above.

src/master/metadata_backend_file.cc
```diff
@@ -382,6 +382,7 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
 	static const int8_t kSuccess = 0;
 	static const int8_t kLastEdge = 1;
 	static inode_t currentParentId;
+	static FSNodeDirectory* currentParentNode;
 	if (init) {
 		currentParentId = 0;
```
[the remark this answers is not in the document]
dmga44: Applied.

src/master/metadata_backend_file.cc
```diff
@@ -506,20 +514,22 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
...
 		}
 
-		hstorage::Handle *handlePtr = new hstorage::Handle(name);
-		auto it = parent->entries.insert({handlePtr, child}).first;
-		parent->entries_hash ^= (*it).first->hash();
+		auto *handlePtr = new hstorage::Handle(name);
+		if (parent->entries.insert({handlePtr, child}).second) {
+			// On successful insert update the hash
+			parent->entries_hash ^= handlePtr->hash();
+		}
```
rolysr, line 525: I think this change is a good opportunity to check if this is completely memory safe on this context. As there is not delete on this `handlePtr` creating using `new`.

Two approaches that can be considered could be:

1.  `std::unique_ptr` to manage the memory automatically:

```
auto handlePtr = std::make_unique<hstorage::Handle>(name);
auto result = parent->entries.insert({handlePtr.get(), child});
if (result.second) {
    parent->entries_hash ^= handlePtr->hash();
    handlePtr.release(); // ownership transferred to entries
}
// no need to delete manually
```

2. Avoid dynamic allocation entirely if possible:
```
hstorage::Handle handle(name);
auto result = parent->entries.insert({&handle, child});
if (result.second) {
    parent->entries_hash ^= handle.hash();
}
```
Which only works if entries can store references or copies safely, then no heap allocation needed.

On the other hand, you can also consider adding here a safer patter, that deletes the handle if the insert failed, for instance:
```
auto* handlePtr = new hstorage::Handle(name);
auto result = parent->entries.insert({handlePtr, child});
if (result.second) {
    // successfully inserted → entries now owns handlePtr
    parent->entries_hash ^= handlePtr->hash();
} else {
    // insert failed → nobody owns handlePtr → delete it
    delete handlePtr;
}
```

src/master/metadata_backend_file.cc
```diff
@@ -506,20 +514,22 @@ static int8_t fs_parseEdge(const std::shared_ptr<MemoryMappedFile> &metadataFile
...
 
 		if (parent->case_insensitive) {
 			HString lowerCaseName = HString::hstringToLowerCase(HString(name));
-			auto lowercaseHandlePtr = new hstorage::Handle(lowerCaseName);
-			auto it =
-			    parent->lowerCaseEntries.insert({lowercaseHandlePtr, child})
-			        .first;
-			parent->lowerCaseEntriesHash ^= (*it).first->hash();
+			auto *lowercaseHandlePtr = new hstorage::Handle(lowerCaseName);
```
rolysr, line 529: Also, consider here my previous comment if applicable.

src/master/metadata_backend_file.cc
dmga44: These are my thoughts on your suggestions:
1. using unique_ptr does not make sense because we are not going to move ownership of those structures, we are going to store them somewhere in heap in such a way that future requests are going to be able to use it in a very lightweighted fashion. Plus, I don't like the idea of instantiating millions of unique_ptr structures while loading metadata.
2. as you mentioned, it could only work if the insertion copied the Handle structure given by its point pointer, which it does not do (it only saves the pointer). The destruction of the Handle at the end of the fs_parseEdge scope should make the inserted pointer to address some unallocated memory.

I'm applying the final suggestion, with some changes (to return kError in the failed insert case).

src/master/metadata_backend_file.cc
dmga44: Similar reply as above.

src/master/metadata_backend_file.cc
rolysr: Very good, that will be enough for this case.

rolysr (approved):

lgsilva3087 (approved):
LGTM

ralcolea (approved):
Nice job @dmga44!! 👍

PR #586, opened 2025-09-22
fix(master): Avoid possible reinit of gMetadata

lgsilva3087:
During initialization, some backends could create the gMetadata instance earlier. This commit adds a partial protection but still allows the old behavior (for compatibility) if called from other places, like from masterconn in Shadow mode (as before).

### DIFF
--- a/src/master/filesystem.cc
+++ b/src/master/filesystem.cc
@@ -207,8 +207,13 @@ void fs_term(const char *fname, bool noLock) {
 }
 #endif
 
-void fs_strinit(void) {
-	gMetadata = new FilesystemMetadata;
+void fs_strinit(bool isFromInit) {
+	if (isFromInit) {
+		if (gMetadata == nullptr) { gMetadata = new FilesystemMetadata; }
+	} else {
+		// Could be called from masterconn in Shadow mode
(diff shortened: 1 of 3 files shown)


ralcolea (approved):
LGTM 👍

src/master/filesystem.cc
```diff
@@ -207,8 +207,13 @@ void fs_term(const char *fname, bool noLock) {
 }
 #endif
 
-void fs_strinit(void) {
-	gMetadata = new FilesystemMetadata;
+void fs_strinit(bool isFromInit) {
+	if (isFromInit) {
+		if (gMetadata == nullptr) { gMetadata = new FilesystemMetadata; }
```
antuan96314, line 212: Why not defining `FilesystemMetadata` as Singleton???

antuan96314 (approved):

src/master/filesystem.cc
lgsilva3087: We could consider it in following changes.
In the current approach it could be reset for example in Shadow mode in the function MasterConn::downloadNext(), which implies a kind of reset of the instance (to be implemented).

rolysr (approved):
LGTM.

PR #590, opened 2025-09-24
feat(master): Improve observability of metadata changes

lgsilva3087:
Improves the observability of metadata changes by:
- Adding a property/signal for changes in the next chunk id (for persistency).
- Emitting a signal on goal changes (to persist them in custom backends).
- Using const references for changelog signals to avoid copying.

### DIFF
--- a/src/master/changelog.cc
+++ b/src/master/changelog.cc
@@ -47,8 +47,8 @@
 #include "master/restore.h"
 #endif  // #if !defined(METARESTORE) && !defined(METALOGGER)
 
-Signal<ChangelogEvent> &getChangelogSignal() {
-	static Signal<ChangelogEvent> gChangelogSignal;
+Signal<const ChangelogEvent &> &getChangelogSignal() {
+	static Signal<const ChangelogEvent &> gChangelogSignal;
 	return gChangelogSignal;
 }
 
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -1210,6 +1210,7 @@ void fsnodes_changefilegoal(FSNodeFile *obj, uint8_t goal) {
 		}
 	}
 	fsnodes_update_checksum(obj);
+	gMetadata->nodeChangedSignal.emit(obj);
 }
 
 void fsnodes_setlength(FSNodeFile *obj, uint64_t length, bool eraseFurtherChunks) {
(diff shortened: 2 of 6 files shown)


ralcolea (approved):
Nice job @lgsilva3087!! 👍

src/master/changelog.cc
```diff
@@ -47,8 +47,8 @@
 #include "master/restore.h"
 #endif  // #if !defined(METARESTORE) && !defined(METALOGGER)
 
-Signal<ChangelogEvent> &getChangelogSignal() {
-	static Signal<ChangelogEvent> gChangelogSignal;
+Signal<const ChangelogEvent &> &getChangelogSignal() {
```
wg1k, line 50: This change to const& is fine as long as signal dispatch stays strictly synchronous. My concern is that with future plans for multi-threading/distributed sharding, any deferred/async delivery would make the reference dangle. Maybe we should keep it by value for safety, or at least document the synchronous guarantee clearly (alternatives: shared_ptr<const T> etc.).

src/master/filesystem_node.cc
```diff
@@ -1210,6 +1210,7 @@ void fsnodes_changefilegoal(FSNodeFile *obj, uint8_t goal) {
 		}
 	}
 	fsnodes_update_checksum(obj);
+	gMetadata->nodeChangedSignal.emit(obj);
```
wg1k, line 1213: One more design-level note: right now the signal is emitting the whole object. That works, but because the object is large, it already pushed us toward passing by reference. If signals ever evolve toward async/queued delivery, this creates a lifetime risk.

Would it make sense to instead emit just a stable identifier (e.g. enum / catalog key / object ID), and let slots fetch the full object if they need it? This keeps the signal payload lightweight, avoids lifetime pitfalls, and decouples slots from the full object layout.

lgsilva3087: In this specific case, the signal sends a pointer. As the slot execution is synchronous so far, the object will survive the slots, byt anyway I will add checks (Most likely in future PRs).

This will take a different meaning if we decide to add more threads to the MDS, which is the only one right now consuming them.

src/master/changelog.cc
lgsilva3087: I agree, so far I am proposing a share-nothing architecture with light weigth MDS processes instead of fully multi-threaded processes. I we decide later other approach, then we need to improve the signals-slots mechanism accordingly.

wg1k (approved):

PR #591, opened 2025-09-24
test: Fix flakiness in test_cgi_validate_html

ralcolea:
Previously, `test_cgi_validate_html` could fail if a periodic metadata dump triggered a changelog rotation, leaving no `changelog.sfs` file available. This was a rare race condition, but it occasionally caused test failures.

This commit fixes the issue by disabling metadata dumps during the test, which prevents unexpected changelog rotations and ensures consistent results.


rolysr (approved):

lgsilva3087 (approved):
LGTM

PR #598, opened 2025-09-25
feat(master): Add initial id generator interface

lgsilva3087:
This commit starts moving the id generation towards a more extensible interface. IIdGenerator provides the interface, compatible with the existing code, while IdGeneratorWithDetainer maintains the current behavior based on an inode pool detainer.

It is expected some custom implementations to arrive soon to provide a distributed id generation for the MDS.

### DIFF
--- a/src/master/filesystem.cc
+++ b/src/master/filesystem.cc
@@ -48,6 +48,10 @@
 #include "master/restore.h"
 #include "slogger/slogger.h"
 
+#ifdef METARESTORE
+#include "master/filesystem_freenode.h"
+#endif
+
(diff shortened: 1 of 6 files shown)


rolysr (approved):
LGTM.

dmga44 (approved):
LGTM.

ralcolea (approved):
Good job @lgsilva3087!! 👍

uristdwarf:
Minor suggestion: I don't think this is really a feat, unless it's something users will use? Maybe refactor is better.

PR #601, opened 2025-09-29
fix: Fix `removeReservedEntry` function usage

rolysr:
Since commit 4386f2dd, the usage of the `removeReservedEntry` function in the master code was unsafe when called from `fs_do_emptyreserved`. The execution path could lead to `removeReservedEntry` being called with a null `FSNodeFile` instance, resulting in a dereference error when accessing its `id` field. This commit fixes the issue by improving `removeReservedEntry` to accept only the required parameters, avoiding null reference passing on id value. The problem was detected when compiling the project using the `release-vcpkg` preset.

### DIFF
--- a/src/master/filesystem_trash_reserved_files.h
+++ b/src/master/filesystem_trash_reserved_files.h
@@ -147,8 +147,14 @@ inline void addTrashEntry(TrashPathContainer &trash, HandleIndexContainer &trash
 inline void removeTrashEntry(TrashPathContainer &trash, HandleIndexContainer &trashHandlesIndex,
                              TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
 	TrashPathKey key(node);
-	uint64_t handleHash = trash.at(key).data();
-	uint64_t pathId = trashReservedToId.getOrAdd(handleHash);
+	auto handleHashIt = trash.find(key);
+	if (handleHashIt == trash.end()) { return; }
+
+	uint64_t handleHash = (*handleHashIt).second.data();
+	auto pathIdIt = trashReservedToId.nameToId.find(handleHash);
+	if (pathIdIt == trashReservedToId.nameToId.end()) { return; }
+
+	uint64_t pathId = pathIdIt->second;
 	trashHandlesIndex.erase(HandleIndexKey(pathId));
 	trashReservedToId.erase(handleHash);
 	trash.erase(key);
@@ -167,8 +173,14 @@ inline void addReservedEntry(ReservedPathContainer &reserved,
 inline void removeReservedEntry(ReservedPathContainer &reserved,
                                 HandleIndexContainer &reservedHandlesIndex,
                                 TrashReservedToIdContainer &trashReservedToId, inode_t inode) {
-	uint64_t handleHash = reserved.at(inode).data();
-	uint64_t pathId = trashReservedToId.getOrAdd(handleHash);
+	auto handleHashIt = reserved.find(inode);
+	if (handleHashIt == reserved.end()) { return; }
+
+	uint64_t handleHash = (*handleHashIt).second.data();
+	auto pathIdIt = trashReservedToId.nameToId.find(handleHash);
+	if (pathIdIt == trashReservedToId.nameToId.end()) { return; }
+
+	uint64_t pathId = pathIdIt->second;
 	reservedHandlesIndex.erase(HandleIndexKey(pathId));
 	trashReservedToId.erase(handleHash);
 	reserved.erase(inode);
@@ -180,8 +192,14 @@ inline void moveTrashToReservedEntry(TrashPathContainer &trash,
                                      HandleIndexContainer &reservedHandlesIndex,
                                      TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
 	TrashPathKey key(node);
-	uint64_t handleHash = trash.at(key).data();
-	uint64_t idFromName = trashReservedToId.getOrAdd(handleHash);
+	auto handleHashIt = trash.find(key);
+	if (handleHashIt == trash.end()) { return; }
+
+	uint64_t handleHash = (*handleHashIt).second.data();
+	auto pathIdIt = trashReservedToId.nameToId.find(handleHash);
+	if (pathIdIt == trashReservedToId.nameToId.end()) { return; }
+
+	uint64_t idFromName = pathIdIt->second;
 
 	trashHandlesIndex.erase(HandleIndexKey(idFromName));
 	hstorage::Handle nameHandle = std::move(trash.at(key));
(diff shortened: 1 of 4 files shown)


uristdwarf (approved):

src/master/filesystem_trash_reserved_files.h
```diff
@@ -166,12 +166,12 @@ inline void addReservedEntry(ReservedPathContainer &reserved,
 
 inline void removeReservedEntry(ReservedPathContainer &reserved,
                                 HandleIndexContainer &reservedHandlesIndex,
-                                TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
-	uint64_t handleHash = reserved.at(node->id).data();
+                                TrashReservedToIdContainer &trashReservedToId, inode_t inode) {
```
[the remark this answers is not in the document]
rolysr: The handle hash is dependent on what is on the reference for that entry on the reserved container. If that handle hash is not found then by definition, that entry does not exists on that container, thus the call to `removeReservedEntry` function would have no sense.  The intention of the auxiliary `removeReservedEntry` function is to substitute the code logic that was before on the portions of the code deleting entries from trash. On the other hand, as the first operation that is done is to retrieve the `uint64_t handleHash = reserved.at(inode).data();` and that did not failed, then by definition the `uint64_t pathId = trashReservedToId.getOrAdd(handleHash);` will work. Also, on that scenario it would have first thrown an Exception from the `handleHash` variable retrieval, thus stopping the execution.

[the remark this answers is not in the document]
rolysr: Now I added new checks on the removal related logic functions.

src/master/filesystem_trash_reserved_files.h
```diff
@@ -147,8 +147,13 @@ inline void addTrashEntry(TrashPathContainer &trash, HandleIndexContainer &trash
 inline void removeTrashEntry(TrashPathContainer &trash, HandleIndexContainer &trashHandlesIndex,
                              TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
 	TrashPathKey key(node);
+	if (trash.find(key) == trash.end()) { return; }
+
 	uint64_t handleHash = trash.at(key).data();
-	uint64_t pathId = trashReservedToId.getOrAdd(handleHash);
+	auto pathIdIt = trashReservedToId.nameToId.find(handleHash);
+	if (pathIdIt == trashReservedToId.nameToId.end()) { return; }
```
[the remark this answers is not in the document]
rolysr: I prefer the `find` version as it gives me also access to the value I want to retrieve from the given key.

src/master/filesystem_trash_reserved_files.h
```diff
@@ -147,8 +147,13 @@ inline void addTrashEntry(TrashPathContainer &trash, HandleIndexContainer &trash
 inline void removeTrashEntry(TrashPathContainer &trash, HandleIndexContainer &trashHandlesIndex,
                              TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
 	TrashPathKey key(node);
+	if (trash.find(key) == trash.end()) { return; }
```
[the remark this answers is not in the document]
rolysr: Same answer as for your previous comment.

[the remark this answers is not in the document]
dmga44: @rolysr I think your answer is not valid here, since you are not reusing the iterator the find provided (you may use it while getting the `handleHash` like you did with `pathIdIt`).

src/master/filesystem_trash_reserved_files.h
```diff
@@ -166,12 +171,17 @@ inline void addReservedEntry(ReservedPathContainer &reserved,
 
 inline void removeReservedEntry(ReservedPathContainer &reserved,
                                 HandleIndexContainer &reservedHandlesIndex,
-                                TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
-	uint64_t handleHash = reserved.at(node->id).data();
-	uint64_t pathId = trashReservedToId.getOrAdd(handleHash);
+                                TrashReservedToIdContainer &trashReservedToId, inode_t inode) {
+	if (reserved.find(inode) == reserved.end()) { return; }
+
+	uint64_t handleHash = reserved.at(inode).data();
```
dmga44: Consider applying a similar pattern to the one below with `pathIdIt `.

src/master/filesystem_trash_reserved_files.h
```diff
@@ -180,8 +190,13 @@ inline void moveTrashToReservedEntry(TrashPathContainer &trash,
                                      HandleIndexContainer &reservedHandlesIndex,
                                      TrashReservedToIdContainer &trashReservedToId, FSNode *node) {
 	TrashPathKey key(node);
+	if (trash.find(key) == trash.end()) { return; }
+
 	uint64_t handleHash = trash.at(key).data();
```
dmga44: Consider applying a similar pattern to the one below with `pathIdIt`.

src/master/filesystem_trash_reserved_files.h
```diff
@@ -194,17 +209,28 @@ inline void moveTrashToReservedEntry(TrashPathContainer &trash,
 inline void updateTrashNameEntry(TrashPathContainer &trash, HandleIndexContainer &trashHandlesIndex,
                                  TrashReservedToIdContainer &trashReservedToId, FSNode *node,
                                  const std::string &newPath) {
-	uint64_t oldHash = trash[TrashPathKey(node)].data();
-	trashHandlesIndex.erase(HandleIndexKey(trashReservedToId.getOrAdd(oldHash)));
+	TrashPathKey key(node);
+	if (trash.find(key) == trash.end()) { return; }
+
+	uint64_t oldHash = trash[key].data();
+
```
dmga44: Consider applying a similar pattern to the one below with `oldPathIdIt`.

src/master/filesystem_trash_reserved_files.h
```diff
@@ -194,17 +209,28 @@ inline void moveTrashToReservedEntry(TrashPathContainer &trash,
...
+	trash[key] = HString(newPath);
+	uint64_t newHash = trash[key].data();
 	trashHandlesIndex.insert({HandleIndexKey(trashReservedToId.getOrAdd(newHash)), node->id});
 }
 
 inline void updateTrashFromOldEntry(TrashPathContainer &trash, FSNode *node,
                                     const TrashPathKey &oldKey) {
+	if (trash.find(oldKey) == trash.end()) { return; }
+
 	hstorage::Handle path = std::move(trash.at(oldKey));
```
dmga44: Same here.

dmga44 (approved):
Please check my comments, there is an extra one in the gemini's comments.

src/master/filesystem_trash_reserved_files.h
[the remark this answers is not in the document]
rolysr: Done.

src/master/filesystem_trash_reserved_files.h
rolysr: Done.

src/master/filesystem_trash_reserved_files.h
rolysr: Done.

src/master/filesystem_trash_reserved_files.h
rolysr: Done.

src/master/filesystem_trash_reserved_files.h
rolysr: Done.

PR #603, opened 2025-09-30
chore(build): Bump some vcpkg ports versions

lgsilva3087:
Upgrades fmt, spdlog and gtest versions for better compatibility with current compiler versions like clang21.


ralcolea (approved):
LGTM 👍

uristdwarf (approved):

PR #605, opened 2025-09-30
fix: Fix issue with assert version check

rolysr:
After commit 69eecb26, an assert check in the
`ChunkWorker::tryReplication` function caused an unused variable warning when building with the `release-vcpkg` preset. This happened because the variable was only used in an assertion, which is omitted in release
builds, leading to a compilation error in release-ready SaunaFS binaries. This commit resolves the issue.


lgsilva3087 (approved):
LGTM

uristdwarf (approved):

ralcolea (approved):
LGTM

PR #612, opened 2025-10-03
Improve path by inode consistency

rolysr:
This pull request includes two changes:

- fix: Avoid non-positive PIDs in path by inode (6213a29f)
Ensures that only positive PIDs are used when managing the mapping for pending path-by-inode requests. This prevents invalid entries and potential inconsistencies that could occur if special files were opened by contexts with non-positive PIDs (such as PID 0).

- test: Add soft and hard links for path by inode (558f1a47)
Expands the path-by-inode test suite to include checks for both soft (symbolic) and hard links. This improves test coverage and verifies that the feature works correctly for these additional cases.

### DIFF
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -951,8 +951,10 @@ EntryParam lookup(Context &ctx, inode_t parent, const char *name) {
 			throw RequestException(status);
 		}
 		status = SAUNAFS_STATUS_OK;
-		PidPathEntry entry{ctx.pid, fullPath};
-		gInodePathInfo.contextPidToPath.insert(entry);
+		if (ctx.pid > 0) {
+			PidPathEntry entry{ctx.pid, fullPath};
+			gInodePathInfo.contextPidToPath.insert(entry);
+		}
 		attr[0] = TYPE_FILE;
 		inode = parent;
 		icacheflag = 0;
--- a/src/mount/sauna_client.cc
+++ b/src/mount/sauna_client.cc
@@ -953,7 +953,7 @@ EntryParam lookup(Context &ctx, inode_t parent, const char *name) {
 		status = SAUNAFS_STATUS_OK;
 		if (ctx.pid > 0) {
 			PidPathEntry entry{ctx.pid, fullPath};
-			gInodePathInfo.contextPidToPath.insert(entry);
+			gInodePathInfo.contextPidToPath[entry]++;
 		}
 		attr[0] = TYPE_FILE;
 		inode = parent;
--- a/src/mount/special_release.cc
+++ b/src/mount/special_release.cc
@@ -98,7 +98,12 @@ namespace InodePathByInode {
 static void release(FileInfo *fi) {
 	std::unique_lock<std::mutex> lock(gInodePathInfo.mtx);
 	auto entry = reinterpret_cast<PidPathEntry *>(fi->fh);
-	if (entry) { gInodePathInfo.contextPidToPath.erase(*entry); }
+	if (entry) {
+		auto it = gInodePathInfo.contextPidToPath.find(*entry);
+		if (it != gInodePathInfo.contextPidToPath.end() && --(it->second) <= 0) {
+			gInodePathInfo.contextPidToPath.erase(it);
+		}
+	}
 	fi->fh = 0;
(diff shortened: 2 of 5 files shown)


tests/test_suites/ShortSystemTests/test_path_by_inode.sh
```diff
@@ -16,3 +16,24 @@ chmod 200 folder
...
 assert_failure cat .saunafs_path_by_inode/6
+
+# test path by inode for softlinks and hardlinks
+touch file5
+echo "data_file5" > file5
+
+ln -s file5 softlink_to_file
+ln file5 hardlink_to_file
+
+assert_equals "file5" "$(cat .saunafs_path_by_inode/8)"
```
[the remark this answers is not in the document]
rolysr: Due to the limitation of the feature, for now, only one path should be returning for each inode. If required next, it could be possible to return all paths for each inode, but for now only the path of the original file is required. And in the case of hard links, their paths will be returned in case original file is deleted. As general approach, path by inode feature will return oldest path to file pointing to that inode in our filesystem, instead of printing all files with same inode. Which is something optional to implement.

src/mount/sauna_client.cc
```diff
@@ -951,8 +951,10 @@ EntryParam lookup(Context &ctx, inode_t parent, const char *name) {
 			throw RequestException(status);
 		}
 		status = SAUNAFS_STATUS_OK;
-		PidPathEntry entry{ctx.pid, fullPath};
-		gInodePathInfo.contextPidToPath.insert(entry);
+		if (ctx.pid > 0) {
+			PidPathEntry entry{ctx.pid, fullPath};
+			gInodePathInfo.contextPidToPath.insert(entry);
```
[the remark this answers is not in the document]
rolysr: The current approach of path by inode feature is to only one request is expected to be handled by PID, and in case there are two requests of the same PID, it is ensure that there will be no release of existing path buffer in case another request for the same PID is pending to read it. In the context of out filesystem, for each path by inode request, a new PID is used in order to serve that specific request. The only way for now to force a same PID to request two different inodes' paths is to create a custom app that handle that specific scenario.

src/mount/sauna_client.cc
```diff
@@ -951,8 +951,15 @@ EntryParam lookup(Context &ctx, inode_t parent, const char *name) {
...
-		PidPathEntry entry{ctx.pid, fullPath};
-		gInodePathInfo.contextPidToPath.insert(entry);
+		if (ctx.pid > 0) {
+			PidPathEntry entry{ctx.pid, fullPath};
+			auto it = gInodePathInfo.contextPidToPath.find(entry);
+			if (it == gInodePathInfo.contextPidToPath.end()) {
+				gInodePathInfo.contextPidToPath[entry] = 1;
+			} else {
+				it->second++;
+			}
```
dmga44: Consider substituting these lines by:
```cpp
gInodePathInfo.contextPidToPath[entry]++;
```

src/mount/special_release.cc
```diff
@@ -98,7 +98,16 @@ namespace InodePathByInode {
...
-	if (entry) { gInodePathInfo.contextPidToPath.erase(*entry); }
+	if (entry) {
+		auto it = gInodePathInfo.contextPidToPath.find(*entry);
+		if (it != gInodePathInfo.contextPidToPath.end()) {
+			if (it->second > 1) {
+				--(it->second);
+			} else {
+				gInodePathInfo.contextPidToPath.erase(it);
+			}
+		}
```
dmga44, line 105: Consider substituing these lines by:
```cpp
if (--gInodePathInfo.contextPidToPath[*entry] <= 0) {
	gInodePathInfo.contextPidToPath.erase(*entry);
}
```
or by:
```cpp
auto it = gInodePathInfo.contextPidToPath.find(*entry);
if (it != gInodePathInfo.contextPidToPath.end() && --(it->second) <= 0) {
	gInodePathInfo.contextPidToPath.erase(it);
}
```

dmga44 (approved):
LGTM, please see my small suggestions.

src/mount/sauna_client.cc
rolysr: Done.

src/mount/special_release.cc
rolysr: Done.

lgsilva3087 (approved):

PR #625, opened 2025-10-13
Fix saunafs command display of non-ASCII chars on Windows.

rolysr:
Previously, special characters in paths (like accented letters or other non-ASCII characters) were not handled correctly on Windows when using
the saunafs tool. This commit ensures that paths are properly converted to UTF-8 before processing, allowing for correct handling of such characters.

### DIFF
--- a/src/common/args_stat_encoding.h
+++ b/src/common/args_stat_encoding.h
@@ -1,0 +1,98 @@
+/*
+   Copyright 2023      Leil Storage OÜ
+
+   This file is part of SaunaFS.
+
+   SaunaFS is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "common/platform.h"
+
+#include <sys/stat.h>
+
+#ifdef _WIN32
+#include <shellapi.h>
+#include <windows.h>
+#include <string>
+#include <vector>
+
+using saunafs_stat_t = struct _stat64;
+
+// Represents UTF-8 command-line arguments for Windows.
+// Keeps argvPtrs alive for the lifetime of the object.
+class Utf8CmdArguments {
+public:
+	Utf8CmdArguments() {
+		argvUtf8_ = get_utf8_argv();
+		argvPtrs_.reserve(argvUtf8_.size() + 1);
+		for (auto &s : argvUtf8_) { argvPtrs_.push_back(s.data()); }
+		argvPtrs_.push_back(nullptr);
+	}
+
+	char **getArgv() const { return const_cast<char **>(argvPtrs_.data()); }
+	int getArgc() const { return static_cast<int>(argvUtf8_.size()); }
+
+private:
+	static std::vector<std::string> get_utf8_argv() {
+		int argcW = 0;
+		LPWSTR *argvW = CommandLineToArgvW(GetCommandLineW(), &argcW);
+		std::vector<std::string> argvUtf8;
+		if (!argvW || argcW <= 0) { return argvUtf8; }
+		argvUtf8.reserve(argcW);
+
+		for (int i = 0; i < argcW; ++i) {
+			int size = WideCharToMultiByte(CP_UTF8, 0, argvW[i], -1, nullptr, 0, nullptr, nullptr);
+			if (size <= 1) {
+				argvUtf8.emplace_back();
+				continue;
+			}
+			std::string utf8(size, '\0');
+			int written =
+			    WideCharToMultiByte(CP_UTF8, 0, argvW[i], -1, &utf8[0], size, nullptr, nullptr);
+			if (written != size || written == 0) {
+				argvUtf8.emplace_back();
+				continue;
+			}
+			utf8.resize(size - 1);  // Remove the null terminator
+			argvUtf8.push_back(std::move(utf8));
+		}
+		LocalFree(argvW);
+		return argvUtf8;
+	}
+
+	std::vector<std::string> argvUtf8_;
+	std::vector<char *> argvPtrs_;
+};
+
+inline int utf8_stat(const char *path_utf8, saunafs_stat_t *st) {
+	int wlen = MultiByteToWideChar(CP_UTF8, 0, path_utf8, -1, nullptr, 0);
+	if (wlen <= 0) { return -1; }
+
+	std::wstring wpath(wlen, L'\0');  // include null terminator
+	MultiByteToWideChar(CP_UTF8, 0, path_utf8, -1, wpath.data(), wlen);
+
+	return _wstat64(wpath.c_str(), st);
+}
+
+// Unified portable wrapper
+inline int stat_portable(const char *path, saunafs_stat_t *st) {
+	return utf8_stat(path, reinterpret_cast<struct _stat64 *>(st));
+}
+#else
+using saunafs_stat_t = struct stat;
+
+// On POSIX just call stat
+inline int stat_portable(const char *path, saunafs_stat_t *st) { return stat(path, st); }
+#endif
--- a/src/tools/main.cc
+++ b/src/tools/main.cc
@@ -33,6 +33,10 @@
 #include "tools/tools_commands.h"
 #include "tools/tools_common_functions.h"
 
+// This import should be placed after other includes to avoid Windows dependency issues
+// with winsock2.h and windows.h included in tools/tools_common_functions.h
+#include "common/args_stat_encoding.h"
+
 static char path_buf[PATH_MAX];
 
 void split(std::vector<char*> &argv_new, std::vector<char> &line) {
@@ -62,6 +66,16 @@ int main(int argc, char **argv) {
 	set_humode();
 
 #ifdef _WIN32
+	// Enable UTF-8 encoding for console I/O on Windows. This is necessary to properly
+	// handle Unicode characters in file paths and other strings on Windows systems,
+	// as the default console code page may not support UTF-8.
+	SetConsoleCP(CP_UTF8);
+	SetConsoleOutputCP(CP_UTF8);
+
+	Utf8CmdArguments args;
+	argc = args.getArgc();
+	argv = args.getArgv();
+
 	socketinit();
 #endif
 
--- a/src/tools/master_functions.cc
+++ b/src/tools/master_functions.cc
@@ -44,6 +44,10 @@
 #include "protocol/matocl.h"
 #include "tools/tools_common_functions.h"
 
+// This import should be placed after other includes to avoid Windows dependency issues
+// with winsock2.h and windows.h
+#include "common/args_stat_encoding.h"
+
 struct master_info_t {
 	uint32_t ip;
 	uint16_t port;
@@ -126,12 +130,10 @@ static bool contains_master_info_name_end(const char *name) {
 static int read_master_info(const char *name, master_info_t *info) {
 	static constexpr int kMasterInfoSize = 14;
 	uint8_t buffer[kMasterInfoSize];
-	struct stat stb;
+	saunafs_stat_t stb;
 	int sd;
 
-	if (stat(name, &stb) < 0) {
-		return -1;
-	}
+	if (stat_portable(name, &stb) < 0) { return -1; }
 
 	if ((stb.st_ino != SPECIAL_INODE_MASTERINFO &&
 	     !contains_master_info_name_end(name)) ||
@@ -266,7 +268,7 @@ void get_next_path_iteration(std::string &path) {
 
 int open_master_conn(const char *name, inode_t *inode, mode_t *mode, [[maybe_unused]] bool needrwfs) {
 	char rpath[PATH_MAX + 1];
-	struct stat stb;
+	saunafs_stat_t stb;
 	[[maybe_unused]] struct statvfs stvfsb;
 	master_info_t master_info;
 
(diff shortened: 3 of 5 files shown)


src/tools/dir_info.cc
```diff
@@ -72,16 +72,17 @@ static int dir_info(const char *fname) {
 	put32bit(&wptr, msgid);
 	putINode(&wptr, inode);
 
+	std::string debugFName = utf8_from_cp(fname);
 	// send the request
 	if (tcpwrite(fd, reqbuff, kReqBuffSize) != kReqBuffSize) {
-		printf("%s: master query: send error\n", fname);
+		printf("%s: master query: send error\n", debugFName.c_str());
```
lgsilva3087: Consider std::println if fits.

src/tools/dir_info.cc
```diff
@@ -72,16 +72,17 @@ static int dir_info(const char *fname) {
...
 		close_master_conn(1);
 		return -1;
 	}
 
 	// read the first part of the answer (cmd and length)
 	if (tcpread(fd, reqbuff, sizeof(cmd) + sizeof(leng)) != sizeof(cmd) + sizeof(leng)) {
-		printf("%s: master query: receive error\n", fname);
+		printf("%s: master query: receive error\n", debugFName.c_str());
```
lgsilva3087: The same with the others.

lgsilva3087 (commented):
Please check if my suggestion is applicable.

src/tools/dir_info.cc
uristdwarf: I think std::println doesn't work on our Ubuntu 22.04 builds, but please check

src/tools/tools_common_functions.h
```diff
@@ -122,3 +123,19 @@ inline uint32_t getGId() {
...
+// Convert a string from the current code page to UTF-8
+inline std::string utf8_from_cp(const char *cp_str) {
+	int wlen = MultiByteToWideChar(CP_ACP, 0, cp_str, -1, nullptr, 0);
+	std::wstring wstr(wlen, 0);
+	MultiByteToWideChar(CP_ACP, 0, cp_str, -1, wstr.data(), wlen);
+
+	int len = WideCharToMultiByte(CP_UTF8, 0, wstr.c_str(), -1, nullptr, 0, nullptr, nullptr);
+	std::string utf8(len - 1, '\0');
+	WideCharToMultiByte(CP_UTF8, 0, wstr.c_str(), -1, utf8.data(), len, nullptr, nullptr);
+	return utf8;
```
uristdwarf: What is this exactly doing here? It's converting from multi-byte (I assume that means UTF8) to UTF-16, and then back to UTF-8?

rolysr: In this case this function converts from the current Windows ANSI code page (which is not UTF-8, but a legacy encoding like CP1252) to UTF-16 (wide string), and then from UTF-16 to UTF-8. This two-step process is required because Windows APIs do not provide a direct conversion from the ANSI code page to UTF-8. This ensures that strings from command-line arguments or file APIs are correctly handled as UTF-8 for cross-platform compatibility and proper Unicode support.

src/tools/dir_info.cc
rolysr: I’d like to use std::println for its convenience and type safety, but currently, support for <print> and std::println is not stable on Windows/MinGW (it requires experimental flag of corresponding library: "-lstdc++exp"). For now, I’ll stick with `printf` to ensure portability and avoid experimental dependencies. Once stable support lands in the standard library for all major toolchains it would be better to migrate to them. At the moment of this comment, I am using latest stable gcc version 15.2.0 (Rev8, Built by MSYS2 project)

src/tools/dir_info.cc
rolysr: Consider previous comment.

src/tools/tools_common_functions.h
```diff
@@ -122,3 +123,23 @@ inline uint32_t getGId() {
...
 #endif
 }
+
+#ifdef _WIN32
+// Converts a string from the current Windows ANSI code page (CP_ACP) to UTF-8.
+// This is necessary because, on Windows, command-line arguments and many file APIs
+// provide strings in the system's active code page, not UTF-8. To ensure consistent
+// Unicode handling across platforms and allow correct processing of non-ASCII input,
+// this function converts such strings to UTF-8 encoding.
+inline std::string utf8_from_cp(const char *cp_str) {
```
lgsilva3087: Since this function passes -1 as the source length, it assumes cp_str is
null-terminated. If it's ever called with a non-null-terminated buffer,
MultiByteToWideChar will read past the end of memory. Maybe clarify this
in the function name or documentation to make that assumption explicit.

lgsilva3087 (commented):
Please see my comment.

src/tools/tools_common_functions.h
rolysr: Changed function, please check newest version.

uristdwarf (approved):

src/common/encoding.h
```diff
@@ -0,0 +1,84 @@
+/*
+   Copyright 2013-2014 EditShare
+   Copyright 2013-2015 Skytechnology sp. z o.o.
```
dmga44: Please remove these lines from the header. This is a completely new file, right?

src/tools/master_functions.cc
```diff
@@ -44,6 +44,9 @@
 #include "protocol/matocl.h"
 #include "tools/tools_common_functions.h"
 
+// This import should be placed after other includes to avoid Windows dependency issues
+// with winsock2.h and windows.h
+#include "common/encoding.h"
 struct master_info_t {
```
dmga44: Consider adding a space in between the new header included and the rest of the code.

tests/test_suites/ShortSystemTests/test_saunafs_non_ascii_chars.sh
```diff
@@ -0,0 +1,34 @@
...
+# Run saunafs fileinfo on each file
+assert_success saunafs_command fileinfo "$polish/${polish}_plik"
+assert_success saunafs_command fileinfo "$chinese/${chinese}_file"
+assert_success saunafs_command fileinfo "$arabic/${arabic}_ملف"
+
+# Clean up
+rm "$polish/${polish}_plik"
+rm "$chinese/${chinese}_file"
+rm "$arabic/${arabic}_ملف"
+rmdir "$polish" "$chinese" "$arabic"
```
dmga44: Nice test :)

dmga44 (approved):

src/common/encoding.h
```diff
@@ -0,0 +1,84 @@
...
+struct Utf8Argv {
+	int argc;
+	char **argv;
+	std::vector<std::string> argvUtf8;
+	std::vector<char *> argvPtrs;
+
+	Utf8Argv() {
+		argvUtf8 = get_utf8_argv();
+		argvPtrs.reserve(argvUtf8.size() + 1);
+		for (auto &s : argvUtf8) argvPtrs.push_back(s.data());
```
lgsilva3087: Use braces, even if it is only one line.

src/tools/main.cc
```diff
@@ -58,13 +62,21 @@ static void print_prefix() {
...
+
+	Utf8Argv args;
+	argc = args.argc;
+	argv = args.argv;
+
 	socketinit();
 #endif
 
+	int status = 0;
+	set_humode();
```
lgsilva3087: Try to not move this lines if not needed.

tests/test_suites/ShortSystemTests/test_saunafs_non_ascii_chars.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Run saunafs fileinfo on each file
+assert_success saunafs_command fileinfo "$polish/${polish}_plik"
+assert_success saunafs_command fileinfo "$chinese/${chinese}_file"
+assert_success saunafs_command fileinfo "$arabic/${arabic}_ملف"
+
+# Clean up
+rm "$polish/${polish}_plik"
+rm "$chinese/${chinese}_file"
+rm "$arabic/${arabic}_ملف"
```
lgsilva3087: Please use braces consistently to expand the variables.

src/tools/tools_common_functions.h
```diff
@@ -28,6 +28,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string>
```
lgsilva3087: Is this needed?

src/common/encoding.h
```diff
@@ -0,0 +1,84 @@
...
+		WideCharToMultiByte(CP_UTF8, 0, argvW[i], -1, utf8.data(), size, nullptr, nullptr);
+		argvUtf8.push_back(std::move(utf8));
+	}
+	LocalFree(argvW);
+	return argvUtf8;
+}
+
+// Returns a pair: {argc, argv} for UTF-8 command-line arguments on Windows.
+// argvPtrs is kept alive by the returned object for the lifetime of main.
+struct Utf8Argv {
```
lgsilva3087: Consider to make it a class with proper member visibility, to force correct usage:

Something like:

```cpp
class Utf8Argv {
public:
	Utf8Argv() {
		argvUtf8 = get_utf8_argv();
		argvPtrs.reserve(argvUtf8.size() + 1);
		for (auto &s : argvUtf8) argvPtrs.push_back(s.data());
		argvPtrs.push_back(nullptr);
		argc = static_cast<int>(argvUtf8.size());
		argv = argvPtrs.data();
	}

	char **get_argv() { return argv; }

	int get_argc() { return argc; }

private:
	static std::vector<std::string> get_utf8_argv() {
		int argcW = 0;
		LPWSTR *argvW = CommandLineToArgvW(GetCommandLineW(), &argcW);
		std::vector<std::string> argvUtf8;
		argvUtf8.reserve(argcW);

		for (int i = 0; i < argcW; ++i) {
			int size = WideCharToMultiByte(CP_UTF8, 0, argvW[i], -1, nullptr, 0, nullptr, nullptr);
			std::string utf8(size - 1, '\0');
			WideCharToMultiByte(CP_UTF8, 0, argvW[i], -1, utf8.data(), size, nullptr, nullptr);
			argvUtf8.push_back(std::move(utf8));
		}
		LocalFree(argvW);
		return argvUtf8;
	}

	int argc;
	char **argv;
	std::vector<std::string> argvUtf8;
	std::vector<char *> argvPtrs;
};
```

src/common/encoding.h
lgsilva3087, line 1: Consider to rename the file to remind comand line arguments encoding, because encoding is too general concept.

src/common/encoding.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include <sys/stat.h>
+
+#ifdef _WIN32
+#include <shellapi.h>
+#include <windows.h>
+#include <string>
+#include <vector>
+
+inline std::vector<std::string> get_utf8_argv() {
```
lgsilva3087: Consider camel case instead of snakes.

src/common/encoding.h
lgsilva3087: Consider renaming it to:
Utf8CmdArguments or WindowsArgumentsAsUtf8.

src/common/encoding.h
```diff
@@ -0,0 +1,84 @@
...
+
+	std::wstring wpath(wlen, L'\0');  // include null terminator
+	MultiByteToWideChar(CP_UTF8, 0, path_utf8, -1, wpath.data(), wlen);
+
+	return _wstat64(wpath.c_str(), st);
+}
+
+// Unified portable wrapper
+inline int stat_portable(const char *path, struct stat *st) {
+	return utf8_stat(path, reinterpret_cast<struct _stat64 *>(st));
```
lgsilva3087: Can you always cast to _stat64 here?

lgsilva3087 (commented):
Please see my comments.

src/common/encoding.h
rolysr: Done.

src/tools/main.cc
rolysr: Done.

tests/test_suites/ShortSystemTests/test_saunafs_non_ascii_chars.sh
rolysr: Done.

src/tools/tools_common_functions.h
rolysr: Done.

src/common/encoding.h
rolysr: Done.

src/common/encoding.h
rolysr: Done.

src/common/encoding.h
rolysr: Done.

src/common/encoding.h
rolysr: Done.

src/common/encoding.h
rolysr: Done.

src/tools/master_functions.cc
rolysr: Done.

src/tools/main.cc
```diff
@@ -62,6 +66,16 @@ int main(int argc, char **argv) {
 	set_humode();
 
 #ifdef _WIN32
+	// Enable UTF-8 encoding for console I/O on Windows. This is necessary to properly
+	// handle Unicode characters in file paths and other strings on Windows systems,
+	// as the default console code page may not support UTF-8.
```
lgsilva3087, line 71: Nice comment.

src/common/args_stat_encoding.h
```diff
@@ -0,0 +1,102 @@
...
+			}
+			utf8.resize(size - 1);  // Remove the null terminator
+			argvUtf8.push_back(std::move(utf8));
+		}
+		LocalFree(argvW);
+		return argvUtf8;
+	}
+
+	int _argc = 0;
+	char **_argv = nullptr;
```
lgsilva3087: Maybe these two are not needed.

lgsilva3087 (approved):
Please see my minor comments.

src/common/encoding.h
rolysr: Done.

src/tools/main.cc
rolysr: Done.

src/common/args_stat_encoding.h
rolysr: Done.

PR #627, opened 2025-10-15
fix(packaging): Conditionally compile FDB modules

wg1k:
Some packaging was failing when the `ENABLE_FOUNDATIONDB` was unset.

This commit ensures that we honor the intended config, and preempt the dependencies binding when it is not intended.


rolysr (approved):
LGTM.

dmga44 (approved):

PR #651, opened 2025-10-28
refactor(master): make FS checksum functions extensible

lgsilva3087:
IFilesystemOperations interface now provides the functions `fs_checksum` and `fs_start_checksum_recalculation`, making them extensible through inheritance. To be more explicit, the functions with the same name in filesystem_checksum.h were placed in the new `checksum` namespace.

The references to these functions were updated to use the polymorphic gFilesystemOperations. Custom operations just need to provide their own implementation.


ralcolea (approved):
Very nice job @lgsilva3087!! 👍 🔥 💪

dmga44 (approved):

PR #652, opened 2025-10-28
chore(protocol,master): remove warnings from cltoma

lgsilva3087:
This commit modernizes a little bit the file cltoma.h to remove the linter warnings and to make the code more readable. As a side effect, the casts in matoclserv.cc are also no longer needed.

### DIFF
--- a/src/protocol/cltoma.h
+++ b/src/protocol/cltoma.h
@@ -23,6 +23,7 @@
 #include "common/platform.h"
 
 #include <array>
+#include <cstdint>
 
 #include "common/access_control_list.h"
 #include "common/acl_type.h"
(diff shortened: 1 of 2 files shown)


src/protocol/cltoma.h
```diff
@@ -314,7 +315,7 @@ SAUNAFS_DEFINE_PACKET_SERIALIZATION(
 SAUNAFS_DEFINE_PACKET_VERSION(cltoma, manageLocksList, kAll, 0)
 SAUNAFS_DEFINE_PACKET_VERSION(cltoma, manageLocksList, kInode, 1)
 
-#define SAU_CLTOMA_MANAGE_LOCKS_LIST_LIMIT 1024
+constexpr uint64_t SAU_CLTOMA_MANAGE_LOCKS_LIST_LIMIT = 1024;
```
rolysr, line 318: Consider renaming this variable to `kSauCltomaLocksListLimit`.

rolysr (approved):
LGTM. Please check my suggestion if applicable.

src/protocol/cltoma.h
lgsilva3087: I was thinking the same, I just didn't change it because in this case matches very closely the name of the related packet. I think at some point we'll do an intensive course to normalize them all.

dmga44 (approved):
LGTM.

PR #660, opened 2025-10-30
feat(kv,fdb): add async get operation support

lgsilva3087:
Implement asynchronous value retrieval for FoundationDB transactions to enable non-blocking operations. This introduces the IFuture interface and FDBFutureValue implementation for handling async results.

Changes include:
- Add IFuture interface for async operation results.
- Implement FDBFutureValue wrapping FDB futures with RAII.
- Add getAsync() methods to ITransaction and implementations.
- Extract FDB_API_VERSION to separate header file.
- Make transaction interfaces non-copyable and non-movable.
- Add unit tests for async operations.

The async interface allows multiple concurrent reads without blocking, improving performance for workloads with multiple independent key lookups.

Reviewers: I am already using it in the MDS.

### DIFF
--- a/src/fdb/fdb_unittest.cc
+++ b/src/fdb/fdb_unittest.cc
@@ -21,6 +21,7 @@
 #include "fdb/fdb.h"
 #include "fdb/fdb_context.h"
 #include "fdb/fdb_kv_engine.h"
+#include "kv/ifuture.h"
 #include "kv/itransaction.h"
 
 #include <gtest/gtest.h>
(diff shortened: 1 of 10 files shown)


antuan96314 (approved):

src/fdb/fdb_unittest.cc
```diff
@@ -140,6 +141,8 @@ TEST_F(FDBKVEngineTest, GetRange) {
 	testGetRange(kvEngine.get(), kStart, kEnd, false, false, kNumKeys);
 }
 
+
+
```
rolysr: Extra spaces here.

rolysr (approved):
LGTM.

src/fdb/fdb_unittest.cc
lgsilva3087: Applied.

PR #680, opened 2025-11-20
ci: Re-enable unit tests in Github Actions

ralcolea:
FoundationDB unit tests started failing previously due to disk space exhaustion on the GitHub Actions runner after building SaunaFS. As a result, unit tests were moved to Jenkins and a dedicated VM was provisioned to run them.
    
This PR introduces the following changes:
- Adds a cleanup stage that frees sufficient disk space on the runner, preventing future FDB test failures caused by space depletion.
- Re-enable unit tests in GitHub Actions.

Signed-off-by: Crash <<crash@leil.io>>

### DIFF
--- a/.github/workflows/run-unit-and-ganesha-tests.yml
+++ b/.github/workflows/run-unit-and-ganesha-tests.yml
@@ -1,4 +1,4 @@
-name: Run Ganesha Tests
+name: Run Unit and Ganesha Tests
 
 on:
     merge_group:
@@ -8,7 +8,7 @@ on:
             - dev
 
 jobs:
-    Run-Ganesha-Tests:
+    Run-Unit-Tests-and-Ganesha:
         runs-on: ubuntu-24.04
 
         env:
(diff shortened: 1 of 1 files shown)


rolysr (approved):
LGTM.

.github/workflows/run-unit-and-ganesha-tests.yml
```diff
@@ -31,6 +31,19 @@ jobs:
...
+            - name: Free Disk Space (Ubuntu)
+              uses: jlumbroso/free-disk-space@main
+              with:
+                  # this might remove tools that are actually needed,
+                  # if set to "true" but frees about 6 GB
+                  tool-cache: false
+                  android: true              # ~14GB
+                  dotnet: true               # ~3GB
+                  haskell: true              # ~8GB
+                  large-packages: true       # ~5GB (Chrome, Firefox, etc)
```
lgsilva3087: Consider to disable this one, looks like the it is the longest running, but keep a comment to enable it if needed again more space.

.github/workflows/run-unit-and-ganesha-tests.yml
```diff
@@ -31,6 +31,19 @@ jobs:
...
+                  docker-images: true        # ~5GB
```
lgsilva3087: We could disable this one too and keep the comment, right now it not releasing space.

lgsilva3087 (approved):
Great. LGTM. We can keep it even if we decide to have the unit tests in Jenkins.

.github/workflows/run-unit-and-ganesha-tests.yml
ralcolea: Nice suggestion! Applied.

.github/workflows/run-unit-and-ganesha-tests.yml
ralcolea: I agree. Applied.

PR #682, opened 2025-11-25
fix(cs): Fix test force read timeouts flakiness

dmga44:
This change targets fixing the instability of the test
test_chunkserver_force_timeout_reads. The test failed due to sudden
stop of one or several of the chunkservers, causing failed checks in
the end or timeout.

The chunkserver stop seems to be related to inconsistencies in the
assumptions of the jobs to be discarded. Those jobs needs to be not
processed at the moment of adding them to the discarded lists in order
to make pendingDelayedJobs_ hold the correct value. The proposed
solution substitutes the status member of the OutputBuffer for a
isProcessed flag to better maintain whether the callback related to the
OutputBuffer has already started or not.

Side changes:
- the ReadHighLevelOp callbacks will use the second parameter to update
the related OutputBuffer.
- since there will be no status updates from the pool worker threads,
there is no need for a mutex member in OutputBuffer.

### DIFF
--- a/src/chunkserver/chunk_high_level_ops.cc
+++ b/src/chunkserver/chunk_high_level_ops.cc
@@ -68,13 +68,16 @@ void GetBlocksHighLevelOp::delayedClose() {
 	pendingDelayedJobs_++;
 }
 
-void ReadHighLevelOp::delayedCloseCallback(uint8_t status, void * /*entry*/) {
+void ReadHighLevelOp::delayedCloseCallback(uint8_t status, void *buffer) {
 	assert(getParentState() == ChunkserverEntry::State::CloseWait);
+	auto outputBuffer = static_cast<OutputBuffer *>(buffer);
+	passert(outputBuffer);
+	outputBuffer->setIsCallbackStarted(true);
 
 	if (status == SAUNAFS_STATUS_OK) { isChunkOpen_ = true; }
 
 	while (!toDiscardReadDataBuffers_.empty() &&
-	       toDiscardReadDataBuffers_.front()->getStatus() != kNotSaunafsStatus) {
+	       toDiscardReadDataBuffers_.front()->isCallbackStarted()) {
 		getReadOutputBufferPool().put(std::move(toDiscardReadDataBuffers_.front()));
 		toDiscardReadDataBuffers_.pop_front();
 		toDiscardReadJobIds_.pop_front();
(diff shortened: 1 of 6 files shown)


src/chunkserver/io_buffers.h
```diff
@@ -306,11 +303,8 @@ class OutputBuffer {
 	/// The number of blocks.
 	const size_t numBlocks_;
 
-	/// Protects the `status` member variable used for custom thread synchronization.
-	std::mutex mutex_;
-
-	/// Status of the buffer's related read operation.
-	uint8_t status{kNotSaunafsStatus};
+	/// Whether the buffer's related callback has started processing or has been processed.
+	bool isProcessed_{false};
```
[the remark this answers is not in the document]
dmga44: All of those functions run in the same thread.

src/chunkserver/io_buffers.h
```diff
@@ -284,16 +282,15 @@ class OutputBuffer {
 	/// @brief Clears the buffer.
 	void clear();
 
-	/// @brief Returns the `status` in a thread-safe manner.
-	uint8_t getStatus() {
-		std::lock_guard<std::mutex> lock(mutex_);
-		return status;
+	/// @brief Returns the `isProcessed` flag.
+	/// Whether the buffer's related callback has started processing or has been processed.
+	bool isProcessed() const {
```
lgsilva3087: Consider:
isCallbackStarted()

src/chunkserver/io_buffers.h
lgsilva3087: Similar here.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -68,13 +68,14 @@ void GetBlocksHighLevelOp::delayedClose() {
 	pendingDelayedJobs_++;
 }
 
-void ReadHighLevelOp::delayedCloseCallback(uint8_t status, void * /*entry*/) {
+void ReadHighLevelOp::delayedCloseCallback(uint8_t status, void *buffer) {
 	assert(getParentState() == ChunkserverEntry::State::CloseWait);
+	auto outputBuffer = static_cast<OutputBuffer *>(buffer);
```
lgsilva3087, line 73: Consider an assert that the cast was successful

lgsilva3087 (approved):

rolysr (approved):
LGTM.

ralcolea (approved):
Great job @dmga44!!! 💪 🔥 🚀

src/chunkserver/io_buffers.h
dmga44: Applied.

src/chunkserver/io_buffers.h
dmga44: Applied.

src/chunkserver/chunk_high_level_ops.cc
dmga44: Applied.

PR #690, opened 2025-12-03
fix: Fix rebalancing related tests

GigaCronos:
#### This commit fix the following tests by adjusting parameters:
- `test_chunk_rebalancing`
- `test_custom_goals_generate_no_accidental_rebalancing`

The other rebalancing tests are not failing.

Modified `chunks.cc` to allow small parameters for `HashCPS`, specifically for testing purposes.
Modified the Jenkinsfile to create a new test suite for testing with loop devices (focusing on rebalance for now).

#### **The tests were failing due to mechanisms that cause the `ChunkWorker` rebalancing logic to use outdated disk usage data:**
- Polling Delay (50ms): The Master's disk usage info is not real-time, causing decisions based on old states.
- Processing Snapshot: The `ChunkWorker::mainLoop` uses a fixed disk usage snapshot for up to 1 second. During this time, its own actions change the real state, making subsequent decisions increasingly inaccurate.

**Test Impact:** This data staleness creates false imbalance readings. One test fails due to imprecise rebalancing, and the other fails from triggering unnecessary operations based on outdated data.

Co-authored-by: Urmas Rist <urmas@leil.io>


Jenkinsfile
```diff
@@ -180,15 +180,15 @@ pipeline {
         }
         stage('Build and test') {
             parallel {
-                stage('Build and run unittests (Ubuntu 24.04)') {
+                stage('Run special tests (Ubuntu 24.04)') {
```
antuan96314, line 183: Why is this needed?

Jenkinsfile
```diff
@@ -180,15 +180,15 @@ pipeline {
...
                     agent {label 'unittests'}
                     stages {
                         stage("Checkout source") {
                             steps {
                                 checkout scm
                             }
                         }
-                        stage("Build and test") {
+                        stage("Build and unit test") {
```
antuan96314, line 191: Why is this needed?

antuan96314 (approved):

Jenkinsfile
uristdwarf: This is because the tests don't work with the existing sfstests tool, due to requiring access to creating and using loop devices, which are kernel level. So it cannot be run in parallel and hard to get it working on Docker (I think it's possible, but I haven't managed it).

Jenkinsfile
uristdwarf: See above

tests/test_suites/RebalancingTests/test_chunk_rebalancing.sh
```diff
@@ -8,9 +8,11 @@ CHUNKSERVERS=5 \
 	MASTER_EXTRA_CONFIG="CHUNKS_LOOP_MIN_TIME = 1`
 			`|CHUNKS_LOOP_MAX_CPU = 90`
 			`|CHUNKS_WRITE_REP_LIMIT = 1`
+			`|CHUNKS_LOOP_MAX_CPS = 2`
 			`|CHUNKS_READ_REP_LIMIT = 2`
 			`|OPERATIONS_DELAY_INIT = 0`
 			`|OPERATIONS_DELAY_DISCONNECT = 0`
+			`|POLL_TIMEOUT_MS = 10`
 			`|ACCEPTABLE_DIFFERENCE = 0.0015" \
```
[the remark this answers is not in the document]
GigaCronos: Done

dage-leil:
LS-39

rolysr (approved):
LGTM.

rolysr:
@GigaCronos nice work on putting these tests back.

PR #691, opened 2025-12-03
fix(master): Remove unused case insensitive check

rolysr:
Adding lowercase entries for each inode and its children during edge parsing is unnecessary when loading metadata as part of filesystem initialization on the master. At that point, there is no context about case-insensitive client connections, so the  case-insensitive check for each inode and its children always  evaluates to false and the code never executes. This commit removes that dead code.


lgsilva3087 (approved):
LGTM

dmga44 (approved):
LGTM.

PR #717, opened 2026-01-22
refactor(master): allow stats operations in kv backends

lgsilva3087:
Propagate transaction context to stats-related operations (getStats,
addStats, subStats, and addSubStats) to enable efficient and consistent
transactional behavior for alternative KV backends.

Main changes:
- Added FilesystemOperationContext parameter to all stats operations,
  allowing them to access parent nodes via idToNodeVerify() when
  recursively propagating stat changes up the directory tree
- Introduced extractIntegralLEFromFuture utility in the KV library to
  parallelize retrieval of the 8 directory statistics using getAsync
- Refactored KV types into separate header (kv_types.h) to resolve
  circular dependencies between ifuture.h and kv_utils.h
- Cascading updates throughout the filesystem layer due to the
  fundamental nature of getStats() and getSize() operations

Since getStats() and related operations are called throughout the
codebase, adding the FilesystemOperationContext parameter required
propagating changes through numerous functions across multiple layers.

The KV backend now gains the ability to:
- Perform directory statistics operations within a single transaction
- Efficiently traverse the filesystem hierarchy when updating ancestor
  directories
- Read cached statistics using the provided transaction context

Notes:
- The in-memory backend remains unaffected by these changes
- Some read-write transactions were intentionally left uncommitted in
  this refactoring; commits will be added as related KV operations are
  implemented
- Integration tests and the FDB implementation are located in the MDS
  repository

Related to LS-242

### DIFF
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -288,7 +288,9 @@ uint8_t FilesystemOperationsBase::undel(const FsContext &context, inode_t inode)
 	return status;
 }
 
-uint8_t FilesystemOperationsBase::purge(const FsContext &context, inode_t inode) {
+uint8_t FilesystemOperationsBase::purge(const FsContext &context,
+                                        const FilesystemOperationContext &fsOpContext,
+                                        inode_t inode) {
 	ChecksumUpdater cu(context.ts());
 	FSNode *p;
 	uint8_t status =
@@ -297,8 +299,6 @@ uint8_t FilesystemOperationsBase::purge(const FsContext &context, inode_t inode)
 		return status;
 	}
 
-	auto fsOpContext = gFSOperations->createFilesystemOperationContext(
-	    FilesystemOperationContext::TransactionType::kReadWrite);
 	status = nodeOperations_->getNodeForOperation(context, fsOpContext, ExpectedNodeType::kAny,
 	                                              MODE_MASK_EMPTY, inode, &p);
 	if (status != SAUNAFS_STATUS_OK) {
@@ -308,7 +308,7 @@ uint8_t FilesystemOperationsBase::purge(const FsContext &context, inode_t inode)
 	}
 	// This should be equal to inode, because p is not a directory
 	inode_t purged_inode = p->id;
-	nodeOperations_->purge(context.ts(), p);
+	nodeOperations_->purge(fsOpContext, context.ts(), p);
 
 	if (context.isPersonalityMaster()) {
 		changeLog(context.ts(), "PURGE(%" PRIiNode ")", purged_inode);
@@ -374,9 +374,10 @@ uint8_t FilesystemOperationsBase::getRootInode(inode_t *rootinode, const uint8_t
 	}
 }
 
-void FilesystemOperationsBase::statfs(const FsContext &context, uint64_t *totalspace,
-                                      uint64_t *availspace, uint64_t *trspace, uint64_t *respace,
-                                      inode_t *inodes) {
+void FilesystemOperationsBase::statfs(const FsContext &context,
+                                      const FilesystemOperationContext &fsOpContext,
+                                      uint64_t *totalspace, uint64_t *availspace, uint64_t *trspace,
+                                      uint64_t *respace, inode_t *inodes) {
 	FSNode *rn;
 	StatsRecord sr;
 	if (context.rootinode() == SPECIAL_INODE_ROOT) {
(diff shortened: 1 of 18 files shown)


src/master/filesystem_operations.h
```diff
@@ -75,19 +75,22 @@ class FilesystemOperationsBase : public IFilesystemOperations {
 	// Common for metarestore and master server (both personalities)
 
 	uint8_t acquire(const FsContext &context, inode_t inode, uint32_t sessionid) override;
-	uint8_t append(const FsContext &context, inode_t inode, inode_t inode_src) override;
+	uint8_t append(const FsContext &context, const FilesystemOperationContext &fsOpContext,
```
antuan96314, line 78: Why modifying method signature to add a second (offset) parameter instead of first parameter to make it consistence with changes to other signatures above?

antuan96314 (approved):

src/master/filesystem_operations.h
lgsilva3087: Hi, the pattern was as follows:
In FilesystemOperations family:
- Add the new OperationContext always after the existing FsContext, as most of the operations already receive it.

In FilesystemNodeOperations family:
- If the operation already receives a FsContext, add the new one after.
- If the operation does not reveive a FsContext, add it as first parameter, because most of the functions will need it.

Please notice that the previous line `acquire` is not yet modified for KV backends, but it will follow the pattern when needed.

src/master/filesystem_operations.cc
```diff
@@ -2652,16 +2657,18 @@ uint8_t FilesystemOperationsBase::applyRepair(uint32_t timestamp, inode_t inode,
 	if (indx > kMaxChunkIndex) { return SAUNAFS_ERROR_INDEXTOOBIG; }
 
 	if (indx >= p->chunks.size()) {
+		safs::log_err("fs_apply_repair: indx {} is greater than number of chunks ({}), inode {}",
+		              indx, p->chunks.size(), inode);
 		return SAUNAFS_ERROR_NOCHUNK;
-		safs::log_err("fs_apply_repair: indx {} is greater than number of chunks ({}), inode {}", indx, p->chunks.size(), inode);
 	}
```
dmga44, line 2663: Nice 👍

src/master/matoclserv.cc
```diff
@@ -3504,7 +3536,19 @@ void matoclserv_fuse_append(matoclserventry *eptr, const uint8_t *data, uint32_t
...
+		auto fsOpContext = gFSOperations->createFilesystemOperationContext(
+		    FilesystemOperationContext::TransactionType::kReadWrite);
+
+		status = gFSOperations->append(context, fsOpContext, inode, inode_src);
+
+		if (status == SAUNAFS_STATUS_OK && fsOpContext.hasReadWriteTransaction()) {
+			if (!fsOpContext.getReadWriteTransaction()->commit()) {
+				safs::log_err(
+				    "matoclserv_fuse_append: transaction failed to commit: inode {}, source inode {}",
+				    inode, inode_src);
```
dmga44: Consider:
```
				safs::log_err("({}): transaction failed to commit: inode {}, source inode {}",
				              __func__, inode, inode_src);
```

src/master/restore.cc
```diff
@@ -219,7 +219,22 @@ int do_append(const char *filename, uint64_t lv, uint32_t ts, const char *ptr) {
...
+	int status = gFSOperations->append(FsContext::getForRestore(ts), fsOpContext, inode, inode_src);
+
+	if (status == SAUNAFS_STATUS_OK && fsOpContext.hasReadWriteTransaction()) {
+		if (!fsOpContext.getReadWriteTransaction()->commit()) {
+			safs::log_err(
+			    "do_append: transaction failed to commit: inode {}, source inode {}",
+			    inode, inode_src);
+			status = SAUNAFS_ERROR_IO;
+		}
+	}
```
dmga44, line 234: Like in many other places that you added this check, can the status be changed into something else? I'd propose to create a new status code, some other that reflects better this type of error.

dmga44 (approved):
LGTM, please consider my minor suggestions.

src/master/restore.cc
lgsilva3087: I will do it while implementing the transaction retry strategy. Should be like SAUNAFS_ERROR_TRANSACTION_COMMIT_FAIL.

src/master/matoclserv.cc
lgsilva3087: Applied.

ralcolea (commented):
LGTM

PR #726, opened 2026-02-02
refactor(master): persist write operations in KV backends

lgsilva3087:
When using KV backends like FDB, metadata changes must be explicitly committed to be persisted. This change adds transaction commit calls to the write related operations.

The implementation is backward compatible with the in-memory backend, transaction commits are only executed when a read-write transaction is available, ensuring seamless operation across both in-memory and KV-based implementations.

Some blank lines were also added for readability.

Related to: LS-189

### DIFF
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -371,6 +371,16 @@ FSNodeDirectory *FilesystemNodeOperationsBase::getFirstParent(FSNode *node) {
 	return gMetadata->root;
 }
 
+std::vector<inode_t> FilesystemNodeOperationsBase::getParentIds(
+    [[maybe_unused]] const FilesystemOperationContext &fsOpContext, FSNode *node) {
+	assert(node);
+
+	std::vector<inode_t> parentIds;
+	parentIds.reserve(node->parents.size());
+	for (const auto &parent : node->parents) { parentIds.push_back(parent.first); }
+	return parentIds;
+}
+
 void FilesystemNodeOperationsBase::subStats(const FilesystemOperationContext &fsOpContext,
                                             FSNodeDirectory *parent, StatsRecord *stats) {
 	if (parent != nullptr) {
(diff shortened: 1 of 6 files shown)


src/master/matoclserv.cc
```diff
@@ -2852,11 +2852,20 @@ void matoclserv_fuse_write_chunk(matoclserventry *eptr, PacketHeader header, con
...
 	                                   min_server_version);
 
+	if (fsOpContext.hasReadWriteTransaction()) {
+		if (!fsOpContext.getReadWriteTransaction()->commit()) {
+			safs::log_err("{}: transaction failed to commit: inode {}, chunk index {}", __func__,
+			              inode, chunkIndex);
+
+			status = SAUNAFS_ERROR_IO;
+		}
+	}
```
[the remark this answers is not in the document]
lgsilva3087: True, added.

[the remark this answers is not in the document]
lgsilva3087: Truncate operations will be handled in dedicated commit/PR.

src/master/filesystem_node.cc
```diff
@@ -371,6 +371,15 @@ FSNodeDirectory *FilesystemNodeOperationsBase::getFirstParent(FSNode *node) {
 	return gMetadata->root;
 }
 
+std::vector<inode_t> FilesystemNodeOperationsBase::getParentIds(
+    [[maybe_unused]] const FilesystemOperationContext &fsOpContext, FSNode *node) {
+	assert(node);
+
+	std::vector<inode_t> parentIds;
```
uristdwarf, line 378: Is it possible to know the size of the amount of node->parents without counting? If so, it's better to do initially allocate size for the vector.

Even if not, allocating a 4 KB (the page size on most systems) is a good start.

uristdwarf: Also note that this is copying memory. If there's a better way to give the parentIDs, it should be done if possible

lgsilva3087: Very nice catch @uristdwarf.

The size is known in the in-memory implementation without counting, I added it to that implementation (reserve). For FDB, in the current state we need to ask the database (until we add caching).

About copying:
- I added the new virtual function updateParentStatsForNode, which in the in-memory implementation just traverses the existing parents container (no copy). I will keep getParentIds for now.
- For the KV backend, returning a vector should not be a problem due to RVO/NRVO.

uristdwarf (approved):

ralcolea (approved):
Great job @lgsilva3087! 👍 💪 🔥

PR #735, opened 2026-02-06
ci: Fix Jenkins unable to delete build files

uristdwarf:
Due to a bug where build files would be root owned after build, Jenkins could not later clean them up. This commit works around that by chowning the directory recursively as jenkins user.


rolysr (approved):

dmga44 (approved):
Thank you, great job.

lgsilva3087 (approved):

PR #740, opened 2026-02-09
fix: Delayed chunk operations missing from shadow changelogs

uristdwarf:
When master begins shutting down, it will end sessions with the metaloggers/shadows. However, if there are delayed chunk operations that are still performed and they are successful, it will send to client that these have succeeded and bump the metadata version, but the metaloggers and shadows will be unaware of these changes. This can cause problems with uraft for example, when a client tries now connecting to a shadow that became master, and expects operations that it did to be done on it, causing metadata loss.

This fixes that issue by ensuring that it will end the session only when all async operations affecting the metadata are finished.

This still leaves the possibility that some metadata will be lost if master crashes or disappears abnormally (as changes to shadows/metaloggers are send async), but it should fix it for normal shutdowns.

Closes LS-317

### DIFF
--- a/src/master/matoclserv.cc
+++ b/src/master/matoclserv.cc
@@ -5804,16 +5804,25 @@ void matoclserv_wantexit() {
 	exiting = 1;
 }
 
+bool matoclserv_client_async_operations_finished() {
+	for (const auto &eptr : matoclservList) {
+		if (!eptr->delayedChunkOperations.empty()) {
+			return false;
+		}
+	}
+	return true;
+}
+
 int matoclserv_canexit() {
 	matoclserventry *adminTerminator = nullptr;
 	static bool terminatorPacketSent = false;
 
+	if (!matoclserv_client_async_operations_finished()) {
+		return 0;
+	}
+
 	for (const auto &eptr : matoclservList) {
 		if (!eptr->outputPackets.empty()) {
-			return 0;
-		}
-
-		if (!eptr->delayedChunkOperations.empty()) {
 			return 0;
 		}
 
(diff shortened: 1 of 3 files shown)


src/master/matoclserv.cc
```diff
@@ -5635,19 +5635,28 @@ void matoclserv_wantexit() {
...
 }
 
+bool matoclserv_client_async_operations_finished() {
+	for (const auto &eptr : matoclservList) {
+		if (!eptr->delayedChunkOperations.empty()) {
+			return false;
+		}
+	}
+	return true;
+}
```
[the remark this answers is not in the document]
uristdwarf: I don't think this really improves readability

src/master/matoclserv.cc
```diff
@@ -5635,19 +5635,28 @@ void matoclserv_wantexit() {
...
 	for (const auto &eptr : matoclservList) {
 		if (!eptr->outputPackets.empty()) {
 			return 0;
 		}
 
-		if (!eptr->delayedChunkOperations.empty()) {
-			return 0;
-		}
-
 		if (eptr->adminTask == AdminTask::kTerminate) {
```
dmga44, line 5829: Considering both ways of implementing it have the same result, why use the new function?

dmga44 (approved):
LGTM.

src/master/matoclserv.cc
uristdwarf: I guess I want to reuse the logic in the function. Maybe there will be more async operations that need finishing in the future.

dmga44: Makes sense to me.

lgsilva3087 (approved):
LGTM

ralcolea (approved):
LGTM!

PR #759, opened 2026-02-21
fix(cs): fix flaky test_helgrind_basic

dmga44:
This change targets improving the stability of the test_helgrind_basic. The test was failing due missing protection of the bgJobPool member of the network worker class when stopping the chunkserver. The solution proposed is to avoid accessing that member when it is no longer needed.

The test_helgrind_basic was modified to explicitly test the stop mechanism.


lgsilva3087 (approved):

rolysr (approved):
LGTM

PR #762, opened 2026-02-24
refactor(master): allow ACL ops in KV backends

lgsilva3087:
Thread FilesystemOperationContext through all ACL methods (deleteAcl, setAcl, getAcl, applySetAcl, applySetRichAcl) in both IFilesystemNodeOperations and IFilesystemOperations, removing the internal transaction creation that was hardcoded inside each FilesystemOperationsBase method body.

Callers (matoclserv, restore) now own the transaction lifecycle, enabling KV backends to participate in ACL operations as part of a single coordinated transaction.

Introduce a protected virtual `getAclForAccess()` method to abstract ACL retrieval inside `access()`. The default in-memory implementation reads from `aclStorage`; KV backends can override it to materialise the ACL from the key-value store.

Pass `FilesystemOperationContext` through `access()` at all call sites so the new virtual hook receives the context it needs. A caller-supplied `RichACL scratch` buffer is provided so temporary ACLs can be returned without extra allocation.

The in-memory base implementation is not affected by the changes.

Also document all affected interface declarations according to their implementations.

Related to: LS-70

### DIFF
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -1938,7 +1938,9 @@ void FilesystemNodeOperationsBase::setExtraAttrRecursive(FSNode *node, uint32_t 
 	fsnodes_update_checksum(node);
 }
 
-uint8_t FilesystemNodeOperationsBase::deleteAcl(FSNode *node, AclType type, uint32_t timeStamp) {
+uint8_t FilesystemNodeOperationsBase::deleteAcl(
+    [[maybe_unused]] const FilesystemOperationContext &fsOpContext, FSNode *node, AclType type,
+    uint32_t timeStamp) {
 	if (type == AclType::kRichACL) {
 		gMetadata->aclStorage.erase(node->id);
 	} else if (type == AclType::kDefault) {
@@ -1979,7 +1981,8 @@ uint8_t FilesystemNodeOperationsBase::deleteAcl(FSNode *node, AclType type, uint
 }
 
 #ifndef METARESTORE
-uint8_t FilesystemNodeOperationsBase::getAcl(FSNode *node, RichACL &acl) {
+uint8_t FilesystemNodeOperationsBase::getAcl(
+    [[maybe_unused]] const FilesystemOperationContext &fsOpContext, FSNode *node, RichACL &acl) {
 	const RichACL *richAcl = gMetadata->aclStorage.get(node->id);
 
 	if (!richAcl) { return SAUNAFS_ERROR_ENOATTR; }
@@ -1991,7 +1994,9 @@ uint8_t FilesystemNodeOperationsBase::getAcl(FSNode *node, RichACL &acl) {
 }
 #endif
 
-uint8_t FilesystemNodeOperationsBase::setAcl(FSNode *node, const RichACL &acl, uint32_t timeStamp) {
+uint8_t FilesystemNodeOperationsBase::setAcl(
+    [[maybe_unused]] const FilesystemOperationContext &fsOpContext, FSNode *node,
+    const RichACL &acl, uint32_t timeStamp) {
 	if (!acl.checkInheritFlags(node->type == FSNodeType::kDirectory)) {
 		return SAUNAFS_ERROR_ENOTSUP;
 	}
@@ -2021,8 +2026,9 @@ uint8_t FilesystemNodeOperationsBase::setAcl(FSNode *node, const RichACL &acl, u
 	return SAUNAFS_STATUS_OK;
 }
 
-uint8_t FilesystemNodeOperationsBase::setAcl(FSNode *node, AclType type,
-                                             const AccessControlList &acl, uint32_t timeStamp) {
+uint8_t FilesystemNodeOperationsBase::setAcl(
+    [[maybe_unused]] const FilesystemOperationContext &fsOpContext, FSNode *node, AclType type,
+    const AccessControlList &acl, uint32_t timeStamp) {
 	if (type != AclType::kDefault && type != AclType::kAccess) {
 		return SAUNAFS_ERROR_EINVAL;
 	}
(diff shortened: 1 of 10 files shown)


src/master/filesystem_node.h
```diff
@@ -174,13 +174,28 @@ class FilesystemNodeOperationsBase : public IFilesystemNodeOperations {
...
-	uint8_t setAcl(FSNode *node, AclType type, const AccessControlList &acl,
-	               uint32_t timeStamp) override;
+
+	/// Stores a RichACL on a node, replacing any previously stored ACL.
+	/// @see IFilesystemNodeOperations::setAcl
+	uint8_t setAcl(const FilesystemOperationContext &fsOpContext, FSNode *node,
+	               const RichACL &acl, uint32_t timeStamp) override;
+
+	/// Merges a POSIX ACL into the node's stored RichACL.
+	/// @see IFilesystemNodeOperations::getAcl
```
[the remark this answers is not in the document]
lgsilva3087: Applied.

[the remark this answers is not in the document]
lgsilva3087: Applied.

src/master/restore.cc
```diff
@@ -804,10 +832,23 @@ int do_setrichacl(const char *filename, uint64_t lv, uint32_t ts, const char *pt
...
+	auto fsOpContext = gFSOperations->createFilesystemOperationContext(
+	    FilesystemOperationContext::TransactionType::kReadWrite);
+
+	int status = gFSOperations->applySetRichAcl(fsOpContext, ts, inode,
+	                                            reinterpret_cast<const char *>(acl_string));
+
+	if (status == SAUNAFS_STATUS_OK && fsOpContext.hasReadWriteTransaction()) {
+		if (!fsOpContext.getReadWriteTransaction()->commit()) {
+			safs::log_err("{}: transaction failed to commit: inode {}", __func__, inode);
+			status = SAUNAFS_ERROR_IO;
```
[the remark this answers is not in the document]
lgsilva3087: Applied. But please remember that the transaction retry strategy (to be implemented) will modify this commits and errors checking.

src/master/matoclserv.cc
```diff
@@ -4356,14 +4368,17 @@ void matoclserv_fuse_getacl(matoclserventry *eptr, const uint8_t *data, uint32_t
 
 	uint8_t status = matoclserv_check_group_cache(eptr, gid);
 
+	auto fsOpContext = gFSOperations->createFilesystemOperationContext(
+	    FilesystemOperationContext::TransactionType::kReadOnly);
+
```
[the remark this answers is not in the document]
lgsilva3087: Fixed.

src/master/filesystem_node.cc
```diff
@@ -2076,12 +2082,21 @@ int FilesystemNodeOperationsBase::nameCheck(const std::string &name) {
...
+int FilesystemNodeOperationsBase::access(const FsContext &context,
+                                         const FilesystemOperationContext &fsOpContext,
+                                         FSNode *node, uint8_t modeMask) {
 	if ((context.sesflags() & SESFLAG_NOMASTERPERMCHECK) || context.uid() == 0) {
 		return 1;  // super user or no permission check
 	}
 
-	const RichACL *nodeAcl = gMetadata->aclStorage.get(node->id);
+	RichACL aclScratch;  // unused by this implementation, needed for KV backends compatibility
+	const RichACL *nodeAcl = getAclForAccess(fsOpContext, node, aclScratch);
```
dmga44: Consider passing (to getAclForAccess) a pointer to a RichAcl instance to avoid constructing a new one in each call. If that would be the case, we'd need to pass here a nullptr.

src/master/filesystem_operations.cc
```diff
@@ -3331,32 +3328,29 @@ uint8_t FilesystemOperationsBase::setAcl(const FsContext &context, inode_t inode
 		gMetadata->metadataVersion++;
 	}
 	return status;
-
-	return SAUNAFS_ERROR_EINVAL;
```
dmga44, line 3335: Nice catch 👍 .

src/master/filesystem_operations.cc
```diff
@@ -3391,11 +3387,12 @@ uint8_t FilesystemOperationsBase::applySetRichAcl(uint32_t timestamp, inode_t in
 	if (!p) {
 		return SAUNAFS_ERROR_ENOENT;
 	}
-	uint8_t status = nodeOperations_->setAcl(p, std::move(acl), timestamp);
-	if (status == SAUNAFS_STATUS_OK) {
-		gMetadata->metadataVersion++;
-	}
-	return status;
+
+	uint8_t status = nodeOperations_->setAcl(fsOpContext, p, acl, timestamp);
```
rolysr, line 3391: Also check here for the `std::move()` deletion and other occurences.

src/master/restore.cc
```diff
@@ -804,10 +832,23 @@ int do_setrichacl(const char *filename, uint64_t lv, uint32_t ts, const char *pt
...
+			safs::log_err("{}: transaction failed to commit: inode {}", __func__, inode);
+			return SAUNAFS_ERROR_IO;
+		}
+	}
+
+	return status;
 }
 
-int do_setquota(const char *filename, uint64_t lv, uint32_t, const char *ptr) {
+int do_setquota(const char *filename, uint64_t lv, uint32_t /*ts*/, const char *ptr) {
```
rolysr, line 851: Is this comment PR related also?

rolysr (approved):
LGTM. Please check my suggestions.

dmga44 (approved):
LGTM, please check my minor suggestion.

src/master/filesystem_operations.cc
lgsilva3087: The move was not needed, because the function takes a const reference.

src/master/restore.cc
lgsilva3087: Just a misc comment about what is expected from the unnamed parameter. It is true that is it our of context, but I don't see a problem.

src/master/filesystem_node.cc
lgsilva3087: Very nice catch. I changed it to be:
```cpp
std::optional<RichACL> aclScratch;
```
This way the ACL is never created for the in-memory implementation, but materialized in KV backends were they are needed.

ralcolea (approved):
Excellent work @lgsilva3087!! 👍 💪 🚀 
I really liked the new documentation!! 🔥

PR #764, opened 2026-02-24
perf(cs): make inputBuffer size progressive

dmga44:
The current implementation of the WriteHighLevelOp class always asks
for maxBlocksPerHddWriteJob_ blocks for its next inputBuffer_, i.e the
structure to land the upcoming data blocks to be written to the drive.
This can became a significant waste of memory when the write high level
operation only writes 1 block, but asks for 32, if the
MAX_BLOCKS_PER_HDD_WRITE_JOB is set to that number, for instance.

The change proposed targets asking for smaller initial input buffers,
to limit the memory wasted. Considering the previous case, the
write high level operation will ask for 4, 8, 16, 32 and continue
asking for input buffers of 32 blocks.

The main downside is a possible increase in the number of drive
operations in certain scenarios.


dmga44:
@gemini-code-assist please review again.

uristdwarf (approved):

lgsilva3087 (approved):
LGTM

ralcolea (approved):
Nice change @dmga44! 👍 💪 🚀

PR #770, opened 2026-02-27
feat(admin): Add default TLS for dump-config LS-380

rolysr:
Recent changes enabled TLS related configs on metadata servers, chunkservers and metaloggers, but corresponding dump-config function on saunafs-admin binary was not properly updated for handling defaults for these option values when being used. This commit adds those defaults values to be dumped to corresponding binaries for making TLS config addition more consistent.


rolysr:
@gemini-code-assist review

dmga44 (approved):

lgsilva3087 (approved):

PR #771, opened 2026-02-27
chore(windows): Fix tests CS stop reliability LS-381

rolysr:
Recently, the test test_truncate_retry has been observed to fail on the Windows client CI due to chunkserver stop operations not taking effect immediately. This delay causes the test to become flaky. To mitigate the issue, a short sleep has been added after stopping the chunkserver to ensure that the stop command and its subsequent operations have fully taken effect before the test proceeds.
    
Additionally, the test test_ec_read_combinations was flaky because insufficient time was given for the file-generate utility to write all file chunk parts. This timing issue also caused intermittent failures on the Windows client CI.

Co-authored-by: Dave <dave@leil.io>
Signed-off-by: rolysr <rolysr@leil.io>

### DIFF
--- a/tests/test_suites/ShortSystemTests/test_ec_read_combinations.sh
+++ b/tests/test_suites/ShortSystemTests/test_ec_read_combinations.sh
@@ -12,6 +12,12 @@ mkdir "$dir"
 saunafs setgoal ec "$dir"
 FILE_SIZE=876M file-generate "$dir/file"
 
+# Allow final write/status propagation to settle on Windows
+# before starting chunkserver shutdown test operations.
+if is_windows_system; then
+	sleep 0.1
+fi
+
 for i in {0..7}; do
 	for cs in {0..2}; do
 		saunafs_chunkserver_daemon $(((i + cs) % 8)) stop
--- a/tests/test_suites/ShortSystemTests/test_truncate_retry.sh
+++ b/tests/test_suites/ShortSystemTests/test_truncate_retry.sh
@@ -13,6 +13,7 @@ FILE_SIZE=1234 file-generate file
 checksum123="$(head -c 123 file | md5sum)"
 
 saunafs_chunkserver_daemon 0 stop
+saunafs_wait_for_ready_chunkservers 0
 assert_awk_finds '/no valid copies/' "$(saunafs fileinfo file)"
 # Following operation should pass - wake up just after 4th try (0.2 + 0.4 + 0.8 + 1.6 = 3.0s)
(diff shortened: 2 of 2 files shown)


rolysr:
@gemini-code-assist review

tests/test_suites/ShortSystemTests/test_ec_read_combinations.sh
```diff
@@ -11,6 +11,9 @@ dir="${info[mount0]}/dir"
 mkdir "$dir"
 saunafs setgoal ec "$dir"
 FILE_SIZE=876M file-generate "$dir/file"
+if is_windows_system; then
+	sleep 0.1
+fi
```
lgsilva3087, line 19: Please document why this is needed for Windows.

tests/test_suites/ShortSystemTests/test_truncate_retry.sh
```diff
@@ -13,13 +13,15 @@ FILE_SIZE=1234 file-generate file
 checksum123="$(head -c 123 file | md5sum)"
 
 saunafs_chunkserver_daemon 0 stop
+saunafs_wait_for_ready_chunkservers 0
```
lgsilva3087, line 16: Is this doing something here?
According to the implementation of `saunafs_chunkserver_daemon`, it should return after the chunkserver exits.

dmga44: It is, the `saunafs_chunkserver_daemon` returns after the CS successfully exits as you said, but `saunafs_wait_for_ready_chunkservers` asks the master how many chunkservers it sees, so the disconnection in the master side should have occurred.

lgsilva3087: So there is a delay between the chunkserver exit and the master report?

dmga44: There could be a delay between the chunkserver exit and the master noticing the missing chunkserver, yes. Small one, but there could be.

tests/test_suites/ShortSystemTests/test_ec_read_combinations.sh
dmga44: Primary reason here is that CS stop below these lines (and some other processes) in our test suite in windows is apparently not as gentle as we were expecting, so we give it a bit more time to finish a couple of things like sending some write end statuses before stopping.

dmga44: @rolysr

uristdwarf (approved):

ralcolea (approved):

PR #772, opened 2026-03-02
build: Add RPM package for client

uristdwarf:
* Fix linker not correctly linking isal-l on Rocky Linux 8
* Update tests/install-packages.sh for RPM packages


tests/install-packages.sh
```diff
@@ -66,11 +64,13 @@ common_packages=(
 	bison
 	byacc
 	dbus
-	doxygen
+	# doxygen
```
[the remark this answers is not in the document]
uristdwarf: doxygen is not actually required, so it can be removed entirely

rolysr (approved):
LGTM.

src/data/CMakeLists.txt
lgsilva3087, line 51: Should we also add this target to the client component?

lgsilva3087 (approved):
LGTM

src/data/CMakeLists.txt
uristdwarf: For now, no, because this requires tests be enabled.

PR #789, opened 2026-03-12
perf(mount): reduce reads while writing

dmga44:
The goal of this change is to reduce the reads needed to perform write operations. This can significantly improve the speed of creation of small files when its sizes are not aligned to the SaunaFS block size and some other types of writes, such as random writes but in a less significant manner.

The current implementation restricts the consolidation of several WriteCacheBlock's into a single ChunkWriter::Operation (from now on operation) with the following conditions:
- the chunk must be the same as the one in the operation.
- the stripe must be the same as the one in the operation.
- the interval of the block to be modified must be the same as the ones the operation has.

If a cache block cannot be added to an operation, it must be put into a separate one and possibly wait for the completion of the one it was trying to attach.

The proposed changes remove the last constraint by making the operations to keep track of the interval it is going to modify and changing the logic that computes the parity parts. The result is that write cache blocks from actual client writes inside a single operation do not need to modify the same interval, at the expense of having to write the interval in the parities that covers them and possibly reading more in each of the data parts blocks to read the interval that covers them.

The chunk connector used while reading the missing parts to recalculate
the parity was changed to the ones always used in the reads to comply
with assumptions made in the chunkserver.

### DIFF
--- a/src/mount/chunk_writer.cc
+++ b/src/mount/chunk_writer.cc
@@ -53,18 +53,14 @@ static uint32_t gcd(uint32_t a, uint32_t b) {
 	}
 }
 
-ChunkWriter::Operation::Operation() : unfinishedWrites(0), offsetOfEnd(0) {}
-
 bool ChunkWriter::Operation::isExpandPossible(JournalPosition newPosition, uint32_t stripeSize) {
 	// If the operation is not empty, the new JournalPosition has to be compatible with
 	// the previous elements of the operation, ie. we can only expand by a new block
-	// of the same stripe and the same (from, to) range
-	for (const JournalPosition& position : journalPositions) {
+	// of the same stripe
+	for (const JournalPosition &position : journalPositions) {
 		sassert(newPosition->chunkIndex == position->chunkIndex);
-		if (newPosition->from != position->from
-				|| newPosition->to != position->to
-				|| (newPosition->blockIndex / stripeSize) != (position->blockIndex / stripeSize)
-				|| newPosition->blockIndex == position->blockIndex) {
+		if (newPosition->blockIndex == position->blockIndex ||
+		    (newPosition->blockIndex / stripeSize) != (position->blockIndex / stripeSize)) {
 			return false;
 		}
 	}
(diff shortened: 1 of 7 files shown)


src/mount/chunk_writer.cc
```diff
@@ -414,20 +418,35 @@ void ChunkWriter::fillOperation(Operation &operation, int first_block, int first
...
 	readBlocks(first_block + first_index, size, block_from, block_to, blocks);
 	assert(blocks.size() == (size_t)size);
 
 	for (int index = 0; index < size; ++index) {
-		// Insert the new block into the journal just after the last block of the operation
-		auto position = journal_.insert(operation.journalPositions.back(), std::move(blocks[index]));
-		operation.journalPositions.push_back(position);
+		bool isBlockInOperation = false;
+		JournalPosition position;
+		// Check if the block is already in the journal, because could be the case that the write
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/write_cache_block.cc
```diff
@@ -95,3 +95,24 @@ const uint8_t* WriteCacheBlock::data() const {
...
+uint8_t* WriteCacheBlock::rawData(uint32_t offset) {
+	return blockData.data() + offset;
+}
+
+void WriteCacheBlock::overwriteWithReadBlock(const WriteCacheBlock& readBlock) {
+	sassert(readBlock.type == kReadBlock);
+	sassert(chunkIndex == readBlock.chunkIndex);
+	sassert(blockIndex == readBlock.blockIndex);
+	sassert(from >= readBlock.from);
+	sassert(to <= readBlock.to);
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/write_cache_block.h
```diff
@@ -52,4 +52,6 @@ struct WriteCacheBlock {
 	uint32_t size() const;
 	const uint8_t* data() const;
 	uint8_t* data();
+	uint8_t* rawData(uint32_t offset);
+	void overwriteWithReadBlock(const WriteCacheBlock& readBlock);
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/chunk_writer.cc
```diff
@@ -53,18 +53,20 @@ static uint32_t gcd(uint32_t a, uint32_t b) {
 	}
 }
 
-ChunkWriter::Operation::Operation() : unfinishedWrites(0), offsetOfEnd(0) {}
+ChunkWriter::Operation::Operation()
+    : unfinishedWrites(0),
+      offsetOfEnd(0),
+      minimumModifiedOffset(SFSBLOCKSIZE),
+      maximumModifiedOffset(0) {}
```
lgsilva3087: Consider inline initializations (in the header) if this is the only place.

src/mount/chunk_writer.cc
```diff
@@ -481,32 +499,42 @@ void ChunkWriter::fillNotExisting(Operation &operation, int first_block,
...
-		assert((position->blockIndex % combinedStripeSize_) == (position->blockIndex - first_block));
+		assert((position->blockIndex % combinedStripeSize_) ==
+		       (position->blockIndex - first_block));
 		assert(((int)position->blockIndex - first_block) < combinedStripeSize_);
-		assert(position->to == operation.journalPositions.front()->to);
-		assert(position->from == operation.journalPositions.front()->from);
+		assert(position->to <= operation.maximumModifiedOffset);
+		assert(position->from >= operation.minimumModifiedOffset);
+		if (position->from != operation.minimumModifiedOffset ||
+		    position->to != operation.maximumModifiedOffset) {
```
lgsilva3087, line 502: Maybe out of scope, but consider renaming meber variables `to` and `from`: For example:
```cpp
position->to != ...
position->from != ...
// Could be
position->toOffset != ...
position->fromOffset != ...
```

src/mount/write_cache_block.cc
```diff
@@ -95,3 +95,26 @@ const uint8_t* WriteCacheBlock::data() const {
 uint8_t* WriteCacheBlock::data() {
 	return blockData.data() + from;
 }
+
+uint8_t* WriteCacheBlock::rawData(uint32_t offset) {
+	assert(offset < SFSBLOCKSIZE);
+	return blockData.data() + offset;
+}
+
+void WriteCacheBlock::overwriteWithReadBlock(const WriteCacheBlock& readBlock) {
```
lgsilva3087, line 104: Consider an integration test to exercise this function.

lgsilva3087 (approved):
Great job. Please see my non-blocking suggestions.

rolysr (approved):
LGTM. Very nice job @dmga44!

src/mount/chunk_writer.cc
dmga44: Changed it to:
```cpp
		uint32_t unfinishedWrites{0};                   // number of write request sent
		uint64_t offsetOfEnd{0};                        // offset in the file
		// minimum modified offset in each block of the operation
		uint32_t minimumModifiedOffset{SFSBLOCKSIZE};
		// maximum modified offset in each block of the operation
		uint32_t maximumModifiedOffset{0};

		Operation() = default;
```

src/mount/write_cache_block.cc
dmga44: I added the test `ShortSystemTests.test_randwrite_same_stripe` with that intent.

src/mount/chunk_writer.cc
dmga44: I think it is out of scope, so I did not apply it. The `WriteCacheBlock` struct needs some refactoring but it is still simple enough for understanding purposed IMO.

PR #821, opened 2026-04-05
feat(chunkserver): add plugin reload mechanism 

GigaCronos:
This PR introduces a plugin-wide reload mechanism, allowing disk plugins to
reconfigure themselves at runtime without requiring a chunkserver restart.

### What changed

- Added a new pure virtual method `reload()` to `DiskPlugin`, making reload
  a first-class lifecycle operation alongside initialization and cleanup.
- Extended `PluginManager` with `reloadPlugins()`, which iterates over all
  registered disk plugins and invokes their reload logic, including logging
  for observability.
- Integrated plugin reload into the chunkserver reload flow by calling
  `pluginManager.reloadPlugins()` from `hddReload()`.

### Why

Previously, plugin configuration was only applied during initialization,
requiring a full restart to reflect any changes. This limitation made runtime
tuning difficult, especially for components sensitive to configuration changes.

With this change, plugins can dynamically adapt to updated configuration,
enabling more flexible operation and reducing downtime.

### Impact

- Establishes a consistent mechanism for runtime reconfiguration across all
  disk plugins.
- Enables components such as the ZoneFS garbage collector to react to config
  changes (e.g., frequency, enable/disable) without restart.
- Improves operational agility and simplifies testing and tuning workflows.

Related to LS-566

### DIFF
--- a/src/chunkserver/chunkserver-common/plugin_manager.cc
+++ b/src/chunkserver/chunkserver-common/plugin_manager.cc
@@ -118,6 +118,15 @@ bool PluginManager::checkVersion(IPlugin *plugin) {
 	return false;
 }
 
+void PluginManager::reloadPlugins() {
+	for (auto &[name, plugin] : diskPlugins_) {
+		if (plugin) {
+			safs::log_info("Reloading plugin {}", plugin->toString());
+			plugin->reload();
+		}
+	}
+}
+
 void PluginManager::cleanupPlugins() {
 	for (auto &[name, plugin] : diskPlugins_) {
 		if (plugin) {
--- a/src/chunkserver/chunkserver-common/iplugin.h
+++ b/src/chunkserver/chunkserver-common/iplugin.h
@@ -38,6 +38,9 @@ public:
 	/// needed.
 	virtual bool initialize() = 0;
 
+	/// Called on chunkserver reload to refresh plugin configuration parameters.
+	virtual void reload() = 0;
+
 	/// Returns the plugin name
 	virtual std::string name() = 0;
 	/// Returns the plugin version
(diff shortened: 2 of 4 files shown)


src/chunkserver/chunkserver-common/plugin_manager.cc
```diff
@@ -118,6 +118,15 @@ bool PluginManager::checkVersion(IPlugin *plugin) {
...
 }
 
+void PluginManager::reloadPlugins() {
+	for (auto &[name, plugin] : diskPlugins_) {
+		if (plugin) {
+			safs::log_info("Reloading plugin {}", plugin->toString());
+			plugin->reload();
+		}
+	}
+}
```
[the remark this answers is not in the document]
GigaCronos: I will keep the same format of initialize and cleanup.

src/chunkserver/chunkserver-common/disk_plugin.h
```diff
@@ -55,6 +55,9 @@ class BOOST_SYMBOL_VISIBLE DiskPlugin : public IPlugin {
 	/// debugging purposes.
 	std::string toString() override;
 
+	/// Reload plugin config.
+	virtual void reload() = 0;
```
lgsilva3087: This function should be in IPlugin (iplugin.h), which is the interface for all plugins. Please notice that there are specialized DiskPlugin (like the SMR) and general like ICE. The reload mechanism should be available in all of them, even if for now it will contain default implementation in ICE.

lgsilva3087 (commented):
It is looking good. Please see my comment.

src/chunkserver/chunkserver-common/disk_plugin.h
GigaCronos: Fixed

lgsilva3087 (approved):
LGTM. Nice to have.

rolysr (approved):
LGTM. @GigaCronos Good addition.

PR #835, opened 2026-04-14
fix: Fix build errors for Fedora

ZaRdEr15:
In order for the binaries to correctly build in Fedora:
- Fix build failure due to missing mutex and cstdint include (in GCC 15 some C++ standard library headers no longer include other headers that were being used internally by the library), 
- Make Timer::now() static to fix compiler warnings.

Additionally: 
- Remove unused empty UPGRADE file,
- Remove executable permissions on .cc files (RPM packaging guidelines).


ralcolea (approved):
Nice job @ZaRdEr15! 🔥 💪 🚀 
Consider using `Squash and Merge` option to merge the PR to dev.
In that way you keep the original commits in the PR history, but a single commit will be merged to dev with the detailed list of the changes in the commit message.

rolysr (approved):
LGTM.

PR #849, opened 2026-04-23
refactor(CS): remove defragmentOrMoveChunk from IDisk

GigaCronos:
Removes defragmentOrMoveChunk function from disk interface, now only needed in ZoneFSDisk, also removed crcData parameter.

The function was also removed from cmrDisk implementation.

Related to LS-538


lgsilva3087 (approved):
LGTM

dmga44 (approved):

PR #875, opened 2026-06-01
refactor(master): return QuotaCheckResult in quota checks

lgsilva3087:
The quota check virtuals returned bool, which conflates two distinct outcomes: "a hard quota would be exceeded" and "the backend could not read the quota state". That is fine for the in-memory backend, where reading quota state cannot fail. It is not fine for alternative KV backends: a quota read there can fail, and such a read failure does not reliably fail the later metadata commit. Folding a failed read into "not exceeded" lets a metadata mutation commit without an enforced quota check (fail-open).

Widen the contract to QuotaCheckResult{status, exceeded} so a backend can surface a read failure as a status instead of a silent false. Add statusFromQuotaCheck, which maps {OK, true} to SAUNAFS_ERROR_QUOTA and propagates any non-OK status before exceeded is inspected. Update quotaExceeded, quotaExceededUg, quotaExceededDir, and quotaExceededDirMove plus their call sites (trySetLength, symlink, mknod, mkdir, rename, append, writeChunk, and snapshot cloneNodeTest) to check status before acting on exceeded.

The in-memory base returns SAUNAFS_STATUS_OK unconditionally, so its behavior is unchanged. This commit lands only the public contract and the call-site propagation; the KV-backed implementation that reports read failures as SAUNAFS_ERROR_IO lives in an alternative backend.

Related to: LS-756, LS-759

### DIFF
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -1762,12 +1762,8 @@ uint8_t FilesystemOperationsBase::symlink(const FsContext &context,
 	}
 
 	if (context.isPersonalityMaster()) {
-		uint8_t quotaStatus = statusFromQuotaCheck(quotaExceededUg(
-		    fsOpContext, context.uid(), context.gid(), {{QuotaResource::kInodes, 1}}));
-		if (quotaStatus != SAUNAFS_STATUS_OK) { return quotaStatus; }
-
-		quotaStatus = statusFromQuotaCheck(
-		    quotaExceededDir(fsOpContext, workDir, {{QuotaResource::kInodes, 1}}));
+		uint8_t quotaStatus = checkQuotaUgDir(
+		    fsOpContext, context.uid(), context.gid(), workDir, {{QuotaResource::kInodes, 1}});
 		if (quotaStatus != SAUNAFS_STATUS_OK) { return quotaStatus; }
 	}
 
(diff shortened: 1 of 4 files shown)


src/master/filesystem_operations.cc
```diff
@@ -1759,11 +1761,14 @@ uint8_t FilesystemOperationsBase::symlink(const FsContext &context,
...
-	     quotaExceededDir(fsOpContext, workDir, {{QuotaResource::kInodes, 1}}))) {
-		return SAUNAFS_ERROR_QUOTA;
+	if (context.isPersonalityMaster()) {
+		uint8_t quotaStatus = statusFromQuotaCheck(quotaExceededUg(
+		    fsOpContext, context.uid(), context.gid(), {{QuotaResource::kInodes, 1}}));
+		if (quotaStatus != SAUNAFS_STATUS_OK) { return quotaStatus; }
+
+		quotaStatus = statusFromQuotaCheck(
+		    quotaExceededDir(fsOpContext, workDir, {{QuotaResource::kInodes, 1}}));
+		if (quotaStatus != SAUNAFS_STATUS_OK) { return quotaStatus; }
```
dmga44: Please note that these couple of quota checks (`quotaExceededUg` and `quotaExceededDir`) are present three times in this file. Consider consolidating into a single function to reduce code duplication.

src/master/snapshot_task.cc
```diff
@@ -30,18 +30,22 @@
...
-	     gFSOperations->quotaExceededDir(fsOpContext, dst_parent,
-	                                    {{QuotaResource::kSize, 1}}))) {
-		return SAUNAFS_ERROR_QUOTA;
+	int quotaStatus = statusFromQuotaCheck(gFSOperations->quotaExceededUg(
+	    fsOpContext, src_node->uid, src_node->gid, {{QuotaResource::kInodes, 1}}));
+	if (quotaStatus != SAUNAFS_STATUS_OK) { return quotaStatus; }
+
+	quotaStatus = statusFromQuotaCheck(
+	    gFSOperations->quotaExceededDir(fsOpContext, dst_parent, {{QuotaResource::kInodes, 1}}));
+	if (quotaStatus != SAUNAFS_STATUS_OK) { return quotaStatus; }
```
dmga44: Consider using the previously described helper function here as well.

dmga44 (approved):
Nice.

Please check my small suggestion.

src/master/filesystem_operations.cc
lgsilva3087: Done. Extracted checkQuotaUgDir (declared on IFilesystemOperations beside quotaExceeded, defined once in FilesystemOperationsBase). The three sites here — symlink, mknod, mkdir — now call it directly. It takes uid/gid and the parent directory explicitly (rather than a node, like quotaExceeded) because the node doesn't exist yet on these create paths, and returns the folded status since every caller propagates it the same way. Status-first propagation is unchanged, so the KV read-failure contract still holds.

src/master/snapshot_task.cc
lgsilva3087: Done — both pairs in cloneNodeTest (inodes and size) now call gFSOperations->checkQuotaUgDir(...).

ralcolea (approved):
Good job @lgsilva3087! 👍

PR #880, opened 2026-06-04
perf: improve write batching

dmga44:
This PR introduces a way for the client to suggest the chunkserver
it is writing to wait up to a point, which is for the new write flush
packet.

The new option in the client sfsusewriteflushpacket is updateable via
the tweaks hidden file and the option sfswritewindowsize was also made
updateable via tweaks hidden file. The triggers to send a write flush
packet are the following:
- write window size packets were sent to the peer chunkserver.
- write end packet is going to be sent to the peer chunkserver.
- new operations collide with some of the previously sent operations.
- the last write to the chunkservers of that ChunkWriter was at least
100ms ago.
- the chunk writer is in flush mode, i.e either a flush was called on
the inode, the chunk is not the last one written in the inode or a lot
of time passed, and there are no new operations.

Please note these changes make the whole behavior of the chunk writer
to depend on the instantiation parameters passed (writeWindowSize and
useWriteFlushPacket), this way encapsulating the logic and allowing
those parameters to be updated via tweaks hidden file. Beside the
previous conditions the implementation of the write flush packet
only sends it if at least one previous block of data was sent and the
peer chunkserver supports it. The new packets are tested with new unittests.

In the chunkserver side the idea revolves around having a
WriteHighLevelOp that depending on the new parameter expectWriteFlush
heavily changes its behavior. The write flush packet also works for the
replication goals (state WriteForward).

This is the comparison of the main behavior changes that depends on
expectWriteFlush parameter:
- expectWriteFlush=0: Prior behavior. Every job_write flushes a single
InputBuffer. The conditions to start a write are that no job is in
progress and there are InputBuffers already freezed (in
writeDataBuffers_) ready to be flushed. The condition to freeze the
current inputBuffer_ is that it is full or there is no job in progress.
- expectWriteFlush=1: New behavior. Every job_write flushes some
InputBuffers that depends on the number of freezed ones between each
write flush packet received (inputBuffersForNextFlush_). The conditions
to start a write are that this queue (inputBuffersForNextFlush_) is not
empty and no write job is in progress. The condition to freeze the
current inputBuffer_ is that it is full or a write flush packet has
arrived.

A new WriteHighLevelOpExpectFlush was created to extract the logic
coming from the new behavior implemented and making the whole
code more readable.

As a side change, the effective write window size was increased by 1.
The unfinished-operations cap counted the trailing newOperations_
entry, which is usually held back to complete a stripe. That left
in-flight pending operations bounded by writeWindowSize_ - 1 in many
cases. Adjust the check so at least writeWindowSize_ operations are
kept in flight in a broader set of cases, without changing the meaning
of the write window size option.

Some of the changes also target reducing the amount of locking/unlocking of the
main mutexes synchronizing the write algorithm in the client side.

Changes include:
- making lastWriteToDataChain and lastWriteToChunkservers atomic the
way locking to protect them is not necessary.
- maintaining an atomic inDelayedQueue member of the ChunkData
structure so the write_block function does not perform extra
lock()/unlock() calls on the inode and global locks for unnecessary
cases.
- consolidate the releasing of blocks while moving them to the writer
journal so even when several blocks are moved, the inode lock is open
and closed once.

Side changes:
- log the chunkId in the case some write error happens. Change the log
level of the SAUNAFS_ERROR_LOCKED errors to log_info.
- maintain lastChunkDataWritten and use it as another condition to
trigger flush mode in the writer.

The PR also introduces changes in the testing framework that randomly enables/disables the sfsusewriteflushpacket option for every mountpoint independently. Also the tweaking ability of the WriteWindowSize and UseWriteFlushPacket options were tested.

The far_eio_trigger case in the chunk_operations_eio library was
extended the way it considers now the end offset of the request instead
of the begining. The reason is that mounts using write flush packet
usually compose a single write operation at offset 0 if not much data
is sent, and it was not triggering the errors as expected in the tests.

Includes the bump of the main CMake version to v5.11.0.

### DIFF
--- a/doc/sfsmount.1.adoc
+++ b/doc/sfsmount.1.adoc
@@ -193,7 +193,8 @@ Try to change nice level to specified value on startup (default: -19).
 Define number of write workers (default: 10).
 
 *-o sfswritewindowsize=*'N'::
-Define write window size (in blocks) for each chunk (default: 15).
+Define write window size (in blocks) for each chunk. Can be updated via
+tweaks hidden file (default: 15).
 
 *-o sfsmemlock*::
 Try to lock memory (must be enabled at build time).
@@ -227,6 +228,12 @@ Equivalent to *--nonempty* option for use in fstab.
 *-o sfsmaxchunkswritteninparallelperinode=*'N'::
 Define the maximum number of chunks that can be written in parallel per inode.
 0 stands for unlimited (default: 0).
+
+*-o sfsusewriteflushpacket=*'0|1'::
+When set to 1, suggests the chunkserver to flush the packets received every
+window size blocks  (option sfswritewindowsize) or when a write needs to finish
+(e.g. when flush() is called on the inode or when enough time has passed). Can
+be updated via tweaks hidden file (default: 0).
 
 *-o sfsdirectio=*'0|1'::
 Whether to use FUSE DirectIO. This may improve performance when reading large
--- a/src/chunkserver/chunk_high_level_ops.cc
+++ b/src/chunkserver/chunk_high_level_ops.cc
@@ -338,6 +338,10 @@ bool WriteHighLevelOp::isWriteJobBeingProcessed() const { return writeJobId_ != 
 
 void WriteHighLevelOp::setNoWriteJobBeingProcessed() { writeJobId_ = 0; }
 
+bool WriteHighLevelOp::isInputBufferFreezable() const {
+	return inputBuffer_ != nullptr && !inputBuffer_->isBeingUpdated();
+}
+
 void WriteHighLevelOp::startOpenWriteJob() {
 	writeJobWriteId_ = 0;
 	writeJobId_ = job_open(
@@ -389,55 +393,135 @@ void WriteHighLevelOp::delayedCloseCallback(uint8_t status, void * /*entry*/) {
 	decreasePendingWriteJobsToCheckAndApplyClosedOnParent();
 }
 
+// ---------------------------------------------------------------------------
+// Differentiating hooks. Each `WriteHighLevelOp` base implementation is
+// immediately followed by the matching `WriteHighLevelOpExpectFlush` override
+// for easy side-by-side comparison.
+// ---------------------------------------------------------------------------
+
+bool WriteHighLevelOp::canStartWriteJobNow() const {
+	return !writeDataBuffers_.empty() && !isWriteJobBeingProcessed();
+}
+
+bool WriteHighLevelOpExpectFlush::canStartWriteJobNow() const {
+	if (!expectsWriteFlush_) { return WriteHighLevelOp::canStartWriteJobNow(); }
+	return !writeDataBuffers_.empty() && !isWriteJobBeingProcessed() &&
+	       !inputBuffersForNextFlush_.empty();
+}
+
+bool WriteHighLevelOp::shouldWriteCurrentInputBuffer() const {
+	return isInputBufferFreezable() && (inputBuffer_->isFull() || !isWriteJobBeingProcessed());
+}
+
+bool WriteHighLevelOpExpectFlush::shouldWriteCurrentInputBuffer() const {
+	if (!expectsWriteFlush_) { return WriteHighLevelOp::shouldWriteCurrentInputBuffer(); }
+	return isInputBufferFreezable() && inputBuffer_->isFull();
+}
+
+void WriteHighLevelOp::freezeCurrentInputBuffer() {
+	writeDataBuffers_.emplace_back(std::move(inputBuffer_));
+}
+
+void WriteHighLevelOpExpectFlush::freezeCurrentInputBuffer() {
+	WriteHighLevelOp::freezeCurrentInputBuffer();
+	if (expectsWriteFlush_) { ++inputBuffersReadySinceLastFlush_; }
+}
+
+void WriteHighLevelOp::collectInputBuffersForNextJob(std::vector<InputBuffer *> &inputBuffers) {
+	if (writeDataBuffers_.front()->currentBlocks() == 0) {
+		safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+	}
+
+	enqueuedInputBuffers_ = 1;
+	inputBuffers.push_back(writeDataBuffers_.front().get());
+}
+
+void WriteHighLevelOpExpectFlush::collectInputBuffersForNextJob(
+    std::vector<InputBuffer *> &inputBuffers) {
+	if (!expectsWriteFlush_) {
+		WriteHighLevelOp::collectInputBuffersForNextJob(inputBuffers);
+		return;
+	}
+
+	enqueuedInputBuffers_ = inputBuffersForNextFlush_.front();
+	inputBuffersForNextFlush_.pop();
+
+	if (writeDataBuffers_.size() < enqueuedInputBuffers_) {
+		safs::log_err("({}) Fewer available write data buffers than expected ({} < {}).", __func__,
+		              writeDataBuffers_.size(), enqueuedInputBuffers_);
+		enqueuedInputBuffers_ = writeDataBuffers_.size();
+	}
+
+	auto inputBuffersIt = writeDataBuffers_.begin();
+	for (uint32_t i = 0; i < enqueuedInputBuffers_; i++) {
+		if (inputBuffersIt->get()->currentBlocks() == 0) {
+			safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+		}
+
+		inputBuffers.push_back(inputBuffersIt->get());
+		inputBuffersIt++;
+	}
+}
+
+bool WriteHighLevelOp::shouldDeferInstantReply(bool /*fromFlushDataCall*/) const { return false; }
+
+bool WriteHighLevelOpExpectFlush::shouldDeferInstantReply(bool fromFlushDataCall) const {
+	return expectsWriteFlush_ && !fromFlushDataCall;
+}
+
+void WriteHighLevelOp::onSwitchingToEagerWriteOnDelayedClose() {}
+
+void WriteHighLevelOpExpectFlush::onSwitchingToEagerWriteOnDelayedClose() {
+	expectsWriteFlush_ = false;
+}
+
+void WriteHighLevelOp::resetFlushBatchingState() {}
+
+void WriteHighLevelOpExpectFlush::resetFlushBatchingState() {
+	inputBuffersReadySinceLastFlush_ = 0;
+	inputBuffersForNextFlush_ = std::queue<uint32_t>();
+}
+
+void WriteHighLevelOp::flushData() {
+	safs::log_warn("({}) Called flushData on a write operation that does not expect write flush.",
+	               __func__);
+}
+
+void WriteHighLevelOpExpectFlush::flushData() {
+	tryInstantReply(true);
+
+	if (inputBuffer_ != nullptr && inputBuffer_->currentBlocks() > 0 &&
+	    !inputBuffer_->isBeingUpdated()) {
+		freezeCurrentInputBuffer();
+	}
+
+	if (inputBuffersReadySinceLastFlush_ > 0) {
+		inputBuffersForNextFlush_.push(inputBuffersReadySinceLastFlush_);
+		inputBuffersReadySinceLastFlush_ = 0;
+
+		if (canStartWriteJobNow()) { startNextWriteJob(); }
+	} else {
+		// This can happen if the connection is created but very little data is sent, which implies
+		// some WriteHighLevelOp receive no write data packets.
+		safs::log_debug("({}) Called flushData with no input buffers ready to be flushed.",
+		                __func__);
+	}
+}
+
+// ---------------------------------------------------------------------------
+// End of differentiating hooks.
+// ---------------------------------------------------------------------------
+
 void WriteHighLevelOp::startNextWriteJob() {
-	if (writeDataBuffers_.empty()) {
-		safs::log_warn("({}) Called with no write data buffers.", __func__);
-		return;
-	}
-
-	if (isWriteJobBeingProcessed()) {
-		safs::log_warn("({}) Called with write job already in progress.", __func__);
-		return;
-	}
-
-	if (expectWriteFlush_ && inputBuffersForNextFlush_.empty()) {
-		safs::log_warn(
-		    "({}) Called with no input buffers ready to be flushed while expecting write flush.",
-		    __func__);
+	if (!canStartWriteJobNow()) {
+		safs::log_warn("({}) Called when not ready to start a write job.", __func__);
 		return;
 	}
 
 	/// Start the next write job: it is always the first write data buffer
 	writeJobWriteId_ = writeDataBuffers_.front()->getLastWriteId();
 	std::vector<InputBuffer *> inputBuffers;
-
-	if (expectWriteFlush_) {
-		enqueuedInputBuffers_ = inputBuffersForNextFlush_.front();
-		inputBuffersForNextFlush_.pop();
-
-		if (writeDataBuffers_.size() < enqueuedInputBuffers_) {
-			safs::log_err("({}) Fewer available write data buffers than expected ({} < {}).",
-			              __func__, writeDataBuffers_.size(), enqueuedInputBuffers_);
-			enqueuedInputBuffers_ = writeDataBuffers_.size();
-		}
-
-		auto inputBuffersIt = writeDataBuffers_.begin();
-		for (uint32_t i = 0; i < enqueuedInputBuffers_; i++) {
-			if (inputBuffersIt->get()->currentBlocks() == 0) {
-				safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
-			}
-
-			inputBuffers.push_back(inputBuffersIt->get());
-			inputBuffersIt++;
-		}
-	} else {
-		if (writeDataBuffers_.front()->currentBlocks() == 0) {
-			safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
-		}
-
-		enqueuedInputBuffers_ = 1;
-		inputBuffers.push_back(writeDataBuffers_.front().get());
-	}
+	collectInputBuffersForNextJob(inputBuffers);
 
 	writeJobId_ = job_write(
 	    *workerJobPool(),
@@ -445,54 +529,23 @@ void WriteHighLevelOp::startNextWriteJob() {
 	    chunkId_, chunkVersion_, chunkType_, inputBuffers);
 }
 
-void WriteHighLevelOp::writeCurrentInputPacket() {
-	writeDataBuffers_.emplace_back(std::move(inputBuffer_));
-	inputBuffersReadySinceLastFlush_++;
-	if (!isWriteJobBeingProcessed() && !expectWriteFlush_) { startNextWriteJob(); }
+void WriteHighLevelOp::writeCurrentInputBuffer() {
+	freezeCurrentInputBuffer();
+	if (canStartWriteJobNow()) { startNextWriteJob(); }
 }
 
 void WriteHighLevelOp::continueWritingIfPossible() {
 	tryInstantReply();
 	// There should not be any write jobs being processed.
 
-	if (!writeDataBuffers_.empty() && (!expectWriteFlush_ || !inputBuffersForNextFlush_.empty())) {
-		// There is a write buffer ready to be written and either we do not expect write flushes (so
-		// we can write it immediately) or we expect write flushes but there are already scheduled
-		// flushes from previous flushData calls, so we can write it immediately.
+	if (canStartWriteJobNow()) {
 		startNextWriteJob();
 		return;
 	}
 
-	if (!expectWriteFlush_ && inputBuffer_ != nullptr && !inputBuffer_->isBeingUpdated()) {
+	if (shouldWriteCurrentInputBuffer()) {
 		assert(inputBuffer_->currentBlocks() > 0);
-		writeCurrentInputPacket();
-	}
-}
-
-void WriteHighLevelOp::flushData() {
-	if (!expectWriteFlush_) {
-		safs::log_warn("({}) Called flushData on a write operation that does not expect write flush.",
-		               __func__);
-	}
-
-	tryInstantReply(true);
-
-	if (inputBuffer_ != nullptr && inputBuffer_->currentBlocks() > 0 && !inputBuffer_->isBeingUpdated()) {
-		writeDataBuffers_.emplace_back(std::move(inputBuffer_));
-		inputBuffersReadySinceLastFlush_++;
-	}
-
-	if (inputBuffersReadySinceLastFlush_ > 0) {
-		inputBuffersForNextFlush_.push(inputBuffersReadySinceLastFlush_);
-		inputBuffersReadySinceLastFlush_ = 0;
-	
-		if (!isWriteJobBeingProcessed() && !writeDataBuffers_.empty()) {
-			startNextWriteJob();
-		}
-	} else {
-		// This can happen if the connection is created but very little data is sent, which implies
-		// some WriteHighLevelOp receive no write data packets.
-		safs::log_debug("({}) Called flushData with no input buffers ready to be flushed.", __func__);
+		writeCurrentInputBuffer();
 	}
 }
 
@@ -571,7 +624,7 @@ void WriteHighLevelOp::prepareInputBufferForWrite(bool isForward) {
 	if (inputBuffer_ != nullptr && inputBuffer_->isFull()) {
 		safs::log_warn("({}) Called with full inputBuffer, isForward: {}. Writing existing buffer.",
 		               __func__, isForward);
-		writeCurrentInputPacket();
+		writeCurrentInputBuffer();
 	}
 
 	if (inputBuffer_ == nullptr) {
@@ -586,14 +639,12 @@ void WriteHighLevelOp::prepareInputBufferForWrite(bool isForward) {
 	inputBuffer_->addNewWriteOperation();
 }
 
-void WriteHighLevelOp::setup(uint64_t chunkId, uint32_t chunkVersion, ChunkPartType chunkType,
-                             bool expectWriteFlush) {
+void WriteHighLevelOp::setup(uint64_t chunkId, uint32_t chunkVersion, ChunkPartType chunkType) {
 	stats_hlopw++;
 
 	chunkId_ = chunkId;
 	chunkVersion_ = chunkVersion;
 	chunkType_ = chunkType;
-	expectWriteFlush_ = expectWriteFlush;
 
 	isChunkLocked_ = masterconn_get_job_pool()->enforceChunkLock(chunkId_, chunkType_);
 	startOpenWriteJob();
@@ -613,10 +664,9 @@ void WriteHighLevelOp::processWriteDataBlock(uint16_t blocknum, uint32_t opOffse
 	inputBuffer_->setupLastWriteOperation(blocknum, opOffset, opSize, writeId, crc);
 	tryInstantReply();
 
-	// No write jobs in progress or current input buffer is full - write it
-	if (inputBuffer_->isFull() || (!isWriteJobBeingProcessed() && !expectWriteFlush_)) {
+	if (shouldWriteCurrentInputBuffer()) {
 		assert(inputBuffer_->currentBlocks() > 0);
-		writeCurrentInputPacket();
+		writeCurrentInputBuffer();
 	}
 }
 
@@ -660,7 +710,7 @@ bool WriteHighLevelOp::trySeal() {
 		// There is an input buffer, all its data has been processed and replied
 		// so we can write it out to complete the operation
 		assert(inputBuffer_->currentBlocks() > 0);
-		writeCurrentInputPacket();
+		writeCurrentInputBuffer();
 	} else {
 		// There is no input buffer, we can only seal if all write data buffers have been replied
 		for (const auto &buffer : writeDataBuffers_) {
@@ -724,7 +774,7 @@ void WriteHighLevelOp::delayedClose() {
 
 	// We do not expect write flush anymore, so all the remaining buffers should be written out, but
 	// one-by-one
-	expectWriteFlush_ = false;
+	onSwitchingToEagerWriteOnDelayedClose();
 	if (!isWriteJobBeingProcessed() && !writeDataBuffers_.empty()) {
 		// No write job in progress, start one to trigger the processing of the remaining buffers,
 		// which will be replied in writeFinishedCallback
@@ -752,8 +802,7 @@ void WriteHighLevelOp::cleanup() {
 		writeDataBuffers_.pop_front();
 	}
 	enqueuedInputBuffers_ = 0;
-	inputBuffersReadySinceLastFlush_ = 0;
-	inputBuffersForNextFlush_ = std::queue<uint32_t>();
+	resetFlushBatchingState();
 	nextInputBufferBlockCount_ =
 	    std::min(kDefaultInitialNextInputBufferBlockCount, maxBlocksPerHddWriteJob_);
 
@@ -794,7 +843,7 @@ void WriteHighLevelOp::tryInstantReply(bool fromFlushDataCall) {
 		return;
 	}
 
-	if (expectWriteFlush_ && !fromFlushDataCall) {
+	if (shouldDeferInstantReply(fromFlushDataCall)) {
 		// If we expect write flushes, we only try instant reply when flushData is called, because
 		// that's the only time when we know that there won't be more data to be flushed for the
 		// current write operation.
--- a/src/chunkserver/chunk_high_level_ops.cc
+++ b/src/chunkserver/chunk_high_level_ops.cc
@@ -394,20 +394,51 @@ void WriteHighLevelOp::startNextWriteJob() {
 		safs::log_warn("({}) Called with no write data buffers.", __func__);
 		return;
 	}
-	if (writeDataBuffers_.front()->currentBlocks() == 0) {
-		safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
-		return;
-	}
 
 	if (isWriteJobBeingProcessed()) {
 		safs::log_warn("({}) Called with write job already in progress.", __func__);
 		return;
 	}
 
+	if (expectWriteFlush_ && inputBuffersForNextFlush_.empty()) {
+		safs::log_warn(
+		    "({}) Called with no input buffers ready to be flushed while expecting write flush.",
+		    __func__);
+		return;
+	}
+
 	/// Start the next write job: it is always the first write data buffer
 	writeJobWriteId_ = writeDataBuffers_.front()->getLastWriteId();
-	enqueuedInputBuffers_ = 1;
-	std::vector<InputBuffer *> inputBuffers = {writeDataBuffers_.front().get()};
+	std::vector<InputBuffer *> inputBuffers;
+
+	if (expectWriteFlush_) {
+		enqueuedInputBuffers_ = inputBuffersForNextFlush_.front();
+		inputBuffersForNextFlush_.pop();
+
+		if (writeDataBuffers_.size() < enqueuedInputBuffers_) {
+			safs::log_err("({}) Fewer available write data buffers than expected ({} < {}).",
+			              __func__, writeDataBuffers_.size(), enqueuedInputBuffers_);
+			enqueuedInputBuffers_ = writeDataBuffers_.size();
+		}
+
+		auto inputBuffersIt = writeDataBuffers_.begin();
+		for (uint32_t i = 0; i < enqueuedInputBuffers_; i++) {
+			if (inputBuffersIt->get()->currentBlocks() == 0) {
+				safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+			}
+
+			inputBuffers.push_back(inputBuffersIt->get());
+			inputBuffersIt++;
+		}
+	} else {
+		if (writeDataBuffers_.front()->currentBlocks() == 0) {
+			safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+		}
+
+		enqueuedInputBuffers_ = 1;
+		inputBuffers.push_back(writeDataBuffers_.front().get());
+	}
+
 	writeJobId_ = job_write(
 	    *workerJobPool(),
 	    [this](uint8_t status, void *entry) { this->writeFinishedCallback(status, entry); },
@@ -416,22 +447,52 @@ void WriteHighLevelOp::startNextWriteJob() {
 
 void WriteHighLevelOp::writeCurrentInputPacket() {
 	writeDataBuffers_.emplace_back(std::move(inputBuffer_));
-	if (!isWriteJobBeingProcessed()) { startNextWriteJob(); }
+	inputBuffersReadySinceLastFlush_++;
+	if (!isWriteJobBeingProcessed() && !expectWriteFlush_) { startNextWriteJob(); }
 }
 
 void WriteHighLevelOp::continueWritingIfPossible() {
 	tryInstantReply();
+	// There should not be any write jobs being processed.
 
-	if (!writeDataBuffers_.empty()) {
-		// there is a write buffer ready to be written, there should not be any
-		// write jobs being processed.
+	if (!writeDataBuffers_.empty() && (!expectWriteFlush_ || !inputBuffersForNextFlush_.empty())) {
+		// There is a write buffer ready to be written and either we do not expect write flushes (so
+		// we can write it immediately) or we expect write flushes but there are already scheduled
+		// flushes from previous flushData calls, so we can write it immediately.
 		startNextWriteJob();
 		return;
 	}
 
-	if (inputBuffer_ != nullptr && !inputBuffer_->isBeingUpdated()) {
+	if (!expectWriteFlush_ && inputBuffer_ != nullptr && !inputBuffer_->isBeingUpdated()) {
 		assert(inputBuffer_->currentBlocks() > 0);
 		writeCurrentInputPacket();
+	}
+}
+
+void WriteHighLevelOp::flushData() {
+	if (!expectWriteFlush_) {
+		safs::log_warn("({}) Called flushData on a write operation that does not expect write flush.",
+		               __func__);
+	}
+
+	tryInstantReply(true);
+
+	if (inputBuffer_ != nullptr && inputBuffer_->currentBlocks() > 0 && !inputBuffer_->isBeingUpdated()) {
+		writeDataBuffers_.emplace_back(std::move(inputBuffer_));
+		inputBuffersReadySinceLastFlush_++;
+	}
+
+	if (inputBuffersReadySinceLastFlush_ > 0) {
+		inputBuffersForNextFlush_.push(inputBuffersReadySinceLastFlush_);
+		inputBuffersReadySinceLastFlush_ = 0;
+	
+		if (!isWriteJobBeingProcessed() && !writeDataBuffers_.empty()) {
+			startNextWriteJob();
+		}
+	} else {
+		// This can happen if the connection is created but very little data is sent, which implies
+		// some WriteHighLevelOp receive no write data packets.
+		safs::log_debug("({}) Called flushData with no input buffers ready to be flushed.", __func__);
 	}
 }
 
@@ -525,12 +586,14 @@ void WriteHighLevelOp::prepareInputBufferForWrite(bool isForward) {
 	inputBuffer_->addNewWriteOperation();
 }
 
-void WriteHighLevelOp::setup(uint64_t chunkId, uint32_t chunkVersion, ChunkPartType chunkType) {
+void WriteHighLevelOp::setup(uint64_t chunkId, uint32_t chunkVersion, ChunkPartType chunkType,
+                             bool expectWriteFlush) {
 	stats_hlopw++;
 
 	chunkId_ = chunkId;
 	chunkVersion_ = chunkVersion;
 	chunkType_ = chunkType;
+	expectWriteFlush_ = expectWriteFlush;
 
 	isChunkLocked_ = masterconn_get_job_pool()->enforceChunkLock(chunkId_, chunkType_);
 	startOpenWriteJob();
@@ -551,7 +614,7 @@ void WriteHighLevelOp::processWriteDataBlock(uint16_t blocknum, uint32_t opOffse
 	tryInstantReply();
 
 	// No write jobs in progress or current input buffer is full - write it
-	if (!isWriteJobBeingProcessed() || inputBuffer_->isFull()) {
+	if (inputBuffer_->isFull() || (!isWriteJobBeingProcessed() && !expectWriteFlush_)) {
 		assert(inputBuffer_->currentBlocks() > 0);
 		writeCurrentInputPacket();
 	}
@@ -659,6 +722,9 @@ void WriteHighLevelOp::delayedClose() {
 		writeDataBuffers_.pop_back();
 	}
 
+	// We do not expect write flush anymore, so all the remaining buffers should be written out, but
+	// one-by-one
+	expectWriteFlush_ = false;
 	if (!isWriteJobBeingProcessed() && !writeDataBuffers_.empty()) {
 		// No write job in progress, start one to trigger the processing of the remaining buffers,
 		// which will be replied in writeFinishedCallback
@@ -667,7 +733,9 @@ void WriteHighLevelOp::delayedClose() {
 
 	// Pending write jobs will be handled in writeFinishedCallback, they need to write out the
 	// buffers
-	parentPendingWriteJobs() += writeDataBuffers_.size();
+	if (!writeDataBuffers_.empty()) {
+		parentPendingWriteJobs() += writeDataBuffers_.size() - enqueuedInputBuffers_ + 1;
+	}
 }
 
 void WriteHighLevelOp::cleanup() {
@@ -684,6 +752,10 @@ void WriteHighLevelOp::cleanup() {
 		writeDataBuffers_.pop_front();
 	}
 	enqueuedInputBuffers_ = 0;
+	inputBuffersReadySinceLastFlush_ = 0;
+	inputBuffersForNextFlush_ = std::queue<uint32_t>();
+	nextInputBufferBlockCount_ =
+	    std::min(kDefaultInitialNextInputBufferBlockCount, maxBlocksPerHddWriteJob_);
 
 	if (inputBuffer_ != nullptr) {
 		/// Drop the input buffer, it won't be used anymore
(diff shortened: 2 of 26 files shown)


src/chunkserver/chunkserver_entry.cc
```diff
@@ -446,6 +457,35 @@ void ChunkserverEntry::writeStatus(const uint8_t *data, PacketHeader::Type type,
...
 }
 
+void ChunkserverEntry::writeFlush(const uint8_t *data, uint32_t length) {
+	TRACETHIS();
+	uint64_t opChunkId;
+
+	try {
+		cltocs::writeFlush::deserialize(data, length, opChunkId);
+	} catch (IncorrectDeserializationException&) {
+		safs::log_info("Received malformed WRITE_END message (length: {})", length);
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -394,20 +394,51 @@ void WriteHighLevelOp::startNextWriteJob() {
...
+			safs::log_err("({}) Fewer available write data buffers than expected ({} < {}).",
+			              __func__, writeDataBuffers_.size(), enqueuedInputBuffers_);
+			enqueuedInputBuffers_ = writeDataBuffers_.size();
+		}
+
+		auto inputBuffersIt = writeDataBuffers_.begin();
+		for (uint32_t i = 0; i < enqueuedInputBuffers_; i++) {
+			if (writeDataBuffers_.front()->currentBlocks() == 0) {
+				safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+			}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -334,11 +334,17 @@ void ChunkserverEntry::writeInit(const uint8_t *data, PacketHeader::Type type,
...
 		deserializePacketVersionNoHeader(data, length, v);
-		sassert(v == cltocs::writeInit::kECChunks);
-		cltocs::writeInit::deserialize(data, length, chunkId, chunkVersion, chunkType, chain);
+		sassert(v == cltocs::writeInit::kECChunks || v == cltocs::writeInit::kECChunksWithWriteFlush);
+		if (v == cltocs::writeInit::kECChunks) {
+			cltocs::writeInit::deserialize(data, length, chunkId, chunkVersion, chunkType, chain);
+		} else {
+			cltocs::writeInit::deserialize(data, length, chunkId, chunkVersion, chunkType, chain,
+			                               expectWriteFlush);
+		}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/chunk_writer.cc
```diff
@@ -266,6 +284,10 @@ uint32_t ChunkWriter::getUnfinishedOperationsCount() {
 	return pendingOperations_.size() + newOperations_.size();
 }
 
+bool ChunkWriter::isWaitingForData() {
+	return getUnfinishedOperationsCount() < writeWindowSize_ + 1;
+}
```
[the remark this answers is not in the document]
dmga44: The problem here is that the last operation from the new operations container can be waiting to be finished (to complete the stripe), so the actual pending operations in progress can be capped in many cases by writeWindowSize_ - 1. This change targets always having at least writeWindowSize_ pending operations for a larger set of cases, without losing much of the previous meaning of the write window size option. I'm adding it to the commit description.

src/mount/chunk_writer.h
```diff
@@ -92,6 +98,12 @@ class ChunkWriter {
 	 */
 	uint32_t getUnfinishedOperationsCount();
 
+	/*!
+	 * \return true if the ChunkWriter is waiting for more data to write, i.e. if the number of
+	 * pending operations is less than writeWindowSize + 1.
+	 */
+	bool isWaitingForData();
```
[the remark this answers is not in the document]
dmga44: I'm changing it to:
```cpp
 	 * unfinished operations is less than writeWindowSize + 1.
```

src/chunkserver/chunkserver_entry.cc
```diff
@@ -446,6 +458,35 @@ void ChunkserverEntry::writeStatus(const uint8_t *data, PacketHeader::Type type,
...
 }
 
+void ChunkserverEntry::writeFlush(const uint8_t *data, uint32_t length) {
+	TRACETHIS();
+	uint64_t opChunkId;
+
+	try {
+		cltocs::writeFlush::deserialize(data, length, opChunkId);
+	} catch (IncorrectDeserializationException&) {
+		safs::log_info("Received malformed WRITE_FLUSH message (length: {})", length);
```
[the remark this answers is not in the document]
dmga44: Previously fixed.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -394,20 +394,51 @@ void WriteHighLevelOp::startNextWriteJob() {
...
+
+		auto inputBuffersIt = writeDataBuffers_.begin();
+		for (uint32_t i = 0; i < enqueuedInputBuffers_; i++) {
+			if ((*inputBuffersIt)->currentBlocks() == 0) {
+				safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+			}
+
+			inputBuffers.push_back(inputBuffersIt->get());
+			inputBuffersIt++;
+		}
```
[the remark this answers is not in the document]
dmga44: Applied.

doc/sfsmount.1.adoc
```diff
@@ -228,6 +228,11 @@ Equivalent to *--nonempty* option for use in fstab.
 Define the maximum number of chunks that can be written in parallel per inode.
 0 stands for unlimited (default: 0).
 
+*-o sfsusewriteflushpacket=*'0|1'::
+When set to 1, suggests the chunkserver to flush the packets received every
+window size blocks or when write needs to finish, i.e a flush is called on the
+inode or the time  (default: 0).
```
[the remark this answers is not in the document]
dmga44: Applied.

tests/tools/saunafs.sh
```diff
@@ -749,6 +750,14 @@ create_sfsmount_cfg_() {
...
 	echo "sfsmaster=$ip_address"
 	echo "sfsport=${saunafs_info_[matocl]}"
+
+	# Make the tests more robust by randomizing write flush packet usage if not set by the test
+	# Note each mount can have its own setting, so we check for the mount-specific variable.
+	if [[ ${saunafs_info_[use_write_flush_packet]} == NOT_SET ]]; then
+		echo "sfsusewriteflushpacket=$((RANDOM % 2))"
+	else
+		echo "sfsusewriteflushpacket=${saunafs_info_[use_write_flush_packet]}"
+	fi
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -173,8 +173,9 @@ struct inodedata {
 	std::atomic_bool emptyChunkDataList = true;
 	std::atomic<uint64_t> totalCachedBlocks = 0;
 	std::mutex mutex;
-	Timer lastWriteToDataChain;     // inodeLock
-	Timer lastWriteToChunkservers;  // inodeLock
+	ChunkData *lastChunkDataWritten = nullptr;   // inodeLock
+	std::atomic<Timer> lastWriteToDataChain;
+	std::atomic<Timer> lastWriteToChunkservers;
 
 	inodedata(inode_t inode) : inode(inode) {}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -900,9 +922,7 @@ void ChunkJobWriter::processDataChain(ChunkWriter &writer) {
 		    std::min(parent->maxfleng - chunkIndex_ * SFSCHUNKSIZE, (uint64_t)SFSCHUNKSIZE));
 		inodeLock.unlock();
 		if (writer.startNewOperations(can_expect_next_block) > 0) {
-			inodeLock.lock();
-			parent->lastWriteToChunkservers.reset();
-			inodeLock.unlock();
+			parent->lastWriteToChunkservers = Timer();
 		}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -1039,7 +1062,7 @@ void write_data_term(void) {
 int write_block(ChunkData *chunkData, uint16_t pos, uint32_t from, uint32_t to, const uint8_t *data,
                 UniqueLock &inodeLock) {
 	inodedata *parent = chunkData->getParent();
-	parent->lastWriteToDataChain.reset();
+	parent->lastWriteToDataChain = Timer();
```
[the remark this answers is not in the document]
dmga44: Applied.

dmga44:
@gemini-code-assist Please review again.

src/mount/writedata.cc
```diff
@@ -1059,14 +1092,16 @@ int write_block(ChunkData *chunkData, uint16_t pos, uint32_t from, uint32_t to,
 	inodeLock.lock();
 	chunkData->pushToChain(
 	    WriteCacheBlock(chunkData->chunkIndex, pos, WriteCacheBlock::kWritableBlock));
+	parent->lastChunkDataWritten = chunkData;
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -416,25 +447,49 @@ void WriteHighLevelOp::startNextWriteJob() {
...
+		writeDataBuffers_.emplace_back(std::move(inputBuffer_));
+		inputBuffersReadySinceLastFlush_++;
+	}
+
+	inputBuffersForNextFlush_.push(inputBuffersReadySinceLastFlush_);
+	inputBuffersReadySinceLastFlush_ = 0;
+
+	if (!isWriteJobBeingProcessed() && !writeDataBuffers_.empty()) {
+		startNextWriteJob();
+	}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -1039,7 +1072,7 @@ void write_data_term(void) {
 int write_block(ChunkData *chunkData, uint16_t pos, uint32_t from, uint32_t to, const uint8_t *data,
                 UniqueLock &inodeLock) {
 	inodedata *parent = chunkData->getParent();
-	parent->lastWriteToDataChain.reset();
+	parent->lastWriteToDataChainMs.store(inodedata::nowSteadyMs());
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -900,9 +932,7 @@ void ChunkJobWriter::processDataChain(ChunkWriter &writer) {
 		    std::min(parent->maxfleng - chunkIndex_ * SFSCHUNKSIZE, (uint64_t)SFSCHUNKSIZE));
 		inodeLock.unlock();
 		if (writer.startNewOperations(can_expect_next_block) > 0) {
-			inodeLock.lock();
-			parent->lastWriteToChunkservers.reset();
-			inodeLock.unlock();
+			parent->lastWriteToChunkserversMs.store(inodedata::nowSteadyMs());
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -188,12 +199,14 @@ struct inodedata {
 	 */
 	bool requiresFlushing() {
 		return (flushwaiting > 0 ||
-		        lastWriteToDataChain.elapsed_ms() >= kMaximumTimeInDataChainSinceLastWrite_ms ||
-		        lastWriteToChunkservers.elapsed_ms() >= kMaximumTimeInDataChainSinceLastFlush_ms);
+		        elapsedMsSince(lastWriteToDataChainMs.load()) >=
+		            kMaximumTimeInDataChainSinceLastWrite_ms ||
+		        elapsedMsSince(lastWriteToChunkserversMs.load()) >=
+		            kMaximumTimeInDataChainSinceLastFlush_ms);
```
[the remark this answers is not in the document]
dmga44: Applied.

src/mount/writedata.cc
```diff
@@ -188,12 +199,14 @@ struct inodedata {
...
 	}
 
 	bool wasLastBlockRecentlyWritten() const {
-		return lastWriteToDataChain.elapsed_ms() < kVeryRecentWrite_ms;
+		return elapsedMsSince(lastWriteToDataChainMs.load()) < kVeryRecentWrite_ms;
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -416,25 +447,49 @@ void WriteHighLevelOp::startNextWriteJob() {
...
+		               __func__);
+	}
+
+	if (inputBuffer_ != nullptr && inputBuffer_->currentBlocks() > 0 && !inputBuffer_->isBeingUpdated()) {
+		writeDataBuffers_.emplace_back(std::move(inputBuffer_));
+		inputBuffersReadySinceLastFlush_++;
+	}
+
+	inputBuffersForNextFlush_.push(inputBuffersReadySinceLastFlush_);
+	inputBuffersReadySinceLastFlush_ = 0;
```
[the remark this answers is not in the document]
dmga44: Already applied.

tests/test_suites/ShortSystemTests/test_concurrent_random_writes_on_chunk.sh
```diff
@@ -34,12 +35,41 @@ switch_use_chunkserver_side_chunk_lock_thread() {
...
+	touch ${client_tweaking_loop_file}
+	while true; do
+		if [ ! -e ${client_tweaking_loop_file} ]; then
+			break
+		fi
+		# Make the test more robust by switching write flush packet usage on the fly on the
+		# client side, which can cause more interleaving of operations and increase chances of
+		# catching concurrency issues.
+		sleep 0.23
+		for mount in $(seq 0 3); do
```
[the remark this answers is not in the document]
dmga44: It is the very same idea, I'm not implementing this.

dmga44:
@gemini-code-assist Please review again.

src/mount/chunk_writer.h
```diff
@@ -45,9 +45,15 @@ class ChunkWriter {
 	 *        to write to them and read from them
 	 * \param dataChainFd - end of pipe; if anything is written to it, ChunkWriter will break its
 	 *        poll call and look for some new data in write cache for the currently written chunk
+	 * \param writeWindowSize - maximum number of operations that can be in progress at the same
+	 *        time; if this number is exceeded, no new operations can be started. When reached,
+	 *        triggers a flush of the data to chunkservers if the flush packet is enabled.
+	 * \param useWriteFlushPacket - if true, flush packet will be used to trigger flush of the data to
+	 *        chunkservers when the number of pending operations reaches writeWindowSize or when
```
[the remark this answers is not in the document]
dmga44: Applied, sort of.

src/common/write_executor.cc
```diff
@@ -97,11 +114,35 @@ void WriteExecutor::addDataPacket(uint32_t writeId,
...
+	cltocs::writeFlush::serialize(packet.buffer, chunkId_);
+	writeDataPacketsSinceLastFlush_ = 0;
+}
+
 void WriteExecutor::addEndPacket() {
 	sassert(isRunning_);
+
+	// For the cases when write prematurely due to write buffering.
+	// Won't do anything if already sent it.
+	addFlushPacket();
```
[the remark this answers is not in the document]
dmga44: Applied.

dmga44:
@gemini-code-assist Please review again.

src/mount/writedata.cc
```diff
@@ -188,12 +199,19 @@ struct inodedata {
 	 */
 	bool requiresFlushing() {
 		return (flushwaiting > 0 ||
-		        lastWriteToDataChain.elapsed_ms() >= kMaximumTimeInDataChainSinceLastWrite_ms ||
-		        lastWriteToChunkservers.elapsed_ms() >= kMaximumTimeInDataChainSinceLastFlush_ms);
+		        elapsedMsSince(lastWriteToDataChainMs.load()) >=
+		            kMaximumTimeInDataChainSinceLastWrite_ms ||
+		        elapsedMsSince(lastWriteToChunkserversMs.load(std::memory_order_relaxed)) >=
+		            kMaximumTimeInDataChainSinceLastFlush_ms);
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -188,12 +199,19 @@ struct inodedata {
...
+	}
+
+	bool wasLastBlockVeryRecentlyWritten() const {
+		return elapsedMsSince(lastWriteToDataChainMs.load()) < kVeryRecentWrite_ms;
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -1039,7 +1086,7 @@ void write_data_term(void) {
 int write_block(ChunkData *chunkData, uint16_t pos, uint32_t from, uint32_t to, const uint8_t *data,
                 UniqueLock &inodeLock) {
 	inodedata *parent = chunkData->getParent();
-	parent->lastWriteToDataChain.reset();
+	parent->lastWriteToDataChainMs.store(inodedata::nowSteadyMs());
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -1059,14 +1106,16 @@ int write_block(ChunkData *chunkData, uint16_t pos, uint32_t from, uint32_t to,
...
 	sassert(chunkData->dataChain.back().expand(from, to, data));
 	UniqueLock globalLock(gMutex, std::defer_lock);
 	if (chunkData->inQueue) {
 		// Consider some speedup if there are no errors and:
 		// - there is a lot of blocks in the write chain
 		// - there are at least two chunks in the write chain
 		if (chunkData->tryCounter == 0 &&
-		    (chunkData->dataChain.size() > chunkData->minimumBlocksToWrite)) {
+		    chunkData->dataChain.size() > chunkData->minimumBlocksToWrite &&
+		    chunkData->inDelayedQueue.load()) {
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -451,13 +478,15 @@ void write_free_chunkdata(ChunkData *fChunkData, UniqueLock &) {
 static void delayed_queue_put(ChunkData *chunkData, uint32_t seconds, UniqueLock &) {
 	delayedQueue.emplace_back(
 	    DelayedQueueEntry(chunkData, seconds * DelayedQueueEntry::kTicksPerSecond));
+	if (chunkData != kNoChunkData) { chunkData->inDelayedQueue.store(true); }
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -451,13 +478,15 @@ void write_free_chunkdata(ChunkData *fChunkData, UniqueLock &) {
...
 }
 
 /* globalLock: LOCKED*/
 static bool delayed_queue_remove(ChunkData *chunkData, UniqueLock &) {
 	for (auto it = delayedQueue.begin(); it != delayedQueue.end(); ++it) {
 		if (it->chunkData == chunkData) {
 			delayedQueue.erase(it);
+			chunkData->inDelayedQueue.store(false);
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -478,6 +507,7 @@ static std::vector<ChunkData *> delayed_queue_remove(inodedata *id, UniqueLock &
 	                         });
 
 	delayedQueue.erase(it, delayedQueue.end());
+	for (auto *chunkData : removedChunkData) { chunkData->inDelayedQueue.store(false); }
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/mount/writedata.cc
```diff
@@ -496,6 +526,7 @@ void *delayed_queue_worker(void *) {
 			if (--it->ticksLeft <= 0) {
 				auto *data = reinterpret_cast<uint8_t *>(it->chunkData);
 				jobsQueue->put(0, 0, data, 1);
+				it->chunkData->inDelayedQueue.store(false);
```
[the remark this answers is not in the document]
dmga44: It triggered helgrind issues, I think it is not worth it.

src/common/write_executor.cc
```diff
@@ -74,9 +77,20 @@ void WriteExecutor::addInitPacket() {
 			return first.chunkserver_version > second.chunkserver_version;
 	});
 
-	sassert((!chain_.empty() && chain_.front().chunkserver_version >= kFirstECVersion) ||
-	        chunkserver_version_ >= kFirstECVersion);
-	cltocs::writeInit::serialize(buffer, chunkId_, chunkVersion_, chunkType_, chain_);
+	auto minChunkserverVersion = chunkserver_version_;
+	if (!chain_.empty()) {
+		minChunkserverVersion = std::min(minChunkserverVersion, chain_.front().chunkserver_version);
+	}
```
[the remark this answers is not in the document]
dmga44: I'm applying this.

utils/chunk_operations_eio.c
```diff
@@ -75,7 +75,7 @@ static int err_on_operation(int fd, const char* opname, size_t offset, size_t si
 	// TODO: remove fixed pattern for SMRs after the basic support
 	if (strstr(filename, always_eio_trigger) || strstr(filename, "sauna_nullb0")) {
 		return EIO;
-	} else if (strstr(filename, far_eio_trigger) && offset > FAR_OFFSET_THRESHOLD) {
+	} else if (strstr(filename, far_eio_trigger) && offset + size > FAR_OFFSET_THRESHOLD) {
```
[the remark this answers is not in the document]
dmga44: Applied.

src/common/write_executor.cc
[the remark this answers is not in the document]
dmga44: I'm taking the last one (`.back`) element.

src/protocol/SFSCommunication.h
```diff
@@ -580,6 +580,10 @@ enum class SugidClearMode : uint8_t {
 #define SAU_CLTOCS_WRITE_DATA (1000U + 212U)
 /// chunkid:64 writeid:32 blocknum:16 offset:32 size:32 crc:32 data:BYTES[size]
 
+// 0x04C0
+#define SAU_CLTOCS_WRITE_FLUSH (1000U + 216U)
+/// chunkid:64
```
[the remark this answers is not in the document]
dmga44: Fixed.

dmga44:
@gemini-code-assist
Please review again.

src/mount/writedata.cc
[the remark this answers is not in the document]
dmga44: I'm adding the check to the `remove_if` lambda:
```cpp
if (entry.chunkData != kNoChunkData && entry.chunkData->getParent() == id)
```
but not inside the loop, because it was checked before.

src/chunkserver/chunk_high_level_ops.h
```diff
@@ -293,9 +307,64 @@ class WriteHighLevelOp : public HighLevelOp {
 	/// being processed or the chunk is not locked for writing.
 	/// Does not violates the order of replies for the write data blocks received from the client,
 	/// neither the amount of write buffering blocks available.
```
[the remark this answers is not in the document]
dmga44: Applied.

src/protocol/SFSCommunication.h
```diff
@@ -571,6 +571,7 @@ enum class SugidClearMode : uint8_t {
 #define SAU_CLTOCS_WRITE_INIT (1000U + 210U)
 /// version==0 chunkid:64 chunkversion:32 chunktype:8 chain:(N * [ip:32 port:16])
 /// version==1 chunkid:64 chunkversion:32 chunktype:16 chain:(N * [ip:32 port:16])
+/// version==2 chunkid:64 chunkversion:32 chunktype:16 chain:(N * [ip:32 port:16]) expectWriteFlush:1
```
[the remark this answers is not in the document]
dmga44: Applied.

tests/test_suites/SanityChecks/test_mount_tweaks.sh
```diff
@@ -14,8 +14,12 @@ expect_equals "false" "$(cat ${info[mount0]}/.saunafs_tweaks | grep -i directio
 expect_equals "true" "$(cat ${info[mount1]}/.saunafs_tweaks | grep -i directio | awk '{print $2}')"
 
 # Modify the options on the fly from tweaks file
+echo "WriteWindowSize=50" | sudo tee ${info[mount0]}/.saunafs_tweaks
+echo "UseWriteFlushPacket=true" | sudo tee ${info[mount0]}/.saunafs_tweaks
 echo "DirectIO=true" | sudo tee ${info[mount0]}/.saunafs_tweaks
 echo "DirectIO=false" | sudo tee ${info[mount1]}/.saunafs_tweaks
```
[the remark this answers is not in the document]
dmga44: That is not how the tweaks hidden file works.

dmga44:
@gemini-code-assist 
Can you please write a comprehensive description of the changes proposed in this PR? The intent is to help the reviewers.

src/mount/chunk_writer.cc
```diff
@@ -266,6 +286,10 @@ uint32_t ChunkWriter::getUnfinishedOperationsCount() {
 	return pendingOperations_.size() + newOperations_.size();
 }
 
+bool ChunkWriter::isWaitingForData() {
+	return getUnfinishedOperationsCount() < writeWindowSize_ + 1;
```
lgsilva3087: Separate from the intended `+ 1` window behavior, a user/admin could mistakenly configure an illogical `writeWindowSize_` value like `UINT32_MAX` through `sfswritewindowsize` or the `WriteWindowSize` tweak. With that value, `writeWindowSize_ + 1` wraps to `0`, so `isWaitingForData()` always returns false and the writer stops pulling normal writable blocks from the data chain. Could we write this as `getUnfinishedOperationsCount() <= writeWindowSize_` instead? That preserves the intended behavior for normal values and avoids this overflow edge case.

src/mount/writedata.cc
lgsilva3087, line 1098: On this fast path, `lastWriteToDataChainMs` was already refreshed at the top of `write_block()`, but `lastChunkDataWritten` is left unchanged. The new flush decision below uses `chunkData_ != parent->lastChunkDataWritten` to detect that the current chunk is no longer the last one written. With interleaved writes, a contiguous append to chunk A can return here while `lastChunkDataWritten` still points to chunk B, so A may enter flush mode earlier than intended even though it was just written. Should we set `parent->lastChunkDataWritten = chunkData` before this return as well?

lgsilva3087 (commented):
First pass.

doc/sfsmount.1.adoc
```diff
@@ -228,6 +229,12 @@ Equivalent to *--nonempty* option for use in fstab.
 Define the maximum number of chunks that can be written in parallel per inode.
 0 stands for unlimited (default: 0).
 
+*-o sfsusewriteflushpacket=*'0|1'::
+When set to 1, suggests the chunkserver to flush the packets received every
+window size blocks or when a write needs to finish (e.g. when flush() is called
```
lgsilva3087: Is `every window size blocks` here the same as option sfswritewindowsize?

src/mount/chunk_writer.cc
dmga44: Indeed, nice catch. Applied 👍

tests/tools/saunafs.sh
```diff
@@ -749,6 +750,20 @@ create_sfsmount_cfg_() {
 	local this_mount_exports_variable="MOUNT_${1}_EXTRA_EXPORTS"
 	echo "sfsmaster=$ip_address"
 	echo "sfsport=${saunafs_info_[matocl]}"
+
+	# Make the tests more robust by randomizing write flush packet usage if not set by the test.
+	# Each mount can override this via MOUNT_<id>_USE_WRITE_FLUSH_PACKET.
+	local this_mount_use_write_flush_packet_variable="MOUNT_${1}_USE_WRITE_FLUSH_PACKET"
+	local this_mount_use_write_flush_packet=${!this_mount_use_write_flush_packet_variable-}
+	if [[ ${this_mount_use_write_flush_packet} == NOT_SET ]]; then
```
lgsilva3087: With the global default of `NOT_SET` (line 37), every mount picks `sfsusewriteflushpacket` via `$((RANDOM % 2))` unless a test pins it. Good for coverage, but `$RANDOM` is never seeded and the resolved value only lands in the generated `$etcdir/sfsmount<id>.cfg` (this function's stdout is redirected there), never in the test log. Since CI could run in disposable containers, a failing run leaves no record of which mounts had flush on or off, and the combination can't be reproduced on a re-run.

Could we log the resolved value per mount, and/or make the seed recorded/overridable, so a flush-dependent failure is debuggable in CI?

Separately/optionally: this introduces a second way to set a mount option in the testing framework(`MOUNT_X_USE_WRITE_FLUSH_PACKET` alongside the existing `MOUNT_X_EXTRA_CONFIG`). Per-mount randomization is already expressible as `MOUNT_X_EXTRA_CONFIG="sfsusewriteflushpacket=$((RANDOM % 2))"`; the only thing the dedicated path adds is the suite-wide default. If we want to keep that default, a generic per-mount default hook would avoid growing an option-specific branch in shared harness code for every future toggle.

src/mount/writedata.cc
dmga44: Yes, that would work and it fixes the issue you are pointing out. I'm moving the update
```cpp
	parent->lastChunkDataWritten = chunkData;
```
to the top of the function because the write block should not fail.

tests/test_suites/TestTemplates/test_ec_parallel_writing.inc
```diff
@@ -4,8 +4,17 @@ else
...
+
 CHUNKSERVERS=4 \
 	MOUNTS=10 \
+	MOUNT_USE_WRITE_FLUSH_PACKET=1 \
+	MOUNT_0_USE_WRITE_FLUSH_PACKET=0 \
+	MOUNT_1_USE_WRITE_FLUSH_PACKET=0 \
+	MOUNT_2_USE_WRITE_FLUSH_PACKET=0 \
+	MOUNT_3_USE_WRITE_FLUSH_PACKET=0 \
+	MOUNT_4_USE_WRITE_FLUSH_PACKET=NOT_SET \
+	MOUNT_5_USE_WRITE_FLUSH_PACKET=NOT_SET \
```
lgsilva3087: Please see my comment in saunafs.sh

lgsilva3087 (commented):
Second pass.

doc/sfsmount.1.adoc
dmga44: Yes.

lgsilva3087: Maybe we can put it like:
`window size blocks (option sfswritewindowsize)`
Or something that correlate directly to the option.

dmga44: Changed to:
```md
When set to 1, suggests the chunkserver to flush the packets received every
window size blocks  (option sfswritewindowsize) or when a write needs to finish
(e.g. when flush() is called on the inode or when enough time has passed). Can
be updated via tweaks hidden file (default: 0).
```
applied.

tests/tools/saunafs.sh
dmga44: I've added a recording of the initial seed at the start of the `setup_local_empty_saunafs` function, and a way to set via a `RANDOM_SEED` environment variable.

I've also removed the extra `MOUNT_X_USE_WRITE_FLUSH_PACKET` which was redundant and left only `MOUNT_USE_WRITE_FLUSH_PACKET`. I have rearrenged the order in which the options are written to the mount cfg file.

tests/test_suites/TestTemplates/test_ec_parallel_writing.inc
dmga44: This was updated accordingly to the previous comment, and before it was logged what was the initial RANDOM value.

src/mount/chunk_writer.cc
lgsilva3087: Small follow-up: this landed as `< writeWindowSize_`, but the suggestion was `<= writeWindowSize_`, and the two aren't equivalent here.

`< writeWindowSize_ + 1` (the original) equals `<= writeWindowSize_`, which caps unfinished operations at `writeWindowSize_ + 1` -- i.e. effectively `writeWindowSize_` in flight once the trailing stripe-completion op is held back. `< writeWindowSize_` caps them at `writeWindowSize_` (effectively `writeWindowSize_ - 1`), which is the pre-change behavior this PR set out to improve, and it now contradicts the docs updated in this same change:

- chunk_writer.h (ctor): "intentionally allow up to writeWindowSize + 1 unfinished operations to keep at least writeWindowSize writes in progress"
- chunk_writer.h (`isWaitingForData()`): "unfinished operations is less than writeWindowSize + 1"

Could we use `getUnfinishedOperationsCount() <= writeWindowSize_` instead? That keeps the intended `+ 1` window for normal values and still avoids the `UINT32_MAX` overflow.

dmga44: Yes, you are right. I fixed it.

src/mount/writedata.cc
lgsilva3087: Thanks for moving this -- it does fix the fast path. One follow-up on the slow path: setting `lastChunkDataWritten` at the top happens before the lock is released. If `expand()` doesn't apply, we `inodeLock.unlock()` and can block in `write_cb_wait_for_block()`; during that window another writer on the same inode can run `write_block()` for chunk B and overwrite `lastChunkDataWritten = B`. When this thread relocks and appends to chunk A via `pushToChain()`, it no longer refreshes the pointer, so it can stay pointing at B even though A was just written -- the same premature-flush staleness, now on the slow path.

Could we set it in the two successful write paths instead of at the top -- right before the fast-path `return`, and right after `pushToChain()`? Both spots hold the inode lock with no intervening release, so neither can go stale.

dmga44: Considering the same case you are analyzing, i.e two writes on different chunks of the same inode, I think we are having the same issue, just that the updates will be a bit later. The `lastChunkDataWritten` will interleave both chunks being updated, and each one will have its own span of time in which it is the last one being written. The difference in your proposal is that the intervals in which one is the last one is shifted by a very small amount of time. Please also note that the real use of `lastChunkDataWritten` is as a heuristic check to speed up the switch to flush mode in the ChunkWriter sending some data, so some leeway is allowed.

 It does not worth IMO the need for splitting the update into two statements.

lgsilva3087: Agreed. Since `lastChunkDataWritten` only gates flush timing and the other triggers (`requiresFlushing`, the timeouts, window/end) cover liveness, the worst case is a single mistimed flush that self-corrects on the next write. Fine to leave as-is. Thanks for the explanation.

src/protocol/cltocs.h
```diff
@@ -112,6 +97,21 @@ inline void deserialize(const uint8_t* source, uint32_t sourceSize,
 			chunkId, chunkVersion, chunkType, chain);
 }
 
+inline void serialize(std::vector<uint8_t> &destination, uint64_t chunkId, uint32_t chunkVersion,
```
lgsilva3087, line 100: The two new wire formats, WRITE_INIT v2 (with `expectWriteFlush`) here, and the `writeFlush` namespace below, aren't covered in `cltocs_unittest.cc` (it round-trips WriteInit v1, WriteData, WriteEnd, but not these). Consider to add round-trip tests for both, mirroring the existing ones? Cheap guard against field-order/size drift.

src/common/saunafs_version.h
```diff
@@ -55,3 +55,4 @@ constexpr uint32_t kFirstVersionWithClusterId = saunafsVersion(5, 0, 0);
 constexpr uint32_t kFirstVersionWithReadTrashReservedByHandleOffset = saunafsVersion(5, 3, 0);
 constexpr uint32_t kFirstVersionWithUnlockChunkNotice = saunafsVersion(5, 8, 0);
 constexpr uint32_t kFirstVersionWithChunkserverSideChunkLock = saunafsVersion(5, 8, 0);
+constexpr uint32_t kFirstVersionWithWriteFlushPacket = saunafsVersion(5, 10, 0);
```
lgsilva3087: Don't forget to bump it to 5.11 when decided.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -349,7 +356,11 @@ void ChunkserverEntry::writeInit(const uint8_t *data, PacketHeader::Type type,
 		// Create a chain -- connect to the next chunkserver
 		fwdServer = chain[0].address;
 		chain.erase(chain.begin());
-		cltocs::writeInit::serialize(fwdInitPacket, chunkId, chunkVersion, chunkType, chain);
+		if (expectWriteFlush) {
+			cltocs::writeInit::serialize(fwdInitPacket, chunkId, chunkVersion, chunkType, expectWriteFlush, chain);
```
lgsilva3087: Format please

src/mount/fuse/mount_config.h
```diff
@@ -119,6 +119,7 @@ struct sfsopts_ {
 	bool directio;
 	int useinodebasedwritealgorithm;
 	unsigned maxchunkswritteninparallelperinode;
+	bool usewriteflushpacket;
```
lgsilva3087, line 122: Out of scope comment:
I think we need to upgrade this struct to use default member initializers.

src/mount/chunk_writer.h
```diff
@@ -45,9 +45,16 @@ class ChunkWriter {
 	 *        to write to them and read from them
 	 * \param dataChainFd - end of pipe; if anything is written to it, ChunkWriter will break its
 	 *        poll call and look for some new data in write cache for the currently written chunk
+	 * \param writeWindowSize - limit how many operations may be in flight; the writer may
```
lgsilva3087, line 48: Nice docs

lgsilva3087 (approved):
Great job. Please see my minor comments.

src/common/saunafs_version.h
dmga44: Bumped.

src/mount/fuse/mount_config.h
dmga44: I agree. This can be addressed in some misc refactor in future.

src/chunkserver/chunkserver_entry.cc
dmga44: Format applied.

src/protocol/cltocs.h
dmga44: I've added new unittests for the `writeFlush` and `WRITE_INIT v2` packets.

src/mount/writedata.cc
```diff
@@ -496,6 +515,7 @@ void *delayed_queue_worker(void *) {
...
 				it = delayedQueue.erase(it);
```
rolysr, line 530: Just in case, consider setting the `inDelayedQueue` as false for an element after you call the `delayedQueue.erase(...)` on it. My suggestion is to swap the order of these two lines.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -394,20 +394,51 @@ void WriteHighLevelOp::startNextWriteJob() {
...
-	std::vector<InputBuffer *> inputBuffers = {writeDataBuffers_.front().get()};
+	std::vector<InputBuffer *> inputBuffers;
+
+	if (expectWriteFlush_) {
+		enqueuedInputBuffers_ = inputBuffersForNextFlush_.front();
+		inputBuffersForNextFlush_.pop();
+
+		if (writeDataBuffers_.size() < enqueuedInputBuffers_) {
+			safs::log_err("({}) Fewer available write data buffers than expected ({} < {}).",
+			              __func__, writeDataBuffers_.size(), enqueuedInputBuffers_);
```
rolysr: If this is not critical scenario, consider if moving log to warning would be more precise, unless the state that this scenario leaves the chunkserver is completely invalid and some early return is needed.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -394,20 +394,51 @@ void WriteHighLevelOp::startNextWriteJob() {
...
+				safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+			}
+
+			inputBuffers.push_back(inputBuffersIt->get());
+			inputBuffersIt++;
+		}
+	} else {
+		if (writeDataBuffers_.front()->currentBlocks() == 0) {
+			safs::log_warn("({}) Called with no blocks in front InputBuffer.", __func__);
+		}
```
rolysr: Is it needed to warn about this `startNextWriteJob` function with no blocks in front on some InputBuffer? What is the information that could be fetched from these logs? If they are not need needed, consider to not adding them as logs.

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -667,7 +733,9 @@ void WriteHighLevelOp::delayedClose() {
 
 	// Pending write jobs will be handled in writeFinishedCallback, they need to write out the
 	// buffers
-	parentPendingWriteJobs() += writeDataBuffers_.size();
+	if (!writeDataBuffers_.empty()) {
+		parentPendingWriteJobs() += writeDataBuffers_.size() - enqueuedInputBuffers_ + 1;
```
rolysr, line 787: Is it ensured here that this substraction will always be non-negative one?

src/chunkserver/chunkserver_entry.cc
rolysr, line 353: Given that there is a new exception called here, please consider to print the message of received exception as part of the `catch{}` block logic.

src/chunkserver/chunkserver_entry.cc
```diff
@@ -446,6 +458,35 @@ void ChunkserverEntry::writeStatus(const uint8_t *data, PacketHeader::Type type,
...
+	TRACETHIS();
+	uint64_t opChunkId;
+
+	try {
+		cltocs::writeFlush::deserialize(data, length, opChunkId);
+	} catch (IncorrectDeserializationException&) {
+		safs::log_info("Received malformed WRITE_FLUSH message (length: {})", length);
+		state = State::IOFinish;
+		return;
+	}
```
rolysr, line 477: Also here, consider if it would be more useful for debugging to know the function and the exact exception message that triggered the exception `catch{}`

rolysr (commented):
Please take a look at my comments.

src/chunkserver/chunk_high_level_ops.cc
dmga44: Though I implemented the recovery from this escenario, and as you said, it is not critical this log tries to call the attention on this specific violation of the assumption that at least `enqueuedInputBuffers_` are present in `writeDataBuffers_`, which implies an error somewhere else in the code.

src/chunkserver/chunk_high_level_ops.cc
dmga44: They are needed because they show incosistencies: it is not expected empty InputBuffers in the `writeDataBuffers_` list.

src/chunkserver/chunk_high_level_ops.cc
dmga44: `enqueuedInputBuffers_` <= `writeDataBuffers_.size()` should hold, so the substraction should be non-negative.

src/chunkserver/chunkserver_entry.cc
dmga44: I've added the exception message to the log_info.

src/chunkserver/chunkserver_entry.cc
dmga44: I've added the exception message to the log_info.

src/mount/writedata.cc
dmga44: I moved it as suggested.

rolysr (approved):
Very good job @dmga44!

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -389,49 +389,160 @@ void WriteHighLevelOp::delayedCloseCallback(uint8_t status, void * /*entry*/) {
...
+// for easy side-by-side comparison.
+// ---------------------------------------------------------------------------
+
+bool WriteHighLevelOp::canStartWriteJobNow() const {
+	return !writeDataBuffers_.empty() && !isWriteJobBeingProcessed();
+}
+
+bool WriteHighLevelOpExpectFlush::canStartWriteJobNow() const {
+	if (!expectsWriteFlush_) { return WriteHighLevelOp::canStartWriteJobNow(); }
+	return !writeDataBuffers_.empty() && !isWriteJobBeingProcessed() &&
```
ralcolea, line 408: The logic for `expectWriteFlush_ == true` matches the logic of the base function.
Consider to call the base implementation and complement with the concrete difference to remove duplicated logic:
```cpp
bool WriteHighLevelOpExpectFlush::canStartWriteJobNow() const {
	if (!expectsWriteFlush_) { return WriteHighLevelOp::canStartWriteJobNow(); }
	return WriteHighLevelOp::canStartWriteJobNow() &&
	       !inputBuffersForNextFlush_.empty();
}
```

src/chunkserver/chunk_high_level_ops.cc
```diff
@@ -389,49 +389,160 @@ void WriteHighLevelOp::delayedCloseCallback(uint8_t status, void * /*entry*/) {
...
+	       !inputBuffersForNextFlush_.empty();
+}
+
+bool WriteHighLevelOp::shouldWriteCurrentInputBuffer() const {
+	return inputBuffer_ != nullptr && !inputBuffer_->isBeingUpdated() &&
+	       (inputBuffer_->isFull() || !isWriteJobBeingProcessed());
+}
+
+bool WriteHighLevelOpExpectFlush::shouldWriteCurrentInputBuffer() const {
```
ralcolea, line 416: In this function, there is also duplicated logic in the base and the concrete function.
Consider to refactor the common logic to avoid duplicating the logic.

src/mount/chunk_writer.h
```diff
@@ -80,7 +87,7 @@ class ChunkWriter {
 	 * Starts these added operations, which are worth starting right now.
 	 * Returns number of operations started.
 	 */
-	uint32_t startNewOperations(bool can_expect_next_block);
+	uint32_t startNewOperations(bool can_expect_next_block, bool wasLastBlockRecentlyWritten);
```
ralcolea, line 92: Consider to document the parameters of `startNewOperations()`, as it is with the `ChunkWriter` constructor.

src/mount/writedata.cc
```diff
@@ -930,14 +974,14 @@ void ChunkJobWriter::returnJournalToDataChain(std::list<WriteCacheBlock> &&journ
...
-bool ChunkJobWriter::haveBlockWorthWriting(uint32_t unfinishedOperationCount) {
+bool ChunkJobWriter::haveBlockWorthWriting(bool isWaitingForData) {
 	if (!haveAnyBlockInCurrentChunk()) { return false; }
 	const auto &block = chunkData_->dataChain.front();
 	if (block.type != WriteCacheBlock::kWritableBlock) {
 		// Always write data, that was previously written
 		return true;
-	} else if (unfinishedOperationCount >= gWriteWindowSize) {
-		// Don't start new operations if there is already a lot of pending writes
+	} else if (!isWaitingForData) {
```
ralcolea, line 983: The `else` is redundant after the `return`, consider removing it. The same with the final `else`.
Removing all the `else` is the best to have a more clear function.

tests/test_suites/ShortSystemTests/test_concurrent_random_writes_on_chunk.sh
```diff
@@ -34,12 +35,41 @@ switch_use_chunkserver_side_chunk_lock_thread() {
 	echo "switch_use_chunkserver_side_chunk_lock_thread stopped"
 }
 
+switch_use_write_flush_packet_thread() {
```
ralcolea, line 38: @dmga44 very nice addition to the test! 💪

tests/tools/saunafs.sh
```diff
@@ -747,14 +752,25 @@ function validate_and_append_fuse_options() {
 create_sfsmount_cfg_() {
 	local this_mount_cfg_variable="MOUNT_${1}_EXTRA_CONFIG"
 	local this_mount_exports_variable="MOUNT_${1}_EXTRA_EXPORTS"
+
+	# Make the tests more robust by randomizing write flush packet usage if not set by the test.
+	if [[ ${saunafs_info_[mount_use_write_flush_packet]} == NOT_SET ]]; then
```
ralcolea, line 757: My concern with this change is that it will affect most of the tests, even the ones non related with the new feature. In the future, it can cause failures in other tests due to the random usage of write flush packet. Consequently, it can mask the root cause of the failure, or even make some tests to behave flaky, making harder to identify the original failing scenario.

ralcolea (approved):
Very nice job @dmga44! 💪🔥🚀 
I shared some minor suggestions for your consideration, but none of them are stoppers to merge

src/chunkserver/chunk_high_level_ops.cc
dmga44: I actually decided to keep it this way in order to make the condition more readable. It can be also written like this:
```cpp
bool WriteHighLevelOpExpectFlush::canStartWriteJobNow() const {
	return WriteHighLevelOp::canStartWriteJobNow() &&
	       (!expectsWriteFlush_ || !inputBuffersForNextFlush_.empty());
}
```
but I think the clear statements are worth doubling some checks in the code.

tests/tools/saunafs.sh
dmga44: > In the future, it can cause failures in other tests due to the random usage of write flush packet.

If such failures occur then we must fix the implementation. The state of the random at the begining of the run is always printed, so it should be reproducible by taking that random seed value printed.

> Consequently, it can mask the root cause of the failure, or even make some tests to behave flaky

I think this is not the biggest concern here IMO:
- if some tests are flaky we should always fix them.
- all the tests should pass with or without the write flush packet enabled.

The issue I deem the most pressing is the lack of test coverage of this feature, and this is what these specific lines are about.

src/chunkserver/chunk_high_level_ops.cc
dmga44: I've added a new member of the `WriteHighLevelOp` class `isInputBufferFreezable` to consolidate this duplicated logic.

src/mount/chunk_writer.h
dmga44: I've added documentation for those parameters.

src/mount/writedata.cc
dmga44: Applied, I've removed the `else`.

tests/tools/saunafs.sh
ralcolea: I agree with you and I liked your reasoning!

PR #885, opened 2026-06-11
chore(fdb): drop per-miss key-not-found log in get

lgsilva3087:
A missing key is normal control flow in FDB. Transaction::get already signals it via an empty optional, and callers handle it. Logging every absent point read at info level flooded syslog. The file-test loop, for instance, scans inode ranges, with roughly two absent reads per inode, causing excessive unneeded logs.

Remove the log and the now-unused kv/kv_utils.h include. keyToEscapedAscii was its only user.

Related to: LS-817

### DIFF
--- a/src/fdb/fdb.cc
+++ b/src/fdb/fdb.cc
@@ -27,7 +27,6 @@
 
 #include "fdb/fdb.h"
 #include "fdb/fdb_future.h"
-#include "kv/kv_utils.h"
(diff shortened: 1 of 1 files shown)


ralcolea (approved):
LGTM! 👍

antuan96314 (approved):

dmga44 (approved):
LGTM.

PR #886, opened 2026-06-12
feat(master): add FDB transaction size limit

lgsilva3087:
The FoundationDB connector always ran with FDB's default 10 MB per-transaction size limit. Bugs where a transaction accumulates unbounded mutations or conflict ranges only surface at that scale, which is impractical to reach in short tests.

Allow lowering the cap via FDB_TRANSACTION_SIZE_LIMIT_BYTES, applied database-wide when the connector opens the database, following the FDB_TRANSACTION_TIMEOUT_MS pattern. Valid range is [32, 10000000] bytes; 0 (the default) keeps FDB's own limit. An out-of-range value or a failed option application aborts startup rather than running with a limit other than the operator asked for.

Related to: LS-817

### DIFF
--- a/src/master/kv_connector_fdb.cc
+++ b/src/master/kv_connector_fdb.cc
@@ -81,6 +81,38 @@ bool KVConnectorFDB::init() {
 		safs::log_info("FDB transaction timeout set to {} ms", timeoutMs);
 	}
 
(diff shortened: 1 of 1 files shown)


ralcolea (approved):
LGTM! 👍

rolysr (approved):
LGTM.

PR #890, opened 2026-06-16
feat(master): group-commit metadata write pipeline

lgsilva3087:
## Problem

On the KV backend every write RPC committed its own FDB transaction
synchronously, so each write paid a full commit-and-replicate durability
round-trip before replying. That round-trip dominates write latency and
caps write throughput: N concurrent writes paid N durability waits even
when they could have shared one.

The per-op commit was also a plain boolean commit() with no notion of a
retryable failure. A transient FDB conflict (e.g. not_committed, when
another writer touched the same keys) was treated like any other failure:
the handler logged an error and returned EIO to the client, even though
re-running the op on a fresh transaction would have succeeded. Each op
faced this on its own, with no shared retry surface.

## Solution

Route every write RPC through a group-commit batch on the KV backend. Op
bodies run on the event-loop thread and commit as a batch, so one
durability wait is amortized across the batch and conflict-replay is
bounded. Replaces the synchronous per-op commit for the KV backend.

No-op on the traditional in-memory master: with no read-write transaction
the body runs inline and finishes in place, so the master keeps its
existing behavior and only the KV backend batches.

Conflict handling is general, not batch-internal: members of one batch are
serialized on one transaction and cannot conflict with each other, so a
retryable conflict can only come from a writer outside the batch -- a
synchronous single-op handler, a chunkserver-driven write, or a future
second writer on the same keys. Any such conflict replays the whole batch
on a fresh transaction that re-reads current state, bounded by
MATOCL_MAX_COMMIT_RETRIES.

Builds on the KV foundation (#889, merged to dev).

Pipeline guarantees:
- Held bodies: while a batch commit is in flight, new ops are held
  with their bodies unrun and drained into the next batch once the
  commit is durable, so a body's in-memory side effects cannot be
  orphaned by an earlier op's rollback.
- Bounded replay: a batch commit conflict replays the whole batch on
  a fresh transaction, bounded by MATOCL_MAX_COMMIT_RETRIES.
- Reply gated on durability: the client reply runs only after the
  commit lands.
- Never replay an unknown result: the commit predicate is
  RETRYABLE_NOT_COMMITTED, so commit_unknown_result and commit-time
  timeout surface EIO instead of a blind replay.
- Killed-client safety: a run-always onCommit hook records persisted
  side effects (open-file acquire, delayed-chunk resolution) before
  the kill-gated reply, and process_chunk_status suppresses client
  packets for a gone client while still running lock-releasing
  commits.

Handlers on the pipeline: mknod open setattr seteattr truncate write_chunk
write_chunk_end unlink mkdir rmdir rename link symlink setxattr deleteacl
setacl setquota append purge, plus the chunk-status cleanup paths. No bare
write handlers remain.

Conflict-replay behavior is KV-backend-only: batching and replay need a
read-write transaction, which the traditional master never has. The
rewritten handlers and the in-memory open-acquire fix still run on the
master, so its existing regression coverage applies.

Known limitations on the KV backend:
- Sustained-conflict chunk retry: under sustained commit conflict,
  chunk-dispatch ops (truncate duplicate, write_chunk) re-enter a
  still-pending chunk operation on the synchronous batch replay and
  return CHUNKBUSY; the client retries. Self-heals under occasional
  conflict; the full fix is async chunk replay (later work).
- Truncate replay lockId: fixed (the body-allocated lockId survives
  batch replays, mirroring write_chunk) and verified by code reading;
  a deterministic injection regression is infeasible, so it is
  documented rather than tested.
- Conflict-free metadata (atime, dir times, nlink, quota) ships in a
  separate PR.

Related to: LS-860

### DIFF
--- a/src/common/event_loop.cc
+++ b/src/common/event_loop.cc
@@ -32,6 +32,7 @@
 #include "common/exception.h"
 #include "common/massert.h"
 #include "errors/sfserr.h"
+#include "kv/ifuture.h"
 
 #if defined(_WIN32)
   #include "common/sockets.h"
@@ -40,6 +41,25 @@
 ExitingStatus gExitingStatus = ExitingStatus::kRunning;
 bool gReloadRequested = false;
 static bool nextPollNonblocking = false;
+
+/// Runs one event-loop handler, containing retryable KV backend errors.
+///
+/// Op bodies replay these on a fresh transaction at the op boundary, but periodic
+/// and connection handlers (session sweeps, disconnect bookkeeping, ...) have no
+/// such boundary; before this guard an escaped kv::RetryableTransactionError
+/// (e.g. a transaction_timed_out read while the backend is slow or shutting
+/// down) reached std::terminate and aborted the whole server. The interrupted
+/// handler simply runs again on a later loop iteration.
+template <typename Handler, typename... Args>
+static void eventloop_run_handler(const char *kind, Handler &&handler, Args &&...args) {
+	try {
+		handler(std::forward<Args>(args)...);
+	} catch (const kv::RetryableTransactionError &e) {
+		safs::log_warn(
+		    "event loop: retryable backend error escaped a {} handler, skipping this run: {}", kind,
+		    e.what());
+	}
+}
 
 typedef struct pollentry {
 	void (*desc)(std::vector<pollfd>&);
@@ -256,11 +276,11 @@ void eventloop_run() {
 			}
 		} else {
 			for (auto &pollit : gPollEntries) {
-				pollit.serve(pdesc);
+				eventloop_run_handler("poll", pollit.serve, pdesc);
 			}
 		}
 		for (const FunctionEntry &fun : gEachLoopEntries) {
-			fun();
+			eventloop_run_handler("each-loop", fun);
 		}
 
 		uint64_t msecnow = usecnow / 1000;
@@ -308,17 +328,17 @@ void eventloop_run() {
 			if (timeit.millisecond_precision) {
 				if (msecnow >= timeit.nextevent) {
 					timeit.nextevent = msecnow + timeit.period;
-					timeit.fun();
+					eventloop_run_handler("timed", timeit.fun);
 				}
 				continue;
 			}
 
 			if (now >= timeit.nextevent) {
 				if (timeit.mode == TIMEMODE_RUN_LATE) {
-					timeit.fun();
+					eventloop_run_handler("timed", timeit.fun);
 				} else { /* timeit.mode == TIMEMODE_SKIP_LATE */
 					if (now == timeit.nextevent) {
-						timeit.fun();
+						eventloop_run_handler("timed", timeit.fun);
 					}
 				}
 				timeit.nextevent += ((now - timeit.nextevent + timeit.period)
--- a/src/master/filesystem_node.cc
+++ b/src/master/filesystem_node.cc
@@ -2299,8 +2299,19 @@ uint8_t FilesystemNodeOperationsBase::getNodeForOperation(
 
 		if (candidateNode == nullptr) { return SAUNAFS_ERROR_ENOENT; }
 	} else if (context.rootinode() == SPECIAL_INODE_ROOT || (context.rootinode() == 0)) {
-		candidateRoot = getRootNode(fsOpContext);
-		candidateNode = idToNode(fsOpContext, inode);
+		candidateRoot = nullptr;
+
+		if (inode == SPECIAL_INODE_ROOT) {
+			candidateRoot = getRootNode(fsOpContext);
+			candidateNode = candidateRoot;
+		} else {
+			candidateNode = idToNode(fsOpContext, inode);
+			// Only pay the root-node round-trip when the node resolved and a caller
+			// asked for the root; a null node returns ENOENT just below.
+			if (candidateNode != nullptr && rootDirOut != nullptr) {
+				candidateRoot = getRootNode(fsOpContext);
+			}
+		}
 
 		if (candidateNode == nullptr) { return SAUNAFS_ERROR_ENOENT; }
 
--- a/src/master/filesystem_operation_context.h
+++ b/src/master/filesystem_operation_context.h
@@ -22,7 +22,9 @@
 
 #include <memory>
 
+#include "common/type_defs.h"
 #include "kv/itransaction.h"
+#include "kv/kv_types.h"
 
 namespace kv {
 class IReadOnlyTransaction;
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -2834,6 +2834,10 @@ uint8_t FilesystemOperationsBase::acquire(const FsContext &context,
 
 	if (std::find(fileNode->sessionIds.begin(), fileNode->sessionIds.end(), sessionid) !=
 	    fileNode->sessionIds.end()) {
+		// Reached only when openFilesSet does not track the inode (the in-memory master
+		// couples acquire with that insert), so a session already in sessionIds means the two
+		// have drifted -- a structural inconsistency, not a retry. The KV backend defers the
+		// insert and overrides acquire() to tolerate the duplicate.
 		return SAUNAFS_ERROR_EINVAL;
 	}
 
--- a/src/master/matoclserv.cc
+++ b/src/master/matoclserv.cc
@@ -32,14 +32,20 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/eventfd.h>
 #include <sys/resource.h>
 #include <sys/types.h>
 #include <sys/uio.h>
 #include <syslog.h>
 #include <time.h>
 #include <unistd.h>
+#include <algorithm>
+#include <atomic>
+#include <bit>
 #include <cstdint>
+#include <deque>
 #include <fstream>
+#include <functional>
 #include <memory>
 
 #include "common/charts.h"
@@ -71,6 +77,7 @@
 #include "common/type_defs.h"
 #include "common/user_groups.h"
 #include "config/cfg.h"
+#include "kv/ifuture.h"
 #include "kv/itransaction.h"
 #include "master/changelog.h"
 #include "master/chartsdata.h"
@@ -124,18 +131,29 @@ struct matoclserventry;
 // locked chunks
 
 struct DelayedChunkOperation {
-	uint64_t chunkId;       ///< Chunk ID
-	uint64_t fileLength;    ///< File length
-	uint32_t lockId;        ///< Lock ID
-	uint32_t messageId;     ///< Message ID for reply
-	inode_t inode;          ///< Inode
-	uint32_t uid;           ///< Remapped uid of the user which will run the operation
-	uint32_t gid;           ///< Remapped gid of the user which will run the operation
-	uint32_t auid;          ///< Real uid (not remapped) of the user who will run the operation
-	uint32_t agid;          ///< Real gid (not remapped) of the user who will run the operation
-	uint8_t type;           ///< Delayed operation type: FUSE_WRITE, FUSE_TRUNCATE,
-	                        ///< FUSE_TRUNCATE_BEGIN or FUSE_TRUNCATE_END
-	const PacketSerializer* serializer;  ///< Packet serializer for the operation
+	uint64_t chunkId = 0;     ///< Chunk ID
+	uint64_t fileLength = 0;  ///< File length
+	uint32_t lockId = 0;      ///< Lock ID
+	uint32_t messageId = 0;   ///< Message ID for reply
+	inode_t inode = 0;        ///< Inode
+	uint32_t uid = 0;         ///< Remapped uid of the user which will run the operation
+	uint32_t gid = 0;         ///< Remapped gid of the user which will run the operation
+	uint32_t auid = 0;        ///< Real uid (not remapped) of the user who will run the operation
+	uint32_t agid = 0;        ///< Real gid (not remapped) of the user who will run the operation
+	uint8_t type = 0;         ///< Delayed operation type: FUSE_WRITE, FUSE_TRUNCATE,
+	                          ///< FUSE_TRUNCATE_BEGIN or FUSE_TRUNCATE_END
+	const PacketSerializer* serializer = nullptr;  ///< Packet serializer for the operation
+
+	// Deferred-commit handshake. The op body that initiates a chunkserver operation
+	// enqueues this entry BEFORE its own commit is durable (the deferred commit no
+	// longer blocks the event loop, so the chunkserver's status can arrive first). While
+	// commitPending, matoclserv_chunk_status stashes the status here instead of
+	// processing it; the commit continuation then clears the flag and processes any
+	// stashed status (or drops the entry if the commit failed).
+	bool commitPending = false;     ///< enqueuing op's commit still in flight
+	bool statusArrived = false;     ///< chunkserver status arrived while commitPending
+	uint8_t pendingStatus = 0;      ///< stashed chunkserver status
+	bool pendingIsFailedCreate = false;  ///< stashed isFailedCreateOperation flag
 };
 
 ///< This looks to be the client type.
@@ -190,6 +208,10 @@ struct matoclserventry {
 	AdminTask adminTask;  ///< admin task requested by this client
 	///< Delayed chunk operations for this client
 	std::vector<std::unique_ptr<DelayedChunkOperation>> delayedChunkOperations;
+	///< Number of async commits still in flight that will write a reply to this
+	///< client. The entry must not be freed while this is non-zero (see the close
+	///< loop in matoclserv_serve), so deferred replies never touch a dangling eptr.
+	uint32_t pendingCommits = 0;
 };
 
 using WaitEntry = std::tuple<matoclserventry *, inode_t, uint32_t>;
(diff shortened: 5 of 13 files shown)


src/master/session_manager.h
```diff
@@ -153,6 +153,22 @@ class ISessionManager {
 	virtual int insertOpenFile(const FilesystemOperationContext &fsOpContext,
 	                           Session *currentSession, inode_t inode) = 0;
 
+	/// Backend-persist-only half of insertOpenFile: runs IFilesystemOperations::acquire()
+	/// inside @p fsOpContext but does NOT touch Session::openFilesSet. The caller
+	/// records the in-memory open file (addOpenFile) only after the transaction
```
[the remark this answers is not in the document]
lgsilva3087: Fixed: comment now names `recordOpenFile` (the live-`Session` in-memory half this pairs with), not `addOpenFile` (the synthetic-session path the KV backend rejects).

src/master/matoclserv.cc
```diff
@@ -211,6 +230,533 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+// inode is burned. N < MATOCL_MAX_COMMIT_RETRIES recovers; N >= it exhausts to a bounded
+// EIO with no abort. Reloadable. Leave 0 in production.
+static uint32_t gDebugInjectReadConflicts = 0;
+
+// eventfd the commit future's ready callback writes to (from the backend network
+// thread, or inline from the event-loop thread) so poll() wakes immediately to
+// finalize the commit instead of waiting out the poll timeout. -1 until
+// matoclserv_network_init creates it. Linux-only; without it the group commit path
+// still works, just serviced at the poll-timeout cadence.
+static int gCommitWakeupFd = -1;
```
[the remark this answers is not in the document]
lgsilva3087: Fixed: `gCommitWakeupFd` is now `std::atomic<int>`, loaded once per use, with `exchange(-1)` before `close()` at term so a racing backend-thread callback reads -1. Caveat: the shutdown fd-reuse window is narrowed, not fully closed (the write is best-effort/`EBADF`); the int data race is gone.

src/master/matoclserv.cc
```diff
@@ -487,121 +1033,172 @@ uint8_t matoclserv_fuse_write_chunk_respond(matoclserventry *eptr,
...
+		matoclserv_submit_op(
+		    eptr, std::move(runCleanup),
+		    [inode, chunkId](uint8_t commitStatus) {
+			    if (commitStatus != SAUNAFS_STATUS_OK) {
+				    safs::log_err(
+				        "matoclserv_process_chunk_status: transaction failed to commit "
+				        "while releasing the write chunk lock: inode {}, chunkId {}",
+				        inode, chunkId);
+			    }
+		    });
```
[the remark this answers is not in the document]
lgsilva3087: Fixed: the commit-failure log moved to the always-run `onCommit` hook (was kill-gated `finish`). The cleanup commit already ran regardless; `finish` is now empty and `finish_member` null-checks it.

src/master/matoclserv.cc
```diff
@@ -487,121 +1033,172 @@ uint8_t matoclserv_fuse_write_chunk_respond(matoclserventry *eptr,
...
+				    if (commitStatus != SAUNAFS_STATUS_OK) {
+					    safs::log_err(
+					        "matoclserv_process_chunk_status: transaction failed to commit "
+					        "after endSetLength: inode {}, chunkId {}",
+					        inode, chunkId);
+				    }
+				    std::vector<uint8_t> reply;
+				    serializer->serializeFuseTruncate(reply, operationType, messageId, status);
+				    matoclserv_createpacket(eptr, std::move(reply));
+			    });
```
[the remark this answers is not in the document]
lgsilva3087: Fixed: split the continuation. Log moved to `onCommit` (run-always); reply stays in `finish` (kill-gated).

src/master/session_manager.cc
```diff
@@ -112,6 +112,22 @@ int SessionManagerBase::insertOpenFile(const FilesystemOperationContext &fsOpCon
 	return status;
 }
 
+int SessionManagerBase::acquireOpenFilePersist(const FilesystemOperationContext &fsOpContext,
+                                               Session *currentSession, inode_t inode) {
+	if (currentSession->openFilesSet.contains(inode)) {
```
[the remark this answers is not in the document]
lgsilva3087: Leaving as-is. This mirrors the pre-existing `insertOpenFile`, which does the same `currentSession->` access with no `passert`; the precondition is unchanged, so this PR adds no new null-deref risk. Asserting only here and not in `insertOpenFile` would be the one-off inconsistency the reference warns against, and hardening all three is a separate codebase-wide pass.

src/master/session_manager.cc
```diff
@@ -112,6 +112,22 @@ int SessionManagerBase::insertOpenFile(const FilesystemOperationContext &fsOpCon
...
+	}
+
+	// Persist only; the in-memory openFilesSet insert is deferred to commit success
+	// by the caller so a commit retry replays this acquire on a fresh transaction.
+	return gFSOperations->acquire(FsContext::getForMaster(eventloop_time()), fsOpContext, inode,
+	                              currentSession->sessionId);
+}
+
+void SessionManagerBase::recordOpenFile(Session *currentSession, inode_t inode) {
+	currentSession->openFilesSet.insert(inode);
```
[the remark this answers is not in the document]
lgsilva3087: Same as the `acquireOpenFilePersist` thread: this mirrors the pre-existing `insertOpenFile` (no `passert` there). No new null risk is introduced here; consistent hardening is out of scope for this PR.

src/master/matoclserv.cc
```diff
@@ -487,121 +1037,176 @@ uint8_t matoclserv_fuse_write_chunk_respond(matoclserventry *eptr,
...
-
-					chunkOpsIterator = eaptr->delayedChunkOperations.erase(chunkOpsIterator);
-				} else {
-					++chunkOpsIterator;
-				}
-			}
+	for (const auto &entryPtr : matoclservList) {
+		matoclserventry *eaptr = entryPtr.get();
+		if (eaptr->mode == ClientConnectionMode::KILL) { continue; }
+		auto &queue = eaptr->delayedChunkOperations;
```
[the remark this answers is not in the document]
lgsilva3087: This KILL skip is pre-existing (already in dev's `matoclserv_chunk_status`), not introduced by this PR, so it is out of scope here. The killed-client cleanup this PR adds is reached via the on-commit resolve hook; full drain of a disconnected client's delayed ops on fresh-status arrival is a pre-existing gap tracked separately.

src/master/matoclserv.cc
```diff
@@ -487,121 +1037,176 @@ uint8_t matoclserv_fuse_write_chunk_respond(matoclserventry *eptr,
...
-			serializer->serializeFuseWriteChunk(reply, messageId, status);
-			matoclserv_createpacket(eptr, std::move(reply));
+		} else if (!alive) {
+			return;  // killed client: no reply; the chunk stays locked as on any success
 		} else {
+			// The chunkserver write succeeded: reply with the chunk version and locations.
+			// This can still fail (e.g. NOCHUNK), in which case the lock is released below.
 			status = matoclserv_fuse_write_chunk_respond(eptr, serializer,
 					chunkId, messageId, fileLength, lockId);
 		}
```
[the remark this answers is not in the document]
lgsilva3087: Pre-existing behavior: dev already skips KILL clients in the dispatcher, so this branch never ran for a disconnected client there either, and the lock-on-disconnect is released by the same path as before. Not a regression from this PR; the broader killed-client lock-release improvement is a separate follow-up.

src/master/matoclserv.cc
```diff
@@ -6216,6 +6988,13 @@ void matoclserv_serve(const std::vector<pollfd> &pdesc) {
 	for (auto eptrIt = matoclservList.begin(); eptrIt != matoclservList.end();) {
 		auto *eptr = eptrIt->get();
 		if (eptr->mode == ClientConnectionMode::KILL) {
+			// Defer teardown while batch commits still reference this entry; their
+			// replies are dropped in matoclserv_finish_member() and the entry is
+			// freed on a later iteration once pendingCommits hits zero.
+			if (eptr->pendingCommits > 0) {
+				++eptrIt;
+				continue;
+			}
```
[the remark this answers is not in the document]
lgsilva3087: The `pendingCommits`-gated defer is new and an improvement (dev tore KILL entries down unconditionally). The `delayedChunkOperations` drain gap you note is pre-existing, not introduced here. Extending the defer to cover delayed chunk ops is a reasonable follow-up, but out of scope for this PR.

src/master/filesystem_operations.cc
```diff
@@ -2834,7 +2834,9 @@ uint8_t FilesystemOperationsBase::acquire(const FsContext &context,
 
 	if (std::find(fileNode->sessionIds.begin(), fileNode->sessionIds.end(), sessionid) !=
 	    fileNode->sessionIds.end()) {
-		return SAUNAFS_ERROR_EINVAL;
+		// Idempotent: already acquired by this session (e.g. a changelog replay or a
+		// resync after a rolled-back release). Not a client-visible error.
+		return SAUNAFS_STATUS_OK;
```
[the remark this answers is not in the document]
lgsilva3087: Fixed: updated the `acquire()` interface doc to document the idempotent OK return on a duplicate acquire by the same session (was EINVAL). The only callers are in `session_manager.cc` and they key on OK, not EINVAL, so no call-site logic changes.

src/master/filesystem_node.cc
```diff
@@ -2299,8 +2299,15 @@ uint8_t FilesystemNodeOperationsBase::getNodeForOperation(
...
-		candidateNode = idToNode(fsOpContext, inode);
+		candidateRoot = nullptr;
+
+		if (inode == SPECIAL_INODE_ROOT) {
+			candidateRoot = getRootNode(fsOpContext);
+			candidateNode = candidateRoot;
+		} else {
+			candidateNode = idToNode(fsOpContext, inode);
+			if (rootDirOut != nullptr) { candidateRoot = getRootNode(fsOpContext); }
+		}
```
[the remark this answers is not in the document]
lgsilva3087: Fixed: now only fetches the root node when the inode resolved and a caller requested it (`candidateNode != nullptr && rootDirOut != nullptr`); a null node returns ENOENT just below, so the round-trip is skipped on that path.

src/master/matoclserv.cc
```diff
@@ -487,121 +1037,176 @@ uint8_t matoclserv_fuse_write_chunk_respond(matoclserventry *eptr,
...
+			removeChunk = isFailedCreateOperation;
+			if (alive) {
+				serializer->serializeFuseWriteChunk(reply, messageId, status);
+				matoclserv_createpacket(eptr, std::move(reply));
 			}
-			serializer->serializeFuseWriteChunk(reply, messageId, status);
-			matoclserv_createpacket(eptr, std::move(reply));
+		} else if (!alive) {
+			return;  // killed client: no reply; the chunk stays locked as on any success
 		} else {
```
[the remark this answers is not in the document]
lgsilva3087: You're right, and thanks for the push. `matoclserv_resolve_pending_delayed_op` calls `process_chunk_status` directly for a KILL client when a status was stashed before disconnect, so the success `!alive` return was reachable and stranded the lock. Fixed: the killed-client success case now falls through to the cleanup commit (writeEnd, releasing the lock and keeping the chunk) instead of returning.

src/master/filesystem_operation_context.h
```diff
@@ -88,6 +92,31 @@ class FilesystemOperationContext {
...
+	/// so it never serves data across commits: it relies only on the
+	/// transaction's own read-your-writes and MVCC snapshot, which makes it
+	/// correct regardless of how many master processes share the store. Methods are
+	/// const with a mutable backing map because lookups happen on const contexts.
+	///
+	/// @param inode The inode whose cached bytes to look up.
+	/// @return Pointer to the memoized entry (engaged optional = present node,
+	///         disengaged optional = proven-absent node), or nullptr if this
+	///         inode has not been read in this operation yet.
+	const std::optional<kv::Value> *cachedNodeBytes(inode_t inode) const;
```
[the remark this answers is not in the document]
lgsilva3087: Resolved by removing the cache: `cachedNodeBytes`/`cacheNodeBytes`/`invalidateCachedNode` + `nodeBytesCache_` had no consumer in either repo (dead code from an earlier spike), so the fragile-pointer API is gone. It will return with its backend reader in the create-throughput perf work. (Minor: for `std::unordered_map`, element pointers/references are stable across insert and rehash; only erasing that key invalidates - but moot now.)

src/master/filesystem_operations.cc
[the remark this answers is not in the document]
dmga44: Most of the calls to acquire check before if the file has been acquired by the session, so the previous `SAUNAFS_ERROR_EINVAL` showed a structural inconsistency.

I get the idea of making it idempotent if needed, but the reasons posted there don't seem clear to me. The changelog replay should not go here because the metadata state in a shadow/metarestore should be equal to the one saving the ACQUIRE in the first place. A failed release should check if the acquired makes sense before calling it again IMO.

src/master/matoclserv.h
```diff
@@ -60,3 +60,24 @@ void matoclserv_broadcast_metadata_checksum_recalculated(uint8_t status);
 
 /// Check whether there are any async filesystem operations that still need to finished (e.g delayed chunk operations)
 bool matoclserv_client_async_operations_finished();
+
+/// Group-commit (spec 7.5) runtime counters, for diagnosing how well batching amortizes
```
dmga44: spec 7.5?

src/master/matoclserv.h
```diff
@@ -60,3 +60,24 @@ void matoclserv_broadcast_metadata_checksum_recalculated(uint8_t status);
...
+/// commit (the headline number: near 1 means batching is not engaging, latency-bound).
+/// Counted on the single event-loop thread; gathered only while batch-stats accounting is
+/// enabled (see matoclserv_set_batch_stats_enabled).
+struct MatoclBatchStats {
+	uint64_t committedBatches = 0;  ///< batches whose single commit became durable
+	uint64_t committedOps = 0;      ///< total member ops across those batches
+	uint64_t maxBatchSize = 0;      ///< largest committed batch
+	uint64_t batchReplays = 0;      ///< whole-batch conflict replays (external writer)
+	uint64_t heldOps = 0;           ///< ops parked because a batch commit was in flight
+	uint64_t sizeBuckets[5] = {0};  ///< committed-batch size histogram: 1, 2-4, 5-8, 9-16, 17+
```
dmga44: Cool!

tests/test_suites/ShortSystemTests/test_ec_truncate_atomicity.sh
```diff
@@ -1,4 +1,8 @@
-timeout_set '1 minute'
+# 3 minutes: with the master's group-commit pipeline, the
+# many-writers-plus-truncates-on-one-file pattern resolves commit conflicts by replaying
+# ops (optimistic concurrency), which trades single-file hot-spot latency for op-mix
+# throughput. 1 minute was already tight for this test (~41s) before that change.
```
dmga44: Consider:
`The commit conflict resolution in masters using KV backends can require more time than the in-memory master.`
And set the timeout depending on the backend used.

dmga44 (commented):
Pass 1.

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+			continue;
+		}
+		safs::log_err("matoclserv: sync commit failed (err {}, retryable {}), giving up", commitError,
+		              retryable);
+		return SAUNAFS_ERROR_IO;
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Group commit (spec 7.5). Many concurrently-arrived write ops share ONE backend
```
dmga44: spec 7.5?

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+		FilesystemOperationContext ctx = gFSOperations->createFilesystemOperationContext(
+		    FilesystemOperationContext::TransactionType::kReadWrite);
+		uint8_t status;
+		try {
+			if (attempt < gDebugInjectReadConflicts) {
+				// Fault injection (no commit, no write): force the retry path.
+				throw kv::RetryableTransactionError(
+				    1031, "injected retryable read conflict (MATOCL_DEBUG_INJECT_READ_CONFLICTS)");
+			}
+			status = body(ctx);
```
dmga44, line 349: So, if I understood correctly this `body` function should only be able to perform checks and build a set of changes to be applied in KV backends. This is kind of a strong constraint for the operations using it, like the `writeEnd` that uses it now.

I'm assuming that changes in the MDS state / structures held in memory won't be applied in the future implementation of the function wrapped here.

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+	bool hasContext = false;  ///< ctx is created lazily per batch
+};
+
+/// A batch whose single commit is in flight.
+struct InFlightBatch {
+	FilesystemOperationContext ctx;             ///< owns the txn; must outlive `future`
+	std::unique_ptr<kv::ICommitFuture> future;  ///< declared after ctx: destroyed first
+	std::vector<BatchMember> members;
+	uint32_t attempt = 0;  ///< whole-batch commit attempts (conflict replays)
+};
```
dmga44: Consider leaving only one:
```cpp
struct OpBatch {
	FilesystemOperationContext ctx;             ///< owns the txn; must outlive `future`
	std::unique_ptr<kv::ICommitFuture> future;  ///< declared after ctx: destroyed first
	std::vector<BatchMember> members;
	uint32_t attempt = 0;  ///< whole-batch commit attempts (conflict replays)
	bool hasContext = false;  ///< ctx is created lazily per batch
};
```

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+
+/// A converted op waiting for the in-flight batch to resolve; body NOT run yet.
+struct HeldOp {
+	matoclserventry *eptr;
+	OpReplay body;
+	std::function<void(uint8_t status)> finish;
+	std::function<void(uint8_t status)> onCommit;  ///< carried to the BatchMember on join
+};
```
dmga44: This is a subset of `BatchMember`, consider removing and use always `BatchMember`. Should help in `matoclserv_poll_batch`.

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+static std::optional<InFlightBatch> gInFlightBatch;
+static std::deque<HeldOp> gHeldOps;
+// True while matoclserv_poll_batch is resolving an in-flight batch (running
+// finishes / replaying members). A finish can submit a follow-up op (e.g. the
+// truncate finalize via matoclserv_resolve_pending_delayed_op); it must be HELD
+// even in the window where gInFlightBatch is already reset, or its body would
+// run against state the resolution is still rebuilding.
+static bool gBatchResolving = false;
+
+// Group-commit batch-stats accounting (spec 7.5 instrumentation). Off by default; the master
```
dmga44: spec 7.5 instrumentation?

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+MatoclBatchStats matoclserv_get_batch_stats() { return gBatchStats; }
+
+/// Records a committed batch of `size` members in the stats (size-histogram bucketed as
+/// 1, 2-4, 5-8, 9-16, 17+). No-op unless accounting is enabled.
+static void matoclserv_record_committed_batch(size_t size) {
+	if (!gBatchStatsEnabled || size == 0) { return; }
+	gBatchStats.committedBatches++;
+	gBatchStats.committedOps += size;
+	if (size > gBatchStats.maxBatchSize) { gBatchStats.maxBatchSize = size; }
+	size_t bucket = size <= 1 ? 0 : size <= 4 ? 1 : size <= 8 ? 2 : size <= 16 ? 3 : 4;
```
dmga44: Consider:
```cpp
	size_t bucket = std::min<size_t>(std::bit_width(size) - 1, 4);
```
but the groups will be changed to 1, 2-3, 4-7, 8-15, 16+.

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+					    "attempt {}",
+					    e.errorCode(), member.attempts);
+				}
+				restart = true;
+				break;
+			}
+		}
+		if (!restart) { return ctx; }
+	}
+}
```
dmga44, line 537: Consider, in order to remove similar code the following function (check next comment also):
```cpp
static bool matoclserv_replay_members_iteration(std::vector<BatchMember> &members, size_t &idx,
                                                FilesystemOperationContext &ctx) {
	auto *txn = ctx.getReadWriteTransaction();
	BatchMember &member = members[idx];
	const uint64_t mutationsBefore = txn != nullptr ? txn->mutationCount() : 0;
	try {
		if (member.attempts < gDebugInjectReadConflicts) {
			// Fault injection (no write burned): force the read-retry path.
			throw kv::RetryableTransactionError(
			    1031, "injected retryable read conflict (MATOCL_DEBUG_INJECT_READ_CONFLICTS)");
		}
		uint8_t status = member.replay(ctx);
			++idx;
			return false;  // continue replaying
		}
		const bool poisoned = txn != nullptr && txn->mutationCount() != mutationsBefore;
		matoclserv_finish_member(member, status);
		members.erase(members.begin() + idx);
		if (poisoned) {
			return true;  // restart replay on a fresh transaction
		}
	} catch (const kv::RetryableTransactionError &e) {
		// The shared transaction is invalid after a thrown read error:
		// restart on a fresh one regardless of what this member did.
		if (member.attempts >= gMaxCommitRetries) {
			safs::log_warn(
			    "matoclserv: batch member read retry exhausted after {} attempts (err {}): {}",
			    member.attempts, e.errorCode(), e.what());
			matoclserv_finish_member(member, SAUNAFS_ERROR_IO);
			members.erase(members.begin() + idx);
		} else {
			member.attempts++;
			safs::log_info(
			    "matoclserv: retryable read error in batch member (err {}), replaying batch, "
			    "attempt {}",
			    e.errorCode(), member.attempts);
		}
		return true;  // restart replay on a fresh transaction
	}
	return false;  // continue replaying
}
```

if used, the `matoclserv_replay_members` can be summarized in:
```cpp
static FilesystemOperationContext matoclserv_replay_members(std::vector<BatchMember> &members) {
	for (;;) {
		FilesystemOperationContext ctx = gFSOperations->createFilesystemOperationContext(
		    FilesystemOperationContext::TransactionType::kReadWrite);
		bool restart = false;
		for (size_t idx = 0; idx < members.size() && !restart;) {
			restart = matoclserv_replay_members_iteration(members, idx, ctx);
		}
		if (!restart) { return ctx; }
	}
}
```

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+			member.attempts++;
+			safs::log_info(
+			    "matoclserv: retryable read error in op body (err {}), replaying batch, attempt {}",
+			    e.errorCode(), member.attempts);
+			gOpenBatch.members.push_back(std::move(member));
+		}
+		gOpenBatch.ctx = matoclserv_replay_members(gOpenBatch.members);
+	}
+	matoclserv_release_idle_open_batch();
+}
```
dmga44, line 578: If `matoclserv_replay_members_iteration` is used also here then it can be summarized as:
```cpp
static void matoclserv_join_open_batch(BatchMember member) {
	if (!gOpenBatch.hasContext) {
		gOpenBatch.ctx = gFSOperations->createFilesystemOperationContext(
		    FilesystemOperationContext::TransactionType::kReadWrite);
		gOpenBatch.hasContext = true;
	}
	auto *txn = gOpenBatch.ctx.getReadWriteTransaction();
	if (txn == nullptr) {
		// In-memory Master: the body applies in place; reply on its status.
		uint8_t status = member.replay(gOpenBatch.ctx);
		matoclserv_finish_member(member, status);
		matoclserv_release_idle_open_batch();
		return;
	}

	size_t idx = gOpenBatch.members.size();
	gOpenBatch.members.push_back(std::move(member));
	if (matoclserv_replay_members_iteration(gOpenBatch.members, idx, gOpenBatch.ctx)) {
		gOpenBatch.ctx = matoclserv_replay_members(gOpenBatch.members);
	}
	matoclserv_release_idle_open_batch();
}
```

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+/// Entry point for converted write-op handlers: routes the op into the group-commit
+/// batch. `finish` always runs exactly once with the op/commit status; the optional
+/// `onCommit` durable hook runs first, with the same status, even for a killed client.
+static void matoclserv_submit_op(matoclserventry *eptr, OpReplay body,
+                                 std::function<void(uint8_t status)> finish,
+                                 std::function<void(uint8_t status)> onCommit = {}) {
+	eptr->pendingCommits++;
+	if (gInFlightBatch.has_value() || gBatchResolving) {
+		// Hold the body until the in-flight batch is durable so it runs against
+		// committed state (the soundness core of this design; see spec 7.5).
```
dmga44: see spec 7.5?

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+				                                     std::move(members), attempt});
+			}
+			// Empty members: every op was replied during the replay. No txn means
+			// the in-memory Master, where replays applied in place -- their
+			// finishes already ran for error cases; OK members are finished here.
+			else {
+				for (BatchMember &member : members) {
+					matoclserv_finish_member(member, SAUNAFS_STATUS_OK);
+				}
+			}
```
dmga44: Please correct me if I'm wrong:
- in-memory master cannot call this code, it does not populate the `gOpenBatch` which implies not populating the `gInFlightBatch` structure.
- this implies that the only case to call this code is when members list become empty after exhausting each one's retry attempts or failing to replay their operations.

This implies the `else` won't do nothing and the comment is misleading.

src/master/matoclserv.cc
```diff
@@ -211,6 +231,536 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+		std::unique_ptr<kv::ICommitFuture> future;
+		if (gDebugInjectConflicts > 0) {
+			// Fault injection (first attempt): synthesize a retryable conflict.
+			future = std::make_unique<kv::ImmediateCommitFuture>(false, true);
+		} else {
+			future = txn->commitAsync();
+		}
+		future->setReadyCallback(&matoclserv_commit_wakeup, nullptr);
+		gInFlightBatch.emplace(InFlightBatch{std::move(gOpenBatch.ctx), std::move(future),
+		                                     std::move(gOpenBatch.members), 0});
```
dmga44: Consider (very optional) to create a lambda `updateInFlightBatch` similar to these lines to be used here and above when retrying.

src/master/matoclserv.cc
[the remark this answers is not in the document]
dmga44: Though I think applying the change from the AI is out of the scope of this PR, I think it is quite closely related. As you mention the killed-client cleanup this PR adds is reached via the on-commit resolve hook, and the behavior if called on living clients remains, but consider the following scenario:
- a client write tries to create a new chunk.
- the client immediately disconnects.
- the chunk creation fails (enough chunkserver disconnects).
- depending on who replies first (the KV backend or the chunkservers tasked with creating the chunks) the cleanup (`gFSOperations->removeChunkFromFile(context, ctx, inode, chunkId);`) is called.

I think it should be analyzed in `matocl_before_disconnect`.
I think that current in-memory master is not handling this properly and the PR adds a very nice way to handle it for KV backend MDS.

Please reconsider the suggestion from AI and consider adding a more robust handling of the disconnection.

src/master/matoclserv.cc
dmga44, line 2242: Liked the changes in this function!! (`matoclserv_fuse_reserved_inodes`)

dmga44 (commented):
Pass 2.

src/master/filesystem_operations.cc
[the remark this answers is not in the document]
lgsilva3087: You're right, and the reasons I gave were wrong. Reworked:

**Base `acquire` (in-memory master): reverted to `SAUNAFS_ERROR_EINVAL`.** Callers reach `acquire` only when `openFilesSet` doesn't already track the inode (the in-memory master couples acquire with that insert via `matoclserv_insert_open_file`), so a session already in `sessionIds` means the two drifted -- a structural inconsistency that should stay loud. The changelog-replay and failed-release justifications were both wrong, as you said, and are removed.

**KV `acquire`: kept idempotent, but for the real reason.** The group-commit pipeline defers the `openFilesSet` insert (`recordOpenFile`) to commit, so within a batch or on a conflict replay the same `(inode, session)` can re-enter `acquire` while `openFilesSet` is stale and the transaction's read-your-writes already shows the session. Returning EINVAL there would fail a valid second open, so the KV override returns OK; it always (re)writes `NODE_`/`SESS_OPEN_`, and now logs ACQUIRE / bumps the version only when the session was actually absent.

So the contract is intentionally backend-specific -- in-memory surfaces the duplicate, KV tolerates it -- documented in the `acquire()` note. The KV-side change ships as a new commit in the related MDS PR.

src/master/matoclserv.cc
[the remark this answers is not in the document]
lgsilva3087: Agreed, took the suggestion -- thanks for pushing on it.

Confirmed the gap: a delayed write op holds the chunk write-lock until the chunkserver status arrives, but the dispatcher skipped killed entries and dropped that status, and disconnect cleanup only unlocked truncate ops. So a client that disconnects before the create status arrives stranded the write-lock, and a failed create left the speculative chunk dangling -- exactly your scenario. The resolve-hook path you noted handles the ordering where the status is stashed before the kill; the gap was the other ordering (status after the kill), and it affected both the in-memory master and the KV backend.

Fix (new commit): the dispatcher now stashes a killed client's write status instead of dropping it, and disconnect cleanup releases the lock and, on a failed create, drops the chunk -- using the real chunkserver outcome. The commit-resolve hook covers the status-arrives-while-commit-in-flight ordering; the disconnect path covers status-after.

The cleanup commits synchronously in the disconnect handler: it can't go through the group-commit pipeline because the entry is being torn down (a deferred op would reference it after free), so it mirrors the existing truncate unlock there.

One honest limit: lock release is guaranteed, but the failed-create chunk drop only happens when the status arrives before the entry is freed (covers your scenario). A very late chunkserver report would find the lock already released but could leave the chunk reference -- inherent to not deferring teardown indefinitely, and still strictly better than today, where the lock leaks too.

src/master/matoclserv.cc
lgsilva3087: Took all four suggestions, landed in 2f3fa327.

- Merged OpenBatch and InFlightBatch into a single OpBatch with the fields you sketched (ctx, future, members, attempt, hasContext). gOpenBatch is an OpBatch; gInFlightBatch is std::optional<OpBatch>.
- Dropped HeldOp: a held op has the same shape as a BatchMember, so gHeldOps is now std::deque<BatchMember> and the poll_batch drain moves it straight into the join instead of rebuilding it.
- Extracted matoclserv_replay_members_iteration; matoclserv_replay_members is now just the restart loop around it.
- join_open_batch uses that same step: it appends the new member, runs one iteration, and rebuilds via replay_members if the step reports the shared transaction can no longer be used.

dmga44: Nice!

src/master/matoclserv.cc
lgsilva3087: Done. std::bit_width(size) - 1 capped at 4; buckets are now 1 / 2-3 / 4-7 / 8-15 / 16+. Updated the histogram doc and header comment to match, and the profiler line that prints the boundaries.

src/master/matoclserv.cc
lgsilva3087: You're right on both points. The in-memory master finishes ops inline and never populates gOpenBatch, so it never reaches an in-flight batch: the conflict-replay path is KV only, txn is always non-null there, and the else only fired when members was empty, where its loop did nothing. Removed the dead branch and the misleading comment; the condition is now just !members.empty().

Worth adding: that else was not only dead, it was latently unsafe. Had it ever run with non-empty members (the impossible txn-null case it claimed to handle), it would have replied SAUNAFS_STATUS_OK to ops that never committed. So removing it is strictly safer, not just tidier.

src/master/matoclserv.cc
lgsilva3087: Done. Extracted a launchInFlightBatch lambda (build the future, set the callback, move the batch in-flight), used by both the open-batch submit and the conflict-replay retry. The two fault-injection checks collapsed into one (attempt < gDebugInjectConflicts, with submit passing attempt 0), so no behavior change.

src/master/matoclserv.cc
dmga44: Nice.

src/master/matoclserv.cc
```diff
@@ -2012,228 +2620,245 @@ void matoclserv_fuse_setattr(matoclserventry *eptr, const uint8_t *data, uint32_
...
+		                                 (ts->type != FUSE_TRUNCATE_END), ts->lockId, ts->attr,
+		                                 &ts->chunkId);
+
+		// In case of SAUNAFS_ERROR_NOTPOSSIBLE we have to tell the client to write the
+		// chunk before truncating (new client truncating an xor/ec chunk does it itself).
+		if (st == SAUNAFS_ERROR_NOTPOSSIBLE && !isEnd) {
+			uint8_t chunkOperationPending;
+			st = gFSOperations->writeChunk(context, ctx, inode, ts->length / SFSCHUNKSIZE,
+			                               &ts->lockId, &ts->chunkId, &chunkOperationPending,
+			                               &ts->fileLength);
```
dmga44: As we know the calls to `trySetLength` and `writeChunk` can derive to some chunk operations like create, duplicate, truncate and duplicate-truncate. I think this is out of the scope of this PR, but it is quite worth a big deal of analysis how to cope those operations (specially duplicate and duplicate-truncate) with the retry schema implemented for KV backends.

src/master/matoclserv.cc
```diff
@@ -2012,228 +2620,245 @@ void matoclserv_fuse_setattr(matoclserventry *eptr, const uint8_t *data, uint32_
...
+		DelayedChunkOperation *queued = nullptr;  ///< commit-pending entry of the latest run
+		bool tookDelayedPath = false;  ///< onCommit took the delayed path; reply comes via chunk_status
+	};
+	auto ts = std::make_shared<TruncState>();
+	ts->type = isEnd ? FUSE_TRUNCATE_END : FUSE_TRUNCATE;
+	// lockId must survive body replays (mirror write_chunk): a batch-conflict replay of the
+	// non-END duplicate path passes the previously allocated lockId back into writeChunk so it
+	// re-enters its own surviving in-memory lock instead of bouncing off it as LOCKED. For END it
+	// stays the client-supplied reqLockId. So it is initialized ONCE here, not reset per run.
+	ts->lockId = reqLockId;
```
dmga44: I think this line should be moved inside the `runTruncate` function. From our friends:

`Lock leakage in writeChunk's in-memory lock table. ts->lockId is preserved across replays specifically so re-entering writeChunk doesn't bounce off the master's own lock. That works for the master-side lock, but each chunkserver-side DUPTRUNC carries the version it was dispatched with; the chunkserver may end up with locked-but-orphan chunk versions that never get the matching END.`

dmga44 (commented):
A look at `matoclserv_fuse_truncate`.

src/master/matoclserv.h
lgsilva3087: The spec was a local working document I was iterating against; it is not part of the repository, so the reference dangled. Removed it and kept the explanation inline.

src/master/matoclserv.cc
lgsilva3087: The spec was a local working document I was iterating against; it is not part of the repository, so the reference dangled. Removed it and kept the explanation inline.

src/master/matoclserv.cc
lgsilva3087: The spec was a local working document I was iterating against; it is not part of the repository, so the reference dangled. Removed it and kept the explanation inline.

src/master/matoclserv.cc
lgsilva3087: The spec was a local working document I was iterating against; it is not part of the repository, so the reference dangled. Removed it and kept the explanation inline.

src/master/matoclserv.cc
lgsilva3087: Correct. On the KV backend the body may only read, check, and stage backend writes; it must not mutate in-memory MDS structures, because a replay reruns it on a fresh transaction that is not rolled back. writeEnd follows this today (it stages through the backend op and keeps no in-memory side effects of its own). I clarified the contract in the doc comment above the function.

tests/test_suites/ShortSystemTests/test_ec_truncate_atomicity.sh
lgsilva3087: Done, both suggestions. Reworded the comment to attribute the extra time to KV-backend conflict resolution rather than the pipeline, and set the timeout by backend: 3 minutes when METADATA_BACKEND=FDB, the historical 1 minute otherwise. Applied to test_ec and test_xor truncate-atomicity.

src/master/matoclserv.cc
lgsilva3087: Good catch. Moving it inside would reset the lockId on each replay, and writeChunk would then bounce off its own lock with LOCKED, so I kept the init outside. The deeper orphan risk you mention is real though, same root as the 2765 thread and KV-only. Parked it in LS-868 (Conflict resolution for transactions milestone) to handle properly.

src/master/matoclserv.cc
lgsilva3087: Agreed, and out of scope here. Short version: a body can replay on a conflict, but trySetLength/writeChunk also talk to the chunkserver and touch in-memory chunk state outside the transaction, so a rollback does not fully undo them (KV-only; the in-memory master never replays). The single-ref path is guarded; the duplicate/duptrunc path I want to look at more closely, since a replay there may not be harmless. Parked in LS-868 (Conflict resolution for transactions milestone).

dmga44 (approved):
Despite the many suggestions, I actually liked the implementation. I think the ideas are quite cleanly expressed, but some comments are a bit too much and looking at the implementation was better. Anyway, great job!

Question: some operations like `matoclserv_fuse_getreserved` are not changed  like many of the common operation the client sends, is it going to be changed in the future?

As a final suggestion, consider (minor suggestion):
- to include the common `uint8_t status = matoclserv_check_group_cache(eptr, gid);` in the run/body function, to remove a future if and a call to the send/reply function.
- move the update of the session stats a bit up (above the run/body) function when possible.

This way the structure of many functions would look like:
- deserialization
- checks and update stats
- run/body function
- send/reply function
- `matoclserv_submit_op` call.

lgsilva3087:
> Despite the many suggestions, I actually liked the implementation. I think the ideas are quite cleanly expressed, but some comments are a bit too much and looking at the implementation was better. Anyway, great job!
> 
> Question: some operations like `matoclserv_fuse_getreserved` are not changed like many of the common operation the client sends, is it going to be changed in the future?
> 
> As a final suggestion, consider (minor suggestion):
> 
> * to include the common `uint8_t status = matoclserv_check_group_cache(eptr, gid);` in the run/body function, to remove a future if and a call to the send/reply function.
> * move the update of the session stats a bit up (above the run/body) function when possible.
> 
> This way the structure of many functions would look like:
> 
> * deserialization
> * checks and update stats
> * run/body function
> * send/reply function
> * `matoclserv_submit_op` call.

Thanks for the review, it was great.

On the question: read-only ops like matoclserv_fuse_getreserved (gettrash,
getattr, lookup, readdir) stay as they are on purpose. matoclserv_submit_op
batches durable commits, and those handlers open a read-only transaction and
reply inline, so there is nothing to commit.

I like both suggestions. The PR is already long, so I will apply them step by
step in focused follow-ups.

src/common/event_loop.cc
```diff
@@ -41,6 +42,26 @@ ExitingStatus gExitingStatus = ExitingStatus::kRunning;
...
+/// such boundary; before this guard an escaped kv::RetryableTransactionError
+/// (e.g. a transaction_timed_out read while the backend is slow or shutting
+/// down) reached std::terminate and aborted the whole server. The interrupted
+/// handler simply runs again on a later loop iteration.
+template <typename Handler, typename... Args>
+static void eventloop_run_handler(const char *kind, Handler &&handler, Args &&...args) {
+	try {
+		handler(std::forward<Args>(args)...);
+	} catch (const kv::RetryableTransactionError &e) {
+		safs_pretty_syslog(LOG_WARNING,
```
ralcolea: Maybe you can use `safs::log_warn()` instead, to use the modern logging in the new generated code.

src/master/matoclserv.cc
```diff
@@ -211,6 +233,495 @@ std::unordered_map<uint64_t, std::set<WaitEntry, WaitEntryCmp>> gWaitForUnlockMa
...
+
+MatoclBatchStats matoclserv_get_batch_stats() { return gBatchStats; }
+
+/// Records a committed batch of `size` members in the stats (size-histogram bucketed by
+/// power of two as 1, 2-3, 4-7, 8-15, 16+). No-op unless accounting is enabled.
+static void matoclserv_record_committed_batch(size_t size) {
+	if (!gBatchStatsEnabled || size == 0) { return; }
+	gBatchStats.committedBatches++;
+	gBatchStats.committedOps += size;
+	if (size > gBatchStats.maxBatchSize) { gBatchStats.maxBatchSize = size; }
```
ralcolea: Consider using `std::max()` function.

src/master/matoclserv.cc
```diff
@@ -487,121 +998,190 @@ uint8_t matoclserv_fuse_write_chunk_respond(matoclserventry *eptr,
...
+}
+
+/// Handles a completed chunkserver operation for a delayed op already removed from its
+/// client's queue: builds and sends the client reply, committing any metadata
+/// finalization. The truncate finalize commits while concurrent writers' async commits
+/// on the same inode are in flight, so a retryable conflict replays on a fresh
+/// transaction (it was the residual failure of the {ec,xor}_truncate_atomicity tests
+/// under async commit); the commit itself is deferred, with the reply in the commit
+/// continuation, so finalization never blocks the event loop on backend durability.
+static void matoclserv_process_chunk_status(matoclserventry *eptr,
```
ralcolea, line 1047: Very nice logic separation! 💪

src/master/matoclserv.cc
```diff
@@ -2012,228 +2622,245 @@ void matoclserv_fuse_setattr(matoclserventry *eptr, const uint8_t *data, uint32_
...
+		uint32_t type = 0;
+		uint64_t chunkId = 0;
+		uint32_t lockId = 0;
+		uint64_t length = 0;
+		uint64_t fileLength = 0;
+		Attributes attr{};
+		DelayedChunkOperation *queued = nullptr;  ///< commit-pending entry of the latest run
+		bool tookDelayedPath = false;  ///< onCommit took the delayed path; reply comes via chunk_status
+	};
+	auto ts = std::make_shared<TruncState>();
```
ralcolea: Using `ts` to reference `TruncState` can be confusing because in other parts of the codebase `ts` is used to reference timestamp values. I think is worth to use a longer name, like `truncState` or similar.

src/master/matoclserv.cc
```diff
@@ -5435,10 +6148,22 @@ void matocl_before_disconnect(matoclserventry *eptr) {
...
 
-	// unlock locked chunks
+	// Release locks held by delayed chunk ops the client never finished: a killed client sends
+	// no end-of-operation request, so the lock must be released here or it strands and blocks
+	// future writers. For a write op whose status arrived after the client was killed, a failed
+	// create also drops the speculative chunk so it is not left dangling in the file.
 	for (const auto &operation : eptr->delayedChunkOperations) {
 		if (operation->type == FUSE_TRUNCATE) {
 			gFSOperations->endSetLength(fsOpContext, operation->chunkId);
+		} else if (operation->type == FUSE_WRITE) {
```
ralcolea, line 6160: Very nice improvement! 💪

src/master/matoclserv.cc
```diff
@@ -2989,48 +3628,71 @@ void matoclserv_fuse_open(matoclserventry *eptr, const uint8_t *data, uint32_t l
...
+		if (st == SAUNAFS_STATUS_OK) {
+			memcpy(ptr, replyAttr.data(), replyAttr.size());
+		} else {
+			put8bit(&ptr, st);
 		}
-		ptr = matoclserv_createpacket(eptr,MATOCL_FUSE_OPEN, sizeof(msgid) + attr.size());
-	} else {
-		ptr = matoclserv_createpacket(eptr,MATOCL_FUSE_OPEN, sizeof(msgid) + sizeof(status));
-	}
+		if (eptr->sessionData) { eptr->sessionData->currHourOperationsStats[13]++; }
```
ralcolea: `currHourOperationsStats[13]++` now lives inside `sendOpenReply()`.
This can affect the statistics without actually running the open operation, for instance, when `matoclserv_check_group_cache(eptr, gid)` returns a status different than `SAUNAFS_STATUS_OK`.

ralcolea (approved):
Great job @lgsilva3087 for a very complex PR!! 💪🔥🚀 
I shared some minor suggestions for your consideration, but they are not stoppers for merging!

src/common/event_loop.cc
lgsilva3087: Done, switched this to `safs::log_warn()`. I kept the format string as a single literal to match the local clang-format style.

src/master/matoclserv.cc
lgsilva3087: Done, replaced the manual max update with `std::max<uint64_t>()`.

src/master/matoclserv.cc
lgsilva3087: Done, renamed the shared truncate state from `ts` to `truncState` through the replay, reply, and onCommit paths.

src/master/matoclserv.cc
lgsilva3087: Good catch. Moved `currHourOperationsStats[13]++` out of `sendOpenReply()`. It now runs from the submit-op finalization hook, so group-cache precheck failures are not counted, and held ops dropped before execution are not counted either.

ralcolea (approved):
Great job @lgsilva3087! 💪🔥🚀

PR #895, opened 2026-06-29
feat: single-commit FUSE_CREATE RPC

lgsilva3087:
Creating a file drives two client RPCs, FUSE_MKNOD then FUSE_OPEN, so a create costs two round trips to the master. On the KV backend each RPC also runs its own metadata transaction with its own durable commit, so an empty file pays two durable commits, and that commit is the dominant per-op cost capping file-create throughput.

Add a fused SAU_CLTOMA_FUSE_CREATE / SAU_MATOCL_FUSE_CREATE RPC that does mknod + acquire (open registration) in one transaction, one commit. The mount client sends it from the create path; matoclserv_fuse_create runs a replayable op body through the group commit pipeline and applies the in-memory open-file record once on commit success, so a retry replays the persistent acquire cleanly.

Master behavior is preserved. The traditional in-memory Master runs the same handler with no transaction to roll back, so acquire and the open-file record are coupled in the body via matoclserv_insert_open_file (mirroring matoclserv_fuse_open), and a later openCheck failure cannot orphan an untracked acquire. The emitted changelog is unchanged: the same CREATE then ACQUIRE entries, in the same order.

This is not a breaking change. The RPC is version gated on kFirstVersionWithFusedCreate (5.11.0): an older client keeps using the FUSE_MKNOD + FUSE_OPEN pair, and an older master never receives the new packet. No existing packet or on-disk format is altered.

Closes LS-891

### DIFF
--- a/src/master/matoclserv.cc
+++ b/src/master/matoclserv.cc
@@ -3084,6 +3084,101 @@ void matoclserv_fuse_mknod(matoclserventry *eptr, PacketHeader header, const uin
 	});
 }
 
+// Fused create+open: mknod a file and acquire it for the calling session in a single
+// transaction, so an empty-file create costs one durable commit instead of two (the
+// mknod + open round-trip pair). Version-gated on the client; old clients still send
+// the separate FUSE_MKNOD and FUSE_OPEN.
+void matoclserv_fuse_create(matoclserventry *eptr, PacketHeader header, const uint8_t *data) {
+	uint32_t messageId, uid, gid;
+	inode_t parentInode;
+	LegacyString<uint8_t> name;
+	uint16_t mode, umask;
+	uint8_t flags;
+
+	if (header.type == SAU_CLTOMA_FUSE_CREATE) {
+		cltoma::fuseCreate::deserialize(data, header.length, messageId, parentInode, name, mode,
+		                                umask, uid, gid, flags);
+	} else {
+		throw IncorrectDeserializationException(
+				"Unknown packet type for matoclserv_fuse_create: " + std::to_string(header.type));
+	}
+
+	// Reply data shared between the op body and the reply continuation; a retry
+	// refreshes it through the same struct.
+	auto replyData = std::make_shared<std::pair<inode_t, Attributes>>();
+	uint8_t status = matoclserv_check_group_cache(eptr, gid);
+
+	// Replayable op body: mknod + acquire the open file + openCheck, all in one
+	// transaction. On the KV backend the in-memory open-file record (Session::openFilesSet)
+	// is deliberately NOT done here; it is applied once on commit success (below) so a
+	// retry replays the persistent acquire on a fresh transaction without the already-open
+	// early-return skipping it. On the in-memory Master there is no transaction to roll
+	// back, so acquire+record are coupled here (matoclserv_insert_open_file): a later
+	// openCheck failure must not orphan an untracked acquire that disconnect cleanup would
+	// never release (mirrors matoclserv_fuse_open).
+	OpReplay runCreate = [eptr, uid, gid, parentInode, name, mode, umask, flags,
+	                      replyData](FilesystemOperationContext &ctx) -> uint8_t {
+		FsContext context = matoclserv_get_context(eptr, uid, gid);
+		uint8_t opStatus =
+		    gFSOperations->mknod(context, ctx, parentInode, HString(name), FSNodeType::kFile, mode,
+		                         umask, 0, &replyData->first, replyData->second);
+		if (opStatus == SAUNAFS_STATUS_OK) {
+			opStatus = ctx.getReadWriteTransaction() == nullptr
+			               ? matoclserv_insert_open_file(ctx, eptr->sessionData, replyData->first)
+			               : matoclserv_acquire_open_file_persist(ctx, eptr->sessionData,
+			                                                      replyData->first);
+		}
+		if (opStatus == SAUNAFS_STATUS_OK) {
+			opStatus = gFSOperations->openCheck(context, ctx, replyData->first,
+			                                    flags | AFTER_CREATE, replyData->second);
+		}
+		return opStatus;
+	};
+
+	if (eptr->sessionData) {
+		eptr->sessionData->currHourOperationsStats[8]++;
+	}
+
+	// Builds the reply (inode + attr) applying the same data-cache adjustment the
+	// FUSE_OPEN reply uses, since this op opens the file as well as creating it.
+	auto sendCreateReply = [eptr, messageId](uint8_t replyStatus, inode_t ino,
+	                                         Attributes replyAttr) {
+		MessageBuffer reply;
+		if (replyStatus == SAUNAFS_STATUS_OK) {
+			// When the data cache manager cannot keep this open coherent (dcm_open == 0), clear
+			// the ALLOWDATACACHE attribute bit so the client does not cache this file's data.
+			if (dcm_open(ino, eptr->sessionData->sessionId) == 0) {
+				replyAttr[1] &= (0xFF ^ (MATTR_ALLOWDATACACHE << 4));
+			}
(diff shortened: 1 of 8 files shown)


src/mount/mastercomm.cc
```diff
@@ -2290,6 +2290,45 @@ uint8_t fs_mknod(inode_t parent, uint8_t nleng, const uint8_t *name, uint8_t typ
...
+			matocl::fuseCreate::deserialize(message, messageId, status);
+			if (status == SAUNAFS_STATUS_OK) {
+				fs_got_inconsistent("SAU_MATOCL_FUSE_CREATE", message.size(),
+						"version 0 and SAUNAFS_STATUS_OK");
+				return SAUNAFS_ERROR_IO;
+			}
+			return status;
+		} else if (packetVersion == matocl::fuseCreate::kResponsePacketVersion) {
+			matocl::fuseCreate::deserialize(message, messageId, inode, attr);
+			return SAUNAFS_STATUS_OK;
```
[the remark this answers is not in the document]
lgsilva3087: Fixed. `fs_create` now calls `fs_inc_acnt(inode)` on the `kResponsePacketVersion` branch, mirroring `fs_opencheck`. The fused create's acquire is tracked in `acquiredFiles` and reported through the periodic `CLTOMA_FUSE_RESERVED_INODES` updates, so master-side open-file state survives reconnect exactly as it did with the separate FUSE_MKNOD + FUSE_OPEN pair. Good catch.

src/mount/mastercomm.cc
```diff
@@ -2290,6 +2290,45 @@ uint8_t fs_mknod(inode_t parent, uint8_t nleng, const uint8_t *name, uint8_t typ
...
+		} else {
```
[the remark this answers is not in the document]
lgsilva3087: Same fix as the thread above: `fs_create`'s success path now calls `fs_inc_acnt(inode)`, so the fused-create acquire is counted and reserved identically to the previous `fs_opencheck` behavior. No accounting divergence.

src/master/matoclserv.cc
```diff
@@ -3084,6 +3084,91 @@ void matoclserv_fuse_mknod(matoclserventry *eptr, PacketHeader header, const uin
...
+	}
+
+	matoclserv_submit_op(eptr, std::move(runCreate),
+	                     [eptr, sendCreateReply, replyData](uint8_t commitStatus) {
+		                     if (commitStatus == SAUNAFS_STATUS_OK) {
+			                     // Durable: record the open file in memory, exactly once.
+			                     matoclserv_record_open_file(eptr->sessionData, replyData->first);
+		                     }
+		                     sendCreateReply(commitStatus, replyData->first, replyData->second);
+	                     });
```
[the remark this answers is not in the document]
lgsilva3087: Fixed. `matoclserv_record_open_file` moved from the `finish` callback to the `onCommit` hook, which runs even when the client is in `ClientConnectionMode::KILL` -- matching `matoclserv_fuse_open`. A successful backend acquire is now always recorded in `Session::openFilesSet`, so disconnect cleanup can release it. The persistent acquire itself (`matoclserv_acquire_open_file_persist`, or `matoclserv_insert_open_file` on the in-memory backend) stays in the replayable op body; only the in-memory bookkeeping moved to commit-success.

src/master/matoclserv.cc
```diff
@@ -3084,6 +3084,98 @@ void matoclserv_fuse_mknod(matoclserventry *eptr, PacketHeader header, const uin
...
+		}
+		if (st == SAUNAFS_STATUS_OK) {
+			st = gFSOperations->openCheck(context, ctx, replyData->first, flags | AFTER_CREATE,
+			                              replyData->second);
+		}
+		return st;
+	};
+
+	if (eptr->sessionData) {
+		eptr->sessionData->currHourOperationsStats[8]++;
```
antuan96314, line 3139:
```suggestion
		eptr->sessionData->currHourOperationsStats[CONSTANT_MEANINGFULL_NAME]++;
```Consider removing magic number.

**Review by antuan96314 — APPROVED** [review]

File: src/master/matoclserv.cc (hunk 4 above)
Comment (lgsilva3087) [reply src/master/matoclserv.cc:3139]: Good catch on the readability. I would prefer to keep it out of this PR though. Every `currHourOperationsStats` index is a bare literal today (0 to 15 across `matoclserv.cc`), including `matoclserv_fuse_mknod`'s `[8]` right above this one. `8` is the create/mknod counter, which the fused create reuses on purpose since it is a create. Naming only this site would diverge from its siblings and read as if the other 15 indices mean nothing. A possible fix is an enum covering all 16 indices applied to every call site, which I would rather do in a follow up than fold a cross cutting cleanup into this feature commit.

Comment (lgsilva3087) [reply src/master/matoclserv.cc:3139]: https://linear.app/leil/issue/LS-892/refactormaster-name-session-op-stat-indices

File: src/master/matoclserv.cc [hunk 5]
```diff
@@ -3084,6 +3084,98 @@ void matoclserv_fuse_mknod(matoclserventry *eptr, PacketHeader header, const uin
...
+	// is deliberately NOT done here; it is applied once on commit success (below) so a
+	// retry replays the persistent acquire on a fresh transaction without the already-open
+	// early-return skipping it. On the in-memory Master there is no transaction to roll
+	// back, so acquire+record are coupled here (matoclserv_insert_open_file): a later
+	// openCheck failure must not orphan an untracked acquire that disconnect cleanup would
+	// never release (mirrors matoclserv_fuse_open).
+	OpReplay runCreate = [eptr, uid, gid, parentInode, name, mode, umask, flags,
+	                      replyData](FilesystemOperationContext &ctx) -> uint8_t {
+		FsContext context = matoclserv_get_context(eptr, uid, gid);
+		uint8_t st = gFSOperations->mknod(context, ctx, parentInode, HString(name),
```
rolysr: Consider renaming this `st` to `status`

src/master/matoclserv.cc
```diff
@@ -3084,6 +3084,98 @@ void matoclserv_fuse_mknod(matoclserventry *eptr, PacketHeader header, const uin
...
+		eptr->sessionData->currHourOperationsStats[8]++;
+	}
+
+	// Builds the reply (inode + attr) applying the same data-cache adjustment the
+	// FUSE_OPEN reply uses, since this op opens the file as well as creating it.
+	auto sendCreateReply = [eptr, messageId](uint8_t st, inode_t ino, Attributes replyAttr) {
+		MessageBuffer reply;
+		if (st == SAUNAFS_STATUS_OK) {
+			if (dcm_open(ino, eptr->sessionData->sessionId) == 0) {
+				replyAttr[1] &= (0xFF ^ (MATTR_ALLOWDATACACHE << 4));
```
rolysr, line 3151: Consider adding a comment on the goal of this operation on this line for better understanding at once while analyzing this code.

rolysr (approved):
LGTM. Please consider my minor suggestions.

src/master/matoclserv.cc
lgsilva3087: Done. Renamed it to `opStatus` in `runCreate`, and the `sendCreateReply` lambda parameter to `replyStatus` (matching the sibling `sendMkdirReply`). I avoided plain `status` on purpose: this function already has an outer `status` (the group-cache check result), so reusing `status` inside the lambdas would shadow it (`-Wshadow-uncaptured-local`). Left the sibling `sendOpenReply`'s `st` untouched, as it predates this change and is outside this PR's scope.

src/master/matoclserv.cc
lgsilva3087: Added a comment above the `dcm_open` check: when the data cache manager cannot keep this open coherent (`dcm_open == 0`), we clear the `ALLOWDATACACHE` attribute bit so the client does not cache this file's data, mirroring `FUSE_OPEN`.

PR #901, opened 2026-07-02
Rpm build from checkout ref

rolysr:
This PR updates the rpm folder for related Fedora packaging build, introducing fixes needed due to rebranding plus adapting build.sh and review.sh scripts to work on a Jenkins pipeline.

Related to: LS-844

### DIFF
--- a/rpm/build.sh
+++ b/rpm/build.sh
@@ -1,23 +1,67 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # Builds the RPM package locally
 
 FILTERS=filters.toml
+SOURCE_REF=${1:-HEAD}
 
 read NAME VERSION RELEASE < <(rpmspec -q --qf "%{NAME} %{VERSION} %{RELEASE}\n" *.spec | head -1)
+SOURCE_TARBALL=v${VERSION}.tar.gz
+REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null || true)
+
+resolve_source_ref() {
+    local ref=$1
+    local candidate
+
+    for candidate in \
+        "${ref}" \
+        "origin/${ref}" \
+        "refs/heads/${ref}" \
+        "refs/remotes/origin/${ref}" \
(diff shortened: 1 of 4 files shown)


rolysr:
@gemini-code-assist review

svaroq:
LGTM, but consider the AI's suggestions, some of them are quite relevant.

rolysr:
@gemini-code-assist review

rolysr:
@gemini-code-assist review

svaroq (approved):

rpm/build.sh
```diff
@@ -1,29 +1,87 @@
...
+    done
+
+    echo "ERROR: Git reference not found: ${ref}" >&2
+    return 1
+}
+
+create_source_tarball() {
+    local resolved_ref short_ref
+
+    if [ -z "${REPO_ROOT}" ]; then
```
lgsilva3087, line 48: Consider (non-blocking):

If `REPO_ROOT` is empty this silently falls back to `spectool` and packages the upstream release tarball, ignoring an explicitly passed `SOURCE_REF`. This can happen in a real checkout: with `sudo ./build.sh` (as the README suggests), `git rev-parse --show-toplevel` can fail with "dubious ownership" and its stderr is discarded, so you'd package the wrong source without noticing.

Suggestion: fail instead of falling back when a ref was requested:

```bash
if [ -z "${REPO_ROOT}" ]; then
    if [ "${SOURCE_REF}" != "HEAD" ]; then
        echo "ERROR: ${SOURCE_REF} requested but not inside a Git checkout" >&2
        exit 1
    fi
    echo "WARNING: not inside a Git checkout; downloading source tarball from spec." >&2
    spectool -g "${NAME}.spec"
    return
fi
```

lgsilva3087 (approved):
Great. LGTM. See my optional hardening.

rpm/build.sh
rolysr: Done.

PR #903, opened 2026-07-02
fix(cs): avoid moved-from disk access on removal

lgsilva3087:
hddCheckDisks moved the disk's unique_ptr out of gDisks into gDisksToBeDeletedWithPendingChunks and then dereferenced the same vector slot to unregister the disk paths from ChunkTrashManager. The slot holds nullptr after the move, so when a disk was removed from the config while chunks on it were still locked by workers (the only case where the pending chunks list is non empty), the disks thread crashed with a null dereference and took down the whole chunkserver.

Observed in CI as a flaky failure of
LongSystemTests.test_concurrent_replications: the chunkserver that lost a disk segfaulted right after logging "Disk ... successfully removed", never re-registered with the master, and the test timed out waiting for a file part to appear on it.

Unregister the trash manager paths before moving the pointer, and keep the move as the last use of the disk.

Related to: LS-879


dmga44 (approved):
Nice fix!

rolysr (approved):

PR #906, opened 2026-07-03
test: reduce flakiness of io limiting tests

lgsilva3087:
When the CI machine is overloaded, the io limiting tests fail on wall clock overshoots (observed between +2% and +457%) although the limiter is healthy: the old asserts were symmetric, so a read stalled by the scheduler counted as a limiter error.

Only the fast side is a correctness property of a rate limiter, so keep it strict and unscaled. Bound the slow side scaled by the total timeout multiplier, consistent with how test timeouts already model slow machines, and in the transition test bound it on the aggregate of all reads so a single stalled read cannot fail the run alone. Replace the local rescale_value in test_io_limits, which ignored the machine multiplier, with a slow tolerance that has a relative component, and drop the valgrind branches since the multiplier already covers them.

Verified locally: pinning the whole cluster plus busy loops to a few cores reproduces the CI failures on all four tests before the change and passes afterwards, both with a declared timeout multiplier and unpinned; a limiter letting IO through too fast still fails.

Related to: LS-879

### DIFF
--- a/tests/test_suites/ShortSystemTests/test_global_io_limiting_many_clients.sh
+++ b/tests/test_suites/ShortSystemTests/test_global_io_limiting_many_clients.sh
@@ -6,14 +6,16 @@ echo "limit unclassified 1024" > "$iolimits"
 
 # number of mounts
 N=5
-# tolerated relative error (in %)
+# Tolerated relative error (in %) for the strict fast side.
 E=11
 if is_windows_system; then
 	E=15
 fi
-if valgrind_enabled; then
-	E=$((5 * E))
-fi
+# Slow-side tolerance (in %): loose, reads can stall on busy CI machines.
+E_SLOW=$((50 * $(timeout_get_total_multiplier)))
+# Fast-side absolute grace (in ns) for the limiter's token bucket bursts.
+# Larger here: the sequential reader start-up gives the bucket more time to refill.
+FAST_GRACE=$((500 * 1000 * 1000))
 
 CHUNKSERVERS=3 \
 	MOUNTS=$N \
@@ -48,6 +50,7 @@ start_readers | while [ $(head -c1M |wc -c) = $((1024 * 1024)) ]; do
 	abserr=$((time - expected))
 	relerr=$((100 * abserr / expected))
 	echo $expected $time $abserr $relerr
-	assert_near 0 $relerr $E
+	assert_less_or_equal $((expected * (100 - E) / 100 - FAST_GRACE)) ${time}
+	assert_less_or_equal ${time} $((expected * (100 + E_SLOW) / 100))
 	i=$((i+1))
 done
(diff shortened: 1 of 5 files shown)


tests/test_suites/ShortSystemTests/test_io_limits.sh
```diff
@@ -8,6 +8,18 @@ echo "limit unclassified 1024" > "$iolimits"
...
 echo "limit unclassified 2048" > "$iolimits2"
 
+# Check the wall-clock time of a rate-limited operation. The fast side stays strict: an
+# operation finishing faster than the limit allows means the limiter let too much IO
+# through. The slow side gets a relative component and is scaled like test timeouts are,
+# because on a loaded machine any operation can stall for reasons unrelated to the limiter.
+assert_limited_io_time() {
+	local expected_time_ms=$1
+	local actual_time_ms=$2
+	local slow_tolerance_ms=$(( (250 + expected_time_ms * 15 / 100) * $(timeout_get_total_multiplier) ))
```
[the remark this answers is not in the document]
lgsilva3087: The failure mode SC2155 guards against does not really exist here: `timeout_get_total_multiplier` reads the framework's own multiplier file, and if that machinery were broken every test would already fail loudly through the timeout system itself. These scripts also run under the harness's `set -eu` with no ERR trap in play. Keeping the one-line form matches the prevailing style of the test framework (`assert.sh` and other tests use `local var=$(...)` throughout), so splitting this single occurrence would just make it inconsistent.

tests/test_suites/ShortSystemTests/test_global_io_limiting_many_clients.sh
```diff
@@ -6,14 +6,15 @@ echo "limit unclassified 1024" > "$iolimits"
 
 # number of mounts
 N=5
-# tolerated relative error (in %)
+# Tolerated relative error (in %). The fast side stays strict: data arriving faster than the
+# configured rate means the limiter let too much IO through. The slow side is scaled like
+# test timeouts are, because on a loaded machine reads can stall for reasons unrelated to
+# the limiter. The measurements are cumulative, so a single stall also fades as i grows.
```
dmga44: Consider summarizing the comment (applies for other tests):
```
# Tolerated relative error (in %). It is later scaled considering the machine
# multiplier for testing under heavy workloads.
```

dmga44 (approved):
Nice, please check my small suggestion.

tests/test_suites/ShortSystemTests/test_global_io_limiting_many_clients.sh
lgsilva3087: Good idea, applied to the three global tests, and shortened the matching helper comment in test_io_limits the same way.

rolysr (approved):

PR #908, opened 2026-07-06
refactor(master): add EDGEHASH_ key prefix

lgsilva3087:
Declare EDGEHASH_<ParentId><NameHash><Name> so KV backends can store a secondary edge index ordered by the readdir name-hash cursor. A paged readdir can then resume with a first-greater-or-equal range seek instead of matching the exact cursor entry, which keeps listings complete when the cursor entry is deleted between batches and avoids rescanning the directory on every batch. The literal must not start with EDGE_ so primary edge scans cannot swallow the index; rebuildable from the EDGE section, so it carries no section version.

Also normalize two non-ASCII dashes in the FLCK_ doc comment.

Related to: LS-903


ralcolea (approved):
LGTM! 👍

dmga44 (approved):
LGTM.

PR #909, opened 2026-07-08
refactor(master): KV parity for changelog, locks, and repair

lgsilva3087:
Master-side changes so the FDB metadata backend increases functional
parity with the in-memory master. Every new path gates on an active
read-write transaction, so the in-memory master is unaffected: it
never defers and never retries.

Changelog:
- Defer changelog sinks (file, signal, metalogger, notifier) until
  after commit and drain them at a single confirmed-commit choke
  point, so retried KV commits no longer publish aborted or duplicate
  entries.
- Route every publication, deferred and inline, through one version
  sequencer so the emitted stream stays monotonic on a single MDS.

Locks:
- Replay FLOCK and SETLK through the retry harness so normal
  same-inode contention no longer surfaces as EIO, sharing one lock-op
  helper and clearing the applied set on failure. WAITING preserved.
- Route admin unlock and pending-lock interrupt through the same
  harness.

Repair and chunk locks:
- Persist repair edits to the file node via updateNode, otherwise
  erased or repaired chunk-table edits are lost after a restart (the
  in-memory master keeps the node itself and masks this).
- Expose chunk lockid and lockedto accessors so the KV backend can
  round-trip write-lock state without reaching into chunk internals.

Test:
- Track the trash hot-reload chunkserver count in one shell variable
  instead of grepping the test's own source, which broke when sourced
  from the FDB wrapper.

Known limitation: some direct-commit paths (disconnect cleanup,
reserved inodes, repair) still publish inline before durability, and
the sequencer is single-MDS; multi-MDS ordering is future work.

Related to: LS-906 LS-907

### DIFF
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -108,6 +108,12 @@ void FilesystemOperationsBase::changeLog(
 		return;
 	}
 
+	if (fsOpContext.hasReadWriteTransaction()) {
+		// Direct KV commits still publish inline; sequence them with deferred publications
+		// so duplicate staged versions do not reach changelog consumers.
+		version = matoclserv_sequence_published_changelog_version(version);
+	}
+
 	changelog(version, entry);
 
 	if (!getChangelogSignal().empty()) {
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -114,14 +114,7 @@ void FilesystemOperationsBase::changeLog(
 		version = matoclserv_sequence_published_changelog_version(version);
 	}
 
-	changelog(version, entry);
-
-	if (!getChangelogSignal().empty()) {
-		getChangelogSignal().emit({.version = version, .entry = entry});
-	}
-
-	matomlserv_broadcast_logstring(version, (uint8_t *)entry, timeStampLength + entryLength);
-	matontserv_broadcast_message(version, std::string_view(entry, timeStampLength + entryLength));
+	matoclserv_emit_changelog_sinks(version, entry, timeStampLength + entryLength);
 #endif
 }
 
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -100,6 +100,14 @@ void FilesystemOperationsBase::changeLog(
 	}
 
 	uint64_t version = increaseMetadataVersion(fsOpContext);
+
+	if (fsOpContext.deferChangelog()) {
+		// Defer every sink (local log, signal, broadcasts) so replayed attempts publish nothing.
+		fsOpContext.appendDeferredChangelogEntry(version,
+		                                         std::string(entry, timeStampLength + entryLength));
+		return;
+	}
+
 	changelog(version, entry);
 
 	if (!getChangelogSignal().empty()) {
--- a/src/master/filesystem_operations.cc
+++ b/src/master/filesystem_operations.cc
@@ -3278,6 +3278,11 @@ uint8_t FilesystemOperationsBase::repair(const FsContext &context, inode_t inode
 			(*notchanged)++;
 		}
 	}
+	if (*erased > 0 || *repaired > 0) {
+		// Chunk-table and mtime edits only touch the materialized node on KV.
+		// Persist them; in-memory backend already stores the node itself.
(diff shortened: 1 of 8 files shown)


rolysr (approved):
LGTM.

src/master/filesystem_operations.cc
```diff
@@ -100,6 +100,20 @@ void FilesystemOperationsBase::changeLog(
 	}
 
 	uint64_t version = increaseMetadataVersion(fsOpContext);
+
+	if (fsOpContext.deferChangelog()) {
+		// Defer every sink (local log, signal, broadcasts) so replayed attempts publish nothing.
+		fsOpContext.appendDeferredChangelogEntry(version,
+		                                         std::string(entry, timeStampLength + entryLength));
+		return;
+	}
```
dmga44, line 109: Consider some operation was replayed a number of times, wouldn't these lines enqueue several changelog entries of the same operation?

src/master/filesystem_operations.cc
```diff
@@ -100,6 +100,20 @@ void FilesystemOperationsBase::changeLog(
 	}
 
 	uint64_t version = increaseMetadataVersion(fsOpContext);
```
dmga44, line 102: Consider if the following seems right to you:
- do not increase the version of the metadara here, but at the moment the actual changelogs are issued (above `changelog` call here and in `matoclserv_publish_deferred_changelog`).
- this will imply keeping track of `gLastPublishedChangelogVersion` is no longer needed, the metadata version matches it. Also no need of keeping track of the version in the deferred changelog entry.

src/master/filesystem_operations.cc
dmga44: Consider making a function out of these lines , including `uint64_t version = increaseMetadataVersion(fsOpContext);` if applying the suggestion above, to use it also in `matoclserv_publish_deferred_changelog` and reduce code duplication.

src/master/matoclserv.cc
```diff
@@ -5338,6 +5368,31 @@ static void matoclserv_lock_wake_up(std::vector<FileLocks::Owner> &owners, safs_
...
+	if (lockStatus != SAUNAFS_STATUS_OK && lockStatus != SAUNAFS_ERROR_WAITING) {
+		applied.clear();  // op error, nothing committed: never wake owners for it
+		return lockStatus;
+	}
+	if (commitStatus != SAUNAFS_STATUS_OK) {  // conflict retries exhausted
+		applied.clear();
+		return SAUNAFS_ERROR_IO;
+	}
+	return lockStatus;
+}
```
dmga44, line 5400: Some lines in the middle will do some help here :)

dmga44 (approved):
Nice job, @lgsilva3087 .

Please check my relatively minor and non-blocking suggestions.

src/master/filesystem_operations.cc
lgsilva3087: Good question. Each retry or replay creates a fresh FilesystemOperationContext. Deferred entries are owned by that context, so a failed attempt is destroyed with its buffer. Only `matoclserv_commit_confirmed()` drains the context whose transaction committed. A replay therefore produces one published entry per `changelog()` call, not one per attempt.

src/master/filesystem_operations.cc
lgsilva3087: I considered this, but kept the current split. `increaseMetadataVersion()` uses a snapshot read plus atomicAdd() in the same KV transaction as the metadata mutation, while publication happens only after that transaction is known durable. Moving the increment to publication would require a separate durable transaction or outbox, decouple metadata version from the mutation, and still leave direct inline KV publications to sequence. The sequencer is therefore deliberate: it turns colliding staged values into the monotonic published stream, while metadata version remains the mutation counter.

src/master/filesystem_operations.cc
lgsilva3087: Implemented the non-version part of this suggestion. `matoclserv_emit_changelog_sinks()` now owns the common changelog file, signal, metalogger, and notifier fanout. I intentionally kept version allocation outside the helper for the transaction-bound reason above.

src/master/matoclserv.cc
lgsilva3087: Applied. I added blank lines between the retry result and each outcome branch.

PR #913, opened 2026-07-10
fix(rpm): build current workflow ref

rolysr:
GH Actions used fedora:latest without git installed before
checkout, so actions/checkout fell back to a git-less tarball
download, leaving build.sh to fetch the stale published release
tarball instead of this branch's source.

Install git before checkout, trust the workspace ownership, and
switch to fedora:rawhide so the host's %dist matches what
fedpkg mockbuild already targets by default.

Also make build.sh discover the real mock results directory
instead of trusting a precomputed release, and fail loudly when
none is found, since rpmlint exits 0 on an empty glob and was
reporting false success.

Related to: LS-957

### DIFF
--- a/.github/workflows/fedora-rpm-build-review.yml
+++ b/.github/workflows/fedora-rpm-build-review.yml
@@ -2,7 +2,7 @@ name: Fedora RPM build and review
 
 on:
   push:
-    branches: [dev]
+  workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


lgsilva3087 (approved):
LGTM

svaroq (approved):
LGTM, it makes rpmlint validation more reliable! 👍

PR #917, opened 2026-07-14
feat(kv): add getSnapshotAsync and addReadConflictKey

lgsilva3087:
Transactions could either read with conflict tracking or write blindly, with nothing in between. getSnapshotAsync gives pipelined advisory reads that stay out of the conflict footprint; addReadConflictKey declares a key as read, so a blind write aborts on a concurrent commit instead of silently overwriting it.

Related to: LS-949


rolysr (approved):
LGTM.

ralcolea (approved):
Great work @lgsilva3087! 💪 🔥 🚀

PR #918, opened 2026-07-14
chore: ignore FDBTests* test suite symlinks

lgsilva3087:
create-softlinks.sh generates the FDBTests suites as symlinks. Ignore them with a glob so future FDBTests* variants (e.g. FDBTestsLong) are covered without editing .gitignore again.

Related to: LS-972


rolysr (approved):

ralcolea (approved):
LGTM! 👍

PR #922, opened 2026-07-17
refactor(cs): add compression hooks

dmga44:
Add the common-interface hooks the zonefs_disk plugin needs for transparent per-chunk Zstd compression on SMR drives (feat-data-compression, companion zonefs_disk PR):

- Replace IChunk::isDirty() with a releaseCachedResources() no-op hook. The only real isDirty() implementation lived in the zonefs plugin, which now tracks dirtiness at zone level instead; CmrChunk drops its trivial override. The new hook lets a chunk free memory kept only to speed up I/O (e.g. digested compression dictionaries) when it leaves the open-chunks cache, and OpenChunk calls it while closing descriptors so idle chunks do not pin dictionary memory.

- Add IDisk::applyNewChunkFormat(), called from the new-chunk creation path before the header is sized and serialized, so a plugin can stamp its on-disk format (e.g. compressed V2) on brand new chunks only - duplicates and replicas keep the source format.

- Add a helgrind suppression for a glibc TLS-teardown false positive at shutdown, triggered by the plugin's thread_local Zstd contexts: pthread_join() in JobPool::stop() trims glibc's cached dead-thread stacks, and with --free-is-write=yes that free() looks like a race against the dead threads' own thread_local destructors.

### DIFF
--- a/src/chunkserver/chunkserver-common/chunk_interface.h
+++ b/src/chunkserver/chunkserver-common/chunk_interface.h
@@ -165,9 +165,11 @@ public:
 	/// truncating a file.
 	virtual void shrinkToBlocks(uint16_t newBlocks) = 0;
 
-	/// Returns true if the Chunk is in an state considered as dirty.
-	/// For example the Chunk is fragmented and needs defragmentation.
-	virtual bool isDirty() = 0;
+	/// Releases memory the chunk caches only to speed up I/O (e.g. per-chunk
+	/// compression dictionaries). Called when the chunk leaves the open-chunks
+	/// cache (its descriptors are being closed), with the chunk locked, so the
+	/// caches can be rebuilt lazily on the next I/O. Default: no-op.
+	virtual void releaseCachedResources() noexcept {}
 
 	/// Returns the Chunk format. Check \ref ChunkFormat for more information.
 	virtual ChunkFormat chunkFormat() const = 0;
--- a/src/chunkserver/chunkserver-common/disk_interface.h
+++ b/src/chunkserver/chunkserver-common/disk_interface.h
@@ -140,6 +140,11 @@ public:
 	virtual IChunk *instantiateNewConcreteChunk(uint64_t chunkId,
 	                                            ChunkPartType type) = 0;
 
+	/// Applies any disk-specific on-disk format to a brand-new chunk before its
+	/// header is sized and serialized. Called only from the new-chunk creation
+	/// path (not for duplicates). Default: no-op.
+	virtual int applyNewChunkFormat(IChunk * /*chunk*/) = 0;
+
 	/// Sets the number of blocks for \a chunk from \a originalBlocks to \a
 	/// newBlocks.
 	virtual void setChunkBlocks(IChunk *chunk, uint16_t originalBlocks,
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -1205,6 +1205,17 @@ std::pair<int, IChunk *> hddInternalCreateChunk(uint64_t chunkId,
 		return {SAUNAFS_ERROR_IO, ChunkNotFound};
 	}
 
+	// Let the disk stamp any format-specific state (e.g. compression) on the
+	// brand-new chunk before its header is sized and serialized.
+	status = disk->applyNewChunkFormat(chunk);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(chunk);
+		hddIOEnd(chunk);
+		disk->unlinkChunk(chunk);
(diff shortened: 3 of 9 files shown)


rolysr (approved):
LGTM.

src/chunkserver/chunkserver-common/chunk_interface.h
```diff
@@ -165,9 +165,11 @@ class IChunk {
...
 	virtual void shrinkToBlocks(uint16_t newBlocks) = 0;
 
-	/// Returns true if the Chunk is in an state considered as dirty.
-	/// For example the Chunk is fragmented and needs defragmentation.
-	virtual bool isDirty() = 0;
+	/// Releases memory the chunk caches only to speed up I/O (e.g. per-chunk
+	/// compression dictionaries). Called when the chunk leaves the open-chunks
+	/// cache (its descriptors are being closed), with the chunk locked, so the
+	/// caches can be rebuilt lazily on the next I/O. Default: no-op.
+	virtual void releaseCachedResources() {}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/chunkserver-common/disk_interface.h
```diff
@@ -140,6 +140,11 @@ class IDisk {
 	virtual IChunk *instantiateNewConcreteChunk(uint64_t chunkId,
 	                                            ChunkPartType type) = 0;
 
+	/// Applies any disk-specific on-disk format to a brand-new chunk before its
+	/// header is sized and serialized. Called only from the new-chunk creation
+	/// path (not for duplicates/replicas). Default: no-op.
+	virtual void applyNewChunkFormat(IChunk * /*chunk*/) {}
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1205,6 +1205,10 @@ std::pair<int, IChunk *> hddInternalCreateChunk(uint64_t chunkId,
 		return {SAUNAFS_ERROR_IO, ChunkNotFound};
 	}
 
+	// Let the disk stamp any format-specific state (e.g. compression) on the
+	// brand-new chunk before its header is sized and serialized.
+	disk->applyNewChunkFormat(chunk);
```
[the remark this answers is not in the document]

src/chunkserver/chunkserver-common/disk_interface.h
```diff
@@ -140,6 +140,11 @@ class IDisk {
 	virtual IChunk *instantiateNewConcreteChunk(uint64_t chunkId,
 	                                            ChunkPartType type) = 0;
 
+	/// Applies any disk-specific on-disk format to a brand-new chunk before its
+	/// header is sized and serialized. Called only from the new-chunk creation
+	/// path (not for duplicates/replicas). Default: no-op.
+	virtual int applyNewChunkFormat(IChunk * /*chunk*/) { return 0; }
```
lgsilva3087: Consider to avoid the implementation in the interface. Maybe DiskWithFD is better/recommended place.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1205,6 +1205,15 @@ std::pair<int, IChunk *> hddInternalCreateChunk(uint64_t chunkId,
 		return {SAUNAFS_ERROR_IO, ChunkNotFound};
 	}
 
+	// Let the disk stamp any format-specific state (e.g. compression) on the
+	// brand-new chunk before its header is sized and serialized.
+	status = disk->applyNewChunkFormat(chunk);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(chunk);
+		hddDeleteChunkFromRegistry(chunk);
```
lgsilva3087, line 1215: Should we do more things in the failure scenario before deleting the chunk from the registry. I see hddIOBegin was already called, maybe it acquired some resources.

lgsilva3087 (approved):
LGTM, great work. Please see my non-blocking comments.

src/chunkserver/chunkserver-common/disk_interface.h
dmga44: Yes, I moved the implementation to `FDDisk` files and applied the same pattern from the rest of this file:
```cpp
virtual int applyNewChunkFormat(IChunk * /*chunk*/) = 0;
```

src/chunkserver/hddspacemgr.cc
dmga44: Indeed, added a call to `hddIOEnd` and `unlinkChunk`. This effectively protects against some future implementation of `applyNewChunkFormat` from leaking (metadata files and OpenChunks instance) chunks at mid-creation.

PR #923, opened 2026-07-20
fix(tests): stabilize fileinfo availability check

rolysr:
Add an explicit wait for the expected chunkserver availability state before checking fileinfo output. This prevents the test from racing against master state updates after chunkservers are stopped.

The fileinfo assertion depends on the cluster reaching the reduced availability condition first; without waiting for that state, the command can observe stale readiness information and produce inconsistent results.

Related to: LS-980


lgsilva3087 (approved):
LGTM

dmga44 (approved):
LGTM.

PR #924, opened 2026-07-20
chore(mount): remove deprecated Polonaise

lgsilva3087:
Remove the legacy Thrift frontend and its test and build hooks.

Remove Thrift from build and packaging dependencies.

Related to: LS-981


rolysr (approved):
LGTM. Very nice to deprecate this.

dmga44 (approved):
Nice.

PR #929, opened 2026-07-24
fix(cs): retry attribute reads and propagate setChunkBlocks status

dmga44:
From suggestions in a recent PR review, the following changes are applied to strengthen chunkserver reliability:
- retry updateChunkAttributes() a bounded number of times when it returns SAUNAFS_ERROR_IO, since that error signals a transient read failure and not a confirmed-absent chunk. Applies to both the disk-scan path (hddAddChunkFromDiskScan) and the find-or-create path (hddChunkFindOrCreatePlusLock).
- change IDisk::setChunkBlocks() to return a status instead of void, and propagate/surface that status from its two call sites (hddInternalTruncate, hddInternalDuplicateTruncate), so a failed persist is now reported and cleaned up instead of silently leaving a chunk whose blocks count doesn't match its on-disk layout. CmrDisk::setChunkBlocks() always succeeds; ZoneFSDisk::setChunkBlocks() (zonefs_disk submodule, companion PR) now surfaces overwriteAllFragments() failures through this same path.

I also included this changes not strictly related:
- add zstd to the vcpkg manifest dependencies (it was only present under overrides, which pins a version without actually installing it) — needed by the zonefs_disk plugin's compressed-chunk support.
- suppress a second, previously-unseen helgrind TLS-teardown false positive (same root cause as the existing jobpool_stop_tls_teardown_race suppression, reached this time via mainNetworkThreadTerm()'s join instead of JobPool::stop()).

### DIFF
--- a/src/chunkserver/chunkserver-common/hdd_utils.cc
+++ b/src/chunkserver/chunkserver-common/hdd_utils.cc
@@ -20,7 +20,6 @@
 
 #include "chunkserver-common/hdd_utils.h"
 
-#include <cinttypes>
 #include <fcntl.h>
 #include <sys/time.h>
 
@@ -308,11 +307,12 @@ int hddIOBegin(IChunk *chunk, int newFlag, uint32_t chunkVersion) {
 bool hddScansInProgress() { return gScansInProgress != 0; }
 
 int hddUpdateChunkAttributesWithRetry(IDisk *disk, IChunk *chunk, bool isFromScan) {
-	int status = SAUNAFS_ERROR_IO;
-	for (int i = 0; i < kOpenRetryCount && status == SAUNAFS_ERROR_IO; ++i) {
-		if (i > 0) { usleep(kOpenRetry_ms * 1000); }
+	int status;
+	int attempt = 0;
+	do {
+		if (attempt > 0) { usleep((kOpenRetry_ms * 1000) << (attempt - 1)); }
 		status = disk->updateChunkAttributes(chunk, isFromScan);
-	}
+	} while (status == SAUNAFS_ERROR_IO && ++attempt < kOpenRetryCount);
 	return status;
 }
 
--- a/src/chunkserver/chunkserver-common/hdd_utils.cc
+++ b/src/chunkserver/chunkserver-common/hdd_utils.cc
@@ -20,6 +20,7 @@
 
 #include "chunkserver-common/hdd_utils.h"
 
+#include <cinttypes>
 #include <fcntl.h>
 #include <sys/time.h>
 
@@ -306,6 +307,15 @@ int hddIOBegin(IChunk *chunk, int newFlag, uint32_t chunkVersion) {
 
 bool hddScansInProgress() { return gScansInProgress != 0; }
 
+int hddUpdateChunkAttributesWithRetry(IDisk *disk, IChunk *chunk, bool isFromScan) {
+	int status = SAUNAFS_ERROR_IO;
+	for (int i = 0; i < kOpenRetryCount && status == SAUNAFS_ERROR_IO; ++i) {
+		if (i > 0) { usleep(kOpenRetry_ms * 1000); }
+		status = disk->updateChunkAttributes(chunk, isFromScan);
+	}
+	return status;
+}
+
 IChunk *hddChunkFindAndLock(uint64_t chunkId, ChunkPartType chunkType) {
 	LOG_AVG_TILL_END_OF_SCOPE0("chunk_find");
 
@@ -349,8 +359,9 @@ IChunk *hddChunkFindOrCreatePlusLock(IDisk *disk, uint64_t chunkid,
 			chunk->setState(ChunkState::Locked);
 			chunksMapLock.unlock();
 			if (chunk->validAttr() == 0) {
-				if (effectiveDisk->updateChunkAttributes(chunk, false) ==
-				    SAUNAFS_ERROR_NOCHUNK) {
+				int attrStatus =
+				    hddUpdateChunkAttributesWithRetry(effectiveDisk, chunk, false);
+				if (attrStatus != SAUNAFS_STATUS_OK) {
 					// The chunk was found as available, but we can not
 					// update its attributes, let's recreate it only if
 					// requested
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -1839,7 +1839,15 @@ static int hddInternalTruncate(uint64_t chunkId, ChunkPartType chunkType,
 		disk->setNeedRefresh(true);
 	}
 
-	disk->setChunkBlocks(chunk, chunk->blocks(), blocks);
+	status = disk->setChunkBlocks(chunk, chunk->blocks(), blocks);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(chunk);
+		safs::log_warn("truncate: file:{} - set chunk blocks error",
+		               chunk->fullMetaFilename().c_str());
+		hddIOEnd(chunk);
+		hddChunkRelease(chunk);
+		return status;
+	}
 
 	status = hddIOEnd(chunk);
 	if (status != SAUNAFS_STATUS_OK) {
@@ -2327,8 +2335,16 @@ static int hddInternalDuplicateTruncate(uint64_t chunkId, uint32_t chunkVersion,
 		return status;
 	}
 
-	dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
+	status = dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
 	dupDisk->setNeedRefresh(true);
+
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(dupChunk);
+		dupDisk->unlinkChunk(dupChunk);
+		hddDeleteChunkFromRegistry(dupChunk);
+		hddChunkRelease(originalChunk);
+		return status;
+	}
 
 	hddChunkRelease(dupChunk);
 	hddChunkRelease(originalChunk);
@@ -2551,10 +2567,17 @@ static inline void hddAddChunkFromDiskScan(IDisk *disk,
 	chunk->setVersion(version);
 	sassert(chunk->fullMetaFilename() == fullname);
 
-	{
-		disk->updateChunkAttributes(chunk, true);
-		chunk->setValidAttr(0);
+	int attrStatus = hddUpdateChunkAttributesWithRetry(disk, chunk, true);
+	if (attrStatus != SAUNAFS_STATUS_OK) {
+		safs::log_err(
+		    "hddAddChunkFromDiskScan: could not read attributes "
+		    "for {} (status {}); reporting damaged",
+		    fullname.c_str(), attrStatus);
+		hddReportDamagedChunk(chunk->id(), chunk->type());
+		hddChunkRelease(chunk);
+		return;
 	}
+	chunk->setValidAttr(0);
 
 	{
 		std::lock_guard testsLockGuard(gTestsMutex);
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -1842,8 +1842,8 @@ static int hddInternalTruncate(uint64_t chunkId, ChunkPartType chunkType,
 	status = disk->setChunkBlocks(chunk, chunk->blocks(), blocks);
 	if (status != SAUNAFS_STATUS_OK) {
 		hddAddErrorAndPreserveErrno(chunk);
-		safs::log_warn("truncate: file:{} - set chunk blocks error",
-		               chunk->fullMetaFilename().c_str());
+		safs::log_warn("truncate: file:{} - set chunk blocks error (status {})",
+		               chunk->fullMetaFilename(), status);
 		hddIOEnd(chunk);
 		hddChunkRelease(chunk);
 		return status;
@@ -2326,18 +2326,21 @@ static int hddInternalDuplicateTruncate(uint64_t chunkId, uint32_t chunkVersion,
 		return status;
 	}
 
-	status = hddIOEnd(dupChunk);
+	// Persist the new layout before ending dupChunk's IO, so it lands inside
+	// the same fsync as the rest of the header (matches hddInternalTruncate).
+	status = dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
+	dupDisk->setNeedRefresh(true);
+
 	if (status != SAUNAFS_STATUS_OK) {
 		hddAddErrorAndPreserveErrno(dupChunk);
+		hddIOEnd(dupChunk);
 		dupDisk->unlinkChunk(dupChunk);
 		hddDeleteChunkFromRegistry(dupChunk);
 		hddChunkRelease(originalChunk);
 		return status;
 	}
 
-	status = dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
-	dupDisk->setNeedRefresh(true);
-
+	status = hddIOEnd(dupChunk);
 	if (status != SAUNAFS_STATUS_OK) {
 		hddAddErrorAndPreserveErrno(dupChunk);
 		dupDisk->unlinkChunk(dupChunk);
@@ -2359,7 +2362,7 @@ int hddInternalDelete(IChunk *chunk, uint32_t version) {
 		hddChunkRelease(chunk);
 		return SAUNAFS_ERROR_WRONGVERSION;
 	}
-	if (chunk->owner()->unlinkChunk(chunk) < 0) {
+	if (chunk->owner()->unlinkChunk(chunk) != SAUNAFS_STATUS_OK) {
 		uint8_t err = errno;
 		hddAddErrorAndPreserveErrno(chunk);
 		safs_silent_errlog(LOG_WARNING,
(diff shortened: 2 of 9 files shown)


src/chunkserver/chunkserver-common/hdd_utils.cc
```diff
@@ -349,8 +359,9 @@ IChunk *hddChunkFindOrCreatePlusLock(IDisk *disk, uint64_t chunkid,
 			chunk->setState(ChunkState::Locked);
 			chunksMapLock.unlock();
 			if (chunk->validAttr() == 0) {
-				if (effectiveDisk->updateChunkAttributes(chunk, false) ==
-				    SAUNAFS_ERROR_NOCHUNK) {
+				int attrStatus =
+				    hddUpdateChunkAttributesWithRetry(effectiveDisk, chunk, false);
+				if (attrStatus != SAUNAFS_STATUS_OK) {
```
lgsilva3087, line 364: The retry idea makes a lot of sense to me. One thing I would like to check with
you is the condition itself, which also moved from `== SAUNAFS_ERROR_NOCHUNK`
to `!= SAUNAFS_STATUS_OK`. I am not sure whether that part was intended, so
please tell me if I am reading it wrong.

The branch it guards is the destructive one: both arms below call
`effectiveDisk->unlinkChunk(chunk)`, and the `kFindOnly` arm also reports the
chunk damaged and removes it from the registry. With the wider condition,
`SAUNAFS_ERROR_IO` now reaches it too, and IO means something different from
NOCHUNK: NOCHUNK is "the file is gone or is not a usable chunk", where deleting
costs nothing, while IO is "I could not read it right now", where the file is
most likely fine.

For `CmrDisk` this is a no-op, since its `updateChunkAttributes()` returns only
OK or NOCHUNK. The zoned-disk plugin does return `SAUNAFS_ERROR_IO`, for a
transient `open()` or `stat()` failure (EMFILE, EINTR, EIO) and for a failed
metadata read, and it classifies those as IO precisely so that a healthy chunk
is kept. So the effect would show up there rather than on CMR disks.

On the retry window itself, 4 attempts with a flat 5 ms sleep gives us about
15 ms, which may be a bit short for descriptor exhaustion to clear. Since
`hddIOBegin()` above already handles similar pressure with
`(kOpenRetry_ms * 1000) << i` plus a `gOpenChunks.freeUnused()` between
attempts, reusing that shape here could make the retry more effective. Only a
suggestion, not a blocker.

I also understand what motivated the wider condition: on IO the old code fell
through to `return chunk;` with `validAttr()` still 0, so the caller received a
chunk whose attributes were never loaded, which is not great either. Would
failing the request instead of removing the chunk cover that case for you?

```cpp
	int attrStatus =
	    hddUpdateChunkAttributesWithRetry(effectiveDisk, chunk, false);
	if (attrStatus == SAUNAFS_ERROR_NOCHUNK) {
		// existing recreate / report-damaged handling, unchanged
	}
	if (attrStatus != SAUNAFS_STATUS_OK) {
		// Transient read failure: keep the chunk on disk and fail this request.
		hddChunkRelease(chunk);
		return nullptr;
	}
```

That way the caller sees "not available right now" and can come back later,
and we keep the data in place.

src/chunkserver/hddspacemgr.cc
```diff
@@ -2551,10 +2567,17 @@ static inline void hddAddChunkFromDiskScan(IDisk *disk,
 	chunk->setVersion(version);
 	sassert(chunk->fullMetaFilename() == fullname);
 
-	{
-		disk->updateChunkAttributes(chunk, true);
-		chunk->setValidAttr(0);
+	int attrStatus = hddUpdateChunkAttributesWithRetry(disk, chunk, true);
+	if (attrStatus != SAUNAFS_STATUS_OK) {
```
lgsilva3087, line 2574: Reporting the chunk instead of silently ignoring the status is a good change.
What I am unsure about is the early return, so please correct me if I am
misreading the flow.

At this point the chunk is already in `gChunksMap`, added by
`hddChunkFindOrCreatePlusLock()` at the top of the function. The early return
does not undo that, it only skips `disk->chunks().insert(chunk)` and
`hddReportNewChunkToMaster()` below.

Those two are what make the chunk visible. The tester takes its candidates only
from the per-disk list, through `chunks().chunkToTest()` in
`DefaultDiskManager::getChunkToTest()`, so a chunk that never enters that list
is not scrubbed again. And without the new-chunk report the master never learns
this copy exists, so it replicates elsewhere while the copy stays on the disk
using space. The chunk itself is fine, it just becomes unmanaged.

Disk removal is not affected, since it walks `gChunksMap` rather than the
per-disk list, so nothing dangles there.

One more thing worth considering: this also changes CMR behaviour, not only the
zoned plugin. `STAT_CHUNKS_AT_DISK_SCAN` defaults to 1, so the scan does stat
the chunk, and a failed stat returns NOCHUNK. A file removed between readdir and
stat, which is normal during a concurrent delete, would now be reported damaged
and left as a registry-only entry.

Would keeping the chunk tracked cover what you wanted here?

```cpp
	int attrStatus = hddUpdateChunkAttributesWithRetry(disk, chunk, true);
	if (attrStatus != SAUNAFS_STATUS_OK) {
		safs::log_warn("hddAddChunkFromDiskScan: could not read attributes for {} (status {})",
		               fullname, attrStatus);
	}

	// Attributes are re-read on first access, whether or not the scan read them.
	chunk->setValidAttr(0);
```

and then continue into the insert and the report as before.

I should say the previous behaviour was not perfect either: it registered and
announced chunks that were genuinely missing or corrupt, and left the first
access or the tester to reject them. If you want to keep a damaged report for
that case, gating it on NOCHUNK rather than on any non-OK status would separate
the two, and it could still run after the chunk has been inserted and announced.

src/chunkserver/hddspacemgr.cc
```diff
@@ -2551,10 +2567,17 @@ static inline void hddAddChunkFromDiskScan(IDisk *disk,
...
+		safs::log_err(
+		    "hddAddChunkFromDiskScan: could not read attributes "
+		    "for {} (status {}); reporting damaged",
+		    fullname.c_str(), attrStatus);
```
lgsilva3087: Consider to not wrap the comment and to remove the .c_str() call, the moder logging functions don't need it.

src/chunkserver/hddspacemgr.cc
```diff
@@ -2327,9 +2335,17 @@ static int hddInternalDuplicateTruncate(uint64_t chunkId, uint32_t chunkVersion,
 		return status;
 	}
 
-	dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
+	status = dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
```
lgsilva3087: Not something this PR introduces, but it sits right on the line you are
touching, so it may be worth deciding here.

`hddIOEnd(dupChunk)` a few lines above is where the fsync happens, through
`fsyncChunk()` when `PERFORM_FSYNC` is set, and it flushes `metaFD`. For a disk
whose `setChunkBlocks()` persists something, and the zoned plugin's does, that
write lands on the same `metaFD` after the fsync has already run, and nothing
flushes it afterwards. The function then returns OK.

So a duplicate truncate can report success with the fragment table still
unflushed, and a power loss at that point leaves the duplicate describing the
old layout.

What made me raise it is that `hddInternalTruncate()` orders the same two calls
the other way, `setChunkBlocks()` first and `hddIOEnd()` after, so there the
fragment table is inside the fsync. The two paths disagree, and this one looks
like the accidental order rather than a deliberate one.

On CMR disks it makes no difference, since `CmrDisk::setChunkBlocks()` only
updates memory.

Would moving `setChunkBlocks()` above the `hddIOEnd(dupChunk)` call work here,
so it matches the truncate path? The failure handling would need a small rework,
since the current block assumes the chunk's IO is already ended, but the
resulting order would be the same one truncate already uses.

lgsilva3087:
Nothing about the code here, just the description if you feel like tweaking it.

The part I found most valuable while reading the truncate path is one the
description does not mention: before this change `setChunkBlocks()` returned
void, so a failed persist was invisible and `hddInternalTruncate()` could return
success to the master for a truncate whose layout never reached the disk. That
seems like a stronger reason for the change than the blocks count mismatch,
which the next attribute re-read would have corrected anyway.

Two tiny things if you end up editing it: the cleanup part applies to the
duplicate truncate path, and the IO retry does not trigger on `CmrDisk`, whose
`updateChunkAttributes()` returns only OK or NOCHUNK.

src/chunkserver/chunkserver-common/hdd_utils.cc
```diff
@@ -306,6 +307,15 @@ int hddIOBegin(IChunk *chunk, int newFlag, uint32_t chunkVersion) {
 
 bool hddScansInProgress() { return gScansInProgress != 0; }
 
+int hddUpdateChunkAttributesWithRetry(IDisk *disk, IChunk *chunk, bool isFromScan) {
```
lgsilva3087, line 309: Would a small test for the retry helper be worth adding? It looks cheaper than
it might seem: `CmrDisk` is concrete and already constructed directly in
`disk_unittest.cc`, and `updateChunkAttributes()` is virtual, so a subclass
overriding just that one method could return IO a couple of times and then OK.

That would cover the two things worth pinning down: that it retries and
eventually succeeds, and that it gives up after `kOpenRetryCount` attempts
rather than looping.

Not blocking, and no new mocking scaffolding needed as far as I can tell.

lgsilva3087:
A few small things, none of them blocking.

- `hdd_utils.cc:23`: `<cinttypes>` looks unused, nothing in the file needs it
  (`PRINTTHIS` is a TracePrinter macro, not a `PRI*` one).

- `hdd_utils.h:87`: the doc says "with backoff" but the sleep is a constant
  `kOpenRetry_ms`. Worth aligning one to the other, and it resolves itself if
  the retry ends up reusing the `hddIOBegin()` escalation.

- `hdd_utils.cc:311`: `status` is seeded with `SAUNAFS_ERROR_IO` just to enter
  the loop, so a `kOpenRetryCount` of 0 would return an IO error no disk
  produced. Only latent today, but a `do while` would avoid the question.

- `hdd_utils.cc:362`: the wrapped `int attrStatus =` line gets joined by
  clang-format at the repo's 100 column limit, `/leil-format` handles it.

- `hddspacemgr.cc:1846` and `:2575`: the `.c_str()` is not needed for
  `safs::log_warn` / `safs::log_err`, they take the `std::string`. The nearby
  `safs_silent_errlog()` calls do need it, so the mix makes sense.

- `hddspacemgr.cc:2573`: the message is split across two literals. Keeping it
  as one keeps it greppable, and dropping `; reporting damaged` makes it fit
  the line limit.

src/chunkserver/hddspacemgr.cc
```diff
@@ -2327,9 +2335,17 @@ static int hddInternalDuplicateTruncate(uint64_t chunkId, uint32_t chunkVersion,
...
 	dupDisk->setNeedRefresh(true);
 
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(dupChunk);
+		dupDisk->unlinkChunk(dupChunk);
```
lgsilva3087, line 2346: Tiny one, and it follows what the surrounding code already does, so feel free
to ignore it.

`unlinkChunk()` returns a status and can fail for real here, since `CmrDisk`
reports a bad disk path and a failed move-to-trash through it. This block
discards it and drops the registry entry regardless, so if the unlink fails the
duplicate's files stay on disk untracked, and the next scan registers and
announces that copy carrying the fragment table `setChunkBlocks()` never
managed to update.

Almost every other call site discards it too, so this is an existing convention
rather than something the PR changes. The one place that does check,
`hddInternalDelete()`, tests `< 0`, which does not catch the positive
`SAUNAFS_ERROR_*` values, so the convention looks more unsettled than
deliberate.

Even just logging a failed unlink before dropping the entry would leave a trace
to find later.

lgsilva3087 (commented):
Great work. Pleasee see my comments.
Some of them are not introduced by this PR, I decided to comment on them to not forget.

src/chunkserver/chunkserver-common/hdd_utils.cc
dmga44: Yes, the change was intended. Previous `== SAUNAFS_ERROR_NOCHUNK` was equivalent to `!= SAUNAFS_STATUS_OK` because the `updateChunkAttributes` function only returned OK or NOCHUNK statuses.

The current implementation also threads via that `if` the `SAUNAFS_ERROR_IO` statuses coming from `hddUpdateChunkAttributesWithRetry`, which as you mentioned perform destructive actions on the chunk. Specially it performs a `hddReportDamagedChunk` that allows master to repair it and the IO error surfaces after some retries and backoff (I've added it), so it is quite final.

Also, please consider that callers of this function do not issue destructive actions on a chunk with really broken data. So, the function with those responsabilities of cleanup is this one.

dmga44: As a bonus change, I've updated the `CmrDisk::updateChunkAttributes` to mirror the error differentiation of the other implementation.

src/chunkserver/hddspacemgr.cc
dmga44: I went for the report the chunk as damaged and almost applying exactly your suggestion:
```cpp
	int attrStatus = hddUpdateChunkAttributesWithRetry(disk, chunk, true);
	if (attrStatus != SAUNAFS_STATUS_OK) {
		safs::log_warn("hddAddChunkFromDiskScan: could not read attributes for {} (status {})",
		               fullname, attrStatus);
		hddReportDamagedChunk(chunk->id(), chunk->type());
	}

	// Attributes are re-read on first access, whether or not the scan read them.
```

This is to keep it inline with the above comment, that `hddUpdateChunkAttributesWithRetry` is final when non-OK.

src/chunkserver/hddspacemgr.cc

src/chunkserver/hddspacemgr.cc
dmga44: Applied the following way:
```cpp
	// Persist the new layout before ending dupChunk's IO, so it lands inside
	// the same fsync as the rest of the header (matches hddInternalTruncate).
	status = dupDisk->setChunkBlocks(dupChunk, originalChunk->blocks(), blocks);
	dupDisk->setNeedRefresh(true);

		...
	}

	status = hddIOEnd(dupChunk);
```

src/chunkserver/chunkserver-common/hdd_utils.cc
dmga44: I've added a couple of very simple ones: `UpdateChunkAttributesRetrySucceedsAfterTransientErrors` and `UpdateChunkAttributesRetryGivesUpAfterMaxAttempts`.

src/chunkserver/hddspacemgr.cc
dmga44: I left it as it was. There are at least 22 call sites discarding its return value.

Also, since you were mentioning it, the `unlinkChunk` function returned 0 or -1 up to #215 . I think would be nice to make all `unlinkChunk` aligned with the usual `SAUNAFS_*` returns.

src/chunkserver/chunkserver-common/hdd_utils.cc
```diff
@@ -20,6 +20,7 @@
 
 #include "chunkserver-common/hdd_utils.h"
 
+#include <cinttypes>
 #include <fcntl.h>
 #include <sys/time.h>
```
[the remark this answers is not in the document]
dmga44: Applied.

src/chunkserver/chunkserver-common/hdd_utils.cc
```diff
@@ -306,6 +307,15 @@ int hddIOBegin(IChunk *chunk, int newFlag, uint32_t chunkVersion) {
...
 bool hddScansInProgress() { return gScansInProgress != 0; }
 
+int hddUpdateChunkAttributesWithRetry(IDisk *disk, IChunk *chunk, bool isFromScan) {
+	int status = SAUNAFS_ERROR_IO;
+	for (int i = 0; i < kOpenRetryCount && status == SAUNAFS_ERROR_IO; ++i) {
+		if (i > 0) { usleep(kOpenRetry_ms * 1000); }
+		status = disk->updateChunkAttributes(chunk, isFromScan);
+	}
+	return status;
+}
```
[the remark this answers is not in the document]
dmga44: Backoff applied.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1839,7 +1839,15 @@ static int hddInternalTruncate(uint64_t chunkId, ChunkPartType chunkType,
 		disk->setNeedRefresh(true);
 	}
 
-	disk->setChunkBlocks(chunk, chunk->blocks(), blocks);
+	status = disk->setChunkBlocks(chunk, chunk->blocks(), blocks);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(chunk);
+		safs::log_warn("truncate: file:{} - set chunk blocks error",
+		               chunk->fullMetaFilename().c_str());
```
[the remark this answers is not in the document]
dmga44: Status included.

src/chunkserver/hddspacemgr.cc
dmga44: The reporting of the damaged chunk makes the test `test_auto_recovery_xor_repair` fail, so I removed it. Though make sense to me, I think it is better to maintain the dev behavior for now.

rolysr (approved):
LGTM.

lgsilva3087 (approved):
LGTM

PR #932, opened 2026-07-29
test: scale network partition sleeps

lgsilva3087:
Fixed sleep 0.2s and sleep 10s windows didn't scale with a slower environment, so a loaded run could kill chunkservers before the in-flight chunk landed anywhere, losing it permanently instead of exercising the intended recovery path. The window is tightest on transactional backends like FDB, where chunk metadata crosses the network instead of an in-memory update. Scale both with timeout_rescale_seconds, matching the convention used by sibling tests.

Related to: LS-998


rolysr (approved):

dmga44 (approved):
LGTM.

PR #933, opened 2026-07-29
fix(cs): duplicate chunks in the target format

dmga44:
## Summary

`hddInternalDuplicate()` and `hddInternalDuplicateTruncate()` always created the
copy as a legacy V1 chunk and wrote it one block at a time through
`writeChunkBlock()`. That is harmless for plain chunks, but for the SMR
implementation it silently drops compression on every local duplicate, and
simply making the copy format-aware without also fixing the copy loop would mint
**one fragment per block** — reintroducing the write-amplification failure class
already fixed for the garbage collector there.

This PR makes a duplicate adopt its destination disk's configured format, and
rewrites the four block-copy loops as a single batched helper so that
format-awareness does not come at the cost of fragment explosion.

Needs the companion commit on the SMR implementation's `feat-data-compression`
branch: every `IDisk` implementation has to match the new
`serializeEmptyChunkSignature()` signature, and the new SMR integration test
lives in that repo too.

## Changes

**A duplicate takes the destination disk's format.** Both dup paths now call the
existing `applyNewChunkFormat()` hook, exactly as `hddInternalCreateChunk()`
already does. The copy does *not* inherit the source chunk's format — same
behaviour as a replication receiver.

**`serializeEmptyChunkSignature()` gained an optional `formatSource` chunk** so
the placeholder header written for the duplicate matches the format that was
just stamped. The parameter is defaulted; `CmrDisk` ignores it.

**Fixed a latent data-loss bug in `hddInternalDuplicateTruncate()`.** Its
`writeChunkHeader()` ran *after* the data copy and spanned the whole header
region, so on a compressed copy it erased both the dictionary bytes and the
`dictLen` field that the copy's first batch had just persisted there.
`setChunkBlocks()` then restored the fragment table but not the dictionary,
leaving every compressed block in the duplicate undecodable once the chunk left
the open-chunks cache. The header is now written before the copy and only the
CRCs are persisted afterwards, via `writeCrc()`. All three
`writeChunkHeader()` call sites now precede their data writes. The
`getChunkHeaderBuffer()` capture also moved after the format is stamped, since
that call sizes the buffer from the format.

**Collapsed all four block-copy loops** (one in duplicate, three in duptrunc)
into a single `hddCopyChunkBlocks()`, copying `kChunkCopyBatchBlocks` (256)
blocks at a time regardless of format. The only per-format difference left is
the write call:

- a zoned destination goes through the existing multi-block
  `writeChunkBlocks()`, so a compressed copy packs many blocks into one
  fragment instead of one fragment per block. 256 is also the batch a
  compressed write flushes at, so a copy batch maps onto one fragment;
- everything else appends through `writeChunkData()`, which also drops that
  path from one write syscall per 64 KiB block to one per batch.

**Simplified duptrunc's expand/shrink branches.** They differed only in how many
blocks they copy, so the copy moved out of them entirely, leaving just the
expanding CRC fill and the misaligned trailing block as the genuine special
cases.

**Reused the IO buffers pool for staging.** The copy stages its blocks in a
buffer from the existing pool rather than allocating one per call, so
duplications reuse the allocation and the 16 MiB is accounted and reclaimed
alongside the read and write pools under `MAX_BUFFERS_POOL_SIZE_MB` (the
replicator already pools a whole chunk's worth, 64 MiB, through the same pool).
That pool's buffer type was already exactly what this needs — an IO-aligned
block buffer with no header — so it is reused as is and renamed
`ReplicatorBuffer` → `ChunkCopyBuffer` (with its pool accessor and block
counter) now that replication is not its only user. That also renames the pool's
column in the `releaseOldIoBuffers` debug line from `replicate` to `chunk copy`,
which `sfschunkserver_check_no_buffer_in_use()` greps for, so the test helper
and the chunkserver README follow.

## Impact

| | Before | After |
|---|---|---|
| Duplicate of a compressed chunk | copied as uncompressed V1 | keeps the destination's compressed format |
| Fragments per duplicated chunk (SMR) | one per block, up to 1024 | one per 256-block batch, ≤ 4 for a full chunk |
| Write syscalls per chunk (non-SMR) | one per 64 KiB block | one per 16 MiB batch |
| Duptrunc of a compressed chunk | dictionary erased after copy → undecodable blocks | dictionary preserved |

No on-disk format change, no configuration change, and no behavioural change for
plain (non-SMR) chunks beyond the larger write batches.

## Bonus

Two follow-up refactors of the same duplication code, done after the above landed.

**Guarded the failure-path cleanup.** Both duplicate operations unwound a
half-built copy by hand on every early return: end the copy's I/O, unlink it,
drop it from the registry, end the source's I/O, release the source — five
steps, repeated ten times across the two functions (113 cleanup calls in
total), each site having to know which of the five actually applied (a copy
with no files yet, an I/O already ended, whether the source should be
reported damaged). A small RAII guard now owns both chunks: each phase marks
what it completed, and the guard's teardown unwinds only what's still
outstanding, in the same order the hand-written blocks used. A final step
keeps the copy and releases both locks. No intended behaviour change.

**Extracted the phases both operations share.** The two functions ran very
long bodies, most of it the same sequence: pick a destination disk,
register the copy, bump the source's version, open both chunks, stamp the
copy's header from the source's CRCs, copy the blocks, commit — with one
operation's truncation-specific logic and old-style variable declarations
obscuring the overlap with the other. Five short helpers now hold those
phases, three of them shared by both callers. Both operations drop to a
short sequence of named steps instead of one long body.

Comparing the two turned up a real bug: one of the operations was renaming
the *copy* on a version bump instead of the source, but a freshly registered
chunk has no files yet — they're created later — so the rename found nothing
to act on, and every request of that kind carrying a new version failed
before any data moved. That path had never run to completion and nothing
covered it. Fixed to rename the source, matching how every other version-
changing operation in this file behaves.

A handful of other small divergences between the two callers were settled as
no-ops in the shared helpers (a redundant seek to an already-zero offset, an
early CRC copy one operation used to redo later, which chunk's identifying
fields a placeholder header is built from) and two cosmetic log-line
differences were unified. Also drops a dead assertion and three redundant
null checks that were never reachable once the shared setup guarantees the
chunk exists.

### DIFF
--- a/src/chunkserver/hddspacemgr.cc
+++ b/src/chunkserver/hddspacemgr.cc
@@ -1533,18 +1533,275 @@ private:
 	bool committed_ = false;
 };
 
+/// The buffers a chunk copy sets up once and then threads through every step:
+/// the destination's header buffer, which accumulates the copy's CRCs until
+/// they are persisted, and both chunks' in-memory CRC blocks.
+///
+/// The header buffer belongs to the calling thread and is sized for one chunk,
+/// so it only stays valid while nothing else on this thread asks for another
+/// chunk's header buffer.
+struct ChunkCopyBuffers {
+	uint8_t *header = nullptr;
+	uint8_t *sourceCrc = nullptr;
+	uint8_t *copyCrc = nullptr;
+};
+
+/// Picks a destination disk and registers the copy on it, handing the new chunk
+/// to @p guard and to @p copy.
+static int createChunkCopy(ChunkCopyGuard &guard, ChunkPartType chunkType, uint64_t copyChunkId,
+                           uint32_t copyChunkVersion, IChunk **copy) {
+	IChunk *newChunk = ChunkNotFound;
+
+	{
+		std::unique_lock disksUniqueLock(gDisksMutex);
+
+		IDisk *copyDisk = gDiskManager->getDiskForNewChunk(chunkType);
+
+		if (copyDisk == DiskNotFound) {
+			// The guard lives in the caller, so the source chunk is released
+			// well after this scope drops gDisksMutex.
+			return SAUNAFS_ERROR_NOSPACE;
+		}
+
+		newChunk = hddChunkCreate(copyDisk, copyChunkId, chunkType, copyChunkVersion);
+	}
+
+	if (newChunk == ChunkNotFound) { return SAUNAFS_ERROR_CHUNKEXIST; }
+
+	guard.copyRegistered(newChunk);
+	*copy = newChunk;
+
+	return SAUNAFS_STATUS_OK;
+}
+
+/// Opens the source chunk for reading, first moving it from @p chunkVersion to
+/// @p chunkNewVersion when the two differ.
+///
+/// The bump is the same three steps hddInternalUpdateVersion() and
+/// hddInternalTruncate() use, in the same order: rename the files to the new
+/// version, open with the version the header still holds so its CRCs validate,
+/// then rewrite the header. All three act on the chunk whose version changes,
+/// which is the source - the copy is created at its own version and never
+/// renamed.
+static int openSourceForCopy(ChunkCopyGuard &guard, IChunk *sourceChunk, uint32_t chunkVersion,
+                             uint32_t chunkNewVersion, const char *errorMsg) {
+	if (chunkNewVersion == chunkVersion) {
+		const int status = hddIOBegin(sourceChunk, 0);
+
+		if (status != SAUNAFS_STATUS_OK) {
+			hddAddErrorAndPreserveErrno(sourceChunk);
+			guard.sourceIsDamaged();
+			return status;
+		}
+
+		guard.sourceIoBegan();
+
+		return SAUNAFS_STATUS_OK;
+	}
+
+	if (sourceChunk->renameChunkFile(chunkNewVersion) < 0) {
+		hddAddErrorAndPreserveErrno(sourceChunk);
+		safs_silent_errlog(LOG_WARNING, "%s: file:%s - rename error", errorMsg,
+		                   sourceChunk->fullMetaFilename().c_str());
+		return SAUNAFS_ERROR_IO;
+	}
+
+	int status = hddIOBegin(sourceChunk, 0, chunkVersion);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(sourceChunk);
+		return status;  //can't change file version
+	}
+	guard.sourceIoBegan();
+
+	status = sourceChunk->owner()->overwriteChunkVersion(sourceChunk, chunkNewVersion);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(sourceChunk);
+		safs_silent_errlog(LOG_WARNING, "%s: file:%s - write error", errorMsg,
+		                   sourceChunk->fullMetaFilename().c_str());
+		return SAUNAFS_ERROR_IO;
+	}
+
+	return SAUNAFS_STATUS_OK;
+}
+
+/// Opens the copy for writing and gives it a valid header: the destination
+/// disk's format, an empty signature carrying its own id and version, and the
+/// source's CRC block, which the copy then updates in place as blocks arrive.
+static int prepareChunkCopyForWriting(ChunkCopyGuard &guard, IChunk *sourceChunk, IChunk *copy,
+                                      const char *errorMsg, ChunkCopyBuffers *buffers) {
+	int status = hddIOBegin(copy, 1);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(copy);
+		return status;
+	}
+	guard.copyIoBegan();
+
+	IDisk *copyDisk = copy->owner();
+
+	// The copy takes the destination disk's configured format, not the source
+	// chunk's, and must be stamped before getChunkHeaderBuffer() below sizes
+	// the header buffer from it.
+	status = copyDisk->applyNewChunkFormat(copy);
+	if (status != SAUNAFS_STATUS_OK) {
+		hddAddErrorAndPreserveErrno(copy);
+		return SAUNAFS_ERROR_IO;
+	}
+
+	// hddChunkCreate() built the copy from the requested id and version, so the
+	// chunk itself carries what the signature needs.
+	buffers->header = copy->getChunkHeaderBuffer();
+	memset(buffers->header, 0, copy->getHeaderSize());
+
+	uint8_t *headerPtr = buffers->header;
+	copyDisk->serializeEmptyChunkSignature(&headerPtr, copy->id(), copy->version(), copy->type(),
+	                                       copy);
+
+	buffers->sourceCrc = gOpenChunks.getResource(sourceChunk->metaFD()).crcData();
+	buffers->copyCrc = gOpenChunks.getResource(copy->metaFD()).crcData();
+
+	// Into the header buffer to reach the device, and into the in-memory block
+	// the write paths keep up to date.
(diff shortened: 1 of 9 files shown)


rolysr (approved):
LGTM.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
-	IChunk *dupChunk, *originalChunk;
-	IDisk *dupDisk, *originalDisk;
+/// Blocks per read+write batch when copying a chunk. It matters most on a
+/// zoned destination, which writes one fragment per call, so batching keeps a
+/// full chunk well below the fragment limit; 256 matches the batch a compressed
+/// write flushes at, and bounds the staging buffer at 16 MiB.
+static constexpr uint16_t kChunkCopyBatchBlocks = 256;
+
+/// Which side of a copy failed; only a failed read means a damaged source.
+enum class ChunkCopyResult { Ok, ReadFailed, WriteFailed };
```
lgsilva3087: Consider:
```cpp
enum class ChunkCopyResult : std::uint8_t {...}
```

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
+/// still describe the copied bytes) and updating \a dupCrcData.
+///
+/// Both formats copy in batches of kChunkCopyBatchBlocks: zoned destinations
+/// through writeChunkBlocks(), the rest appending to the caller's data seek
+/// through writeChunkData().
+static ChunkCopyResult hddCopyChunkBlocks(IChunk *sourceChunk, IChunk *dupChunk,
+                                          uint16_t blockCount,
+                                          const uint8_t *sourceCrcData,
+                                          uint8_t *dupCrcData,
+                                          const char *errorMsg) {
```
lgsilva3087: Format please.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
+/// Unwinds a half-built chunk copy on the failure paths of duplicate and
+/// duptrunc.
+///
+/// Both operations lock a source chunk, register a copy, open both for I/O and
+/// then run a dozen fallible steps, and every one of them has to undo exactly
+/// the steps that already succeeded - in reverse order, without ending an I/O
+/// twice and without leaving a registered chunk whose file was never written.
+/// Spelled out at each site that is the same five calls repeated nine times.
+/// Here every step records what it completed and the destructor undoes whatever
+/// is still outstanding, so a new early return cannot forget one.
```
lgsilva3087, line 1458: Nice

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
-		}
-	} else {  // chunkNewVersion == chunkVersion
-		status = hddIOBegin(originalChunk, 0);
 		if (status != SAUNAFS_STATUS_OK) {
-			hddAddErrorAndPreserveErrno(originalChunk);
-			hddDeleteChunkFromRegistry(dupChunk);
-			hddReportDamagedChunk(chunkId, chunkType);
-			hddChunkRelease(originalChunk);
+			hddAddErrorAndPreserveErrno(sourceChunk);
+			guard.sourceIsDamaged();
```
lgsilva3087: Consider to rename the function to markSourceAsDamaged. We usually use `is` to check for a boolean result, but the function just sets the variable to true.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
+
+		guard.sourceIoBegan();
+
+		return SAUNAFS_STATUS_OK;
+	}
+
+	if (sourceChunk->renameChunkFile(chunkNewVersion) < 0) {
+		hddAddErrorAndPreserveErrno(sourceChunk);
+		safs_silent_errlog(LOG_WARNING, "%s: file:%s - rename error", errorMsg,
+		                   sourceChunk->fullMetaFilename().c_str());
```
lgsilva3087: Consider to use the new logging functions in new code.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
-			hddAddErrorAndPreserveErrno(dupChunk);
-			safs_silent_errlog(LOG_WARNING,
-			                   "duplicate: file:%s - hdr write error",
-			                   dupChunk->fullMetaFilename().c_str());
-			hddIOEnd(dupChunk);
-			dupDisk->unlinkChunk(dupChunk);
-			hddDeleteChunkFromRegistry(dupChunk);
-			hddIOEnd(originalChunk);
-			hddChunkRelease(originalChunk);
+		if (copyDisk->writeChunkHeader(copy) != SAUNAFS_STATUS_OK) {
```
lgsilva3087, line 1667: **Pre-existing data-integrity risk (non-blocking for this PR):** the destination receives a scanner-valid final ID/version header here before its body is copied. If the chunkserver crashes after this point, `ChunkCopyGuard` cannot remove the incomplete copy. On restart, the scanner can report that partial same-version chunk as valid; because its block count is reconstructed from the copied prefix, missing blocks are returned as zeros rather than detected as corruption. This ordering already existed in the old duplicate/duptrunc paths, so I am not asking to block this PR, but format-aware duplication extends its use. Please track a follow-up to build copies under a non-scannable temporary/incomplete state, durably finish them, and only then atomically publish the final chunk. A kill/restart-between-batches regression test could be included.

src/chunkserver/hddspacemgr.cc
```diff
@@ -1352,234 +1354,516 @@ static int hddInternalTestChunk(uint64_t chunkId, uint32_t version,
...
+			guard.sourceIsDamaged();
 			return status;
 		}
+
+		guard.sourceIoBegan();
+
+		return SAUNAFS_STATUS_OK;
+	}
+
+	if (sourceChunk->renameChunkFile(chunkNewVersion) < 0) {
```
lgsilva3087, line 1594: **Pre-existing latent data-integrity risk (non-blocking for this PR):** after `renameChunkFile()` succeeds, a failure in `hddIOBegin()` or `overwriteChunkVersion()` leaves the source under `chunkNewVersion` while its header still contains `chunkVersion`, or a partially written value. The failed copy is cleaned up, but the source transition is not rolled back, so later access or restart can reject the source. Current master handlers pass the same source version for both arguments, so this branch is not reachable through today's production protocol. Similar ordering already exists in update, truncate, and the old duptrunc path. Please track a transactional, recoverable version transition before enabling callers with distinct source versions, with fault injection after each rename and header-write step.

lgsilva3087 (commented):
Great work. Please see my comments.

src/chunkserver/hddspacemgr.cc

src/chunkserver/hddspacemgr.cc
dmga44: Format applied.

src/chunkserver/hddspacemgr.cc
dmga44: Renamed to `markSourceAsDamaged`.

src/chunkserver/hddspacemgr.cc
dmga44: For most of these cases the newer `safs::log_warn_with_error_code` was used.

lgsilva3087 (approved):
LGTM.

PR #936, opened 2026-07-31
chore: rebrand saunafs directory paths to leil

rolysr:
Rename ETC/DATA/INCL/PLUGINS install directories from "saunafs" to "leil", falling back to the old path at runtime when the new one is missing.

Skip creating a given compat symlink (instead of failing the install) when a real, non-symlink "saunafs" directory already occupies that path, e.g. left over from a pre-rebrand install on a long-lived build machine, or an existing deployment's actual data/config directory - never delete what's already there.

Related to: LS-1005


rolysr:
@gemini-code-assist review

svaroq (approved):
Great job, @rolysr! LGTM 👍

lgsilva3087 (approved):
LGTM

ralcolea (approved):
Great job @rolysr!! 💪🔥🚀

PR #943, opened 2026-09-02
ci: fix git clone for docker image building

svaroq:
Configure Git to use HTTP/1.1 to avoid GitHub clone failures caused by HTTP/2.0 during the Docker build. Because git clone used HTTP/2.0 by default and asked for username and password, which caused build pipelines to fail.


rolysr (approved):

svaroq:
But for now we need it as a workaround

PR #948, opened 2026-09-09
ci: set buildDiscarder to limit the number of builds to keep

svaroq:
Added numToKeepStr parameter for buildDiscarder to limit retained builds to 70 in order to reduce disk usage.


rolysr (approved):

ralcolea (approved):
LGTM! 👍

PR #949, opened 2026-09-09
chore: bump version to v5.13.0

lgsilva3087:
Work in this cycle adds packets to the chunkserver registration exchange. A node must be able to tell whether the peer it just connected to understands them, and that test is a comparison against the release version, so the target release advances to give the comparison a number to use.

The gate itself arrives with the feature. This commit only moves the number, so a build from this point reports v5.13.0.

Related to: LS-1075


svaroq (approved):

ralcolea (approved):
LGTM! 👍

dmga44 (approved):

PR #952, opened 2026-09-10
feat: build the Go API tools and manage them in the test harness

GigaCronos:
Build the leilfs-api Go tools (API server, leil CLI, monitoring) as part of the leilfs CMake build when leilos symlinks their sources into api/, and add API daemon lifecycle support to the test harness. leilfs tracks nothing about the private leilfs-api: a bare checkout skips the api/ build and carries no API-dependent tests.

- CMakeLists.txt: add_subdirectory(api), guarded by EXISTS api/CMakeLists.txt.
- .gitignore: ignore the /api/ tree and tests/test_suites/API_tests.
- tests/tools/saunafs.sh: opt-in API_SERVERS and leilfs_api_daemon (start|stop|restart|kill|isalive), a foreground server managed by PID+pidfile with a TERM->KILL shutdown, per-instance env cfg, leilfs_api_ready (/readyz).
- tests/tools/test.sh: reap stray API servers on teardown.
- tests/tools/config.sh: SAUNAFS_FORCE_LINUX opt-out so WSL2 is not misdetected as a Windows client.


tests/tools/saunafs.sh
```diff
@@ -15,6 +15,7 @@ setup_local_empty_saunafs() {
 	local number_of_masterservers=${MASTERSERVERS:-1}
 	local number_of_chunkservers=${CHUNKSERVERS:-1}
 	local number_of_mounts=${MOUNTS:-1}
+	local number_of_api_servers=${API_SERVERS:-0}
```
antuan96314, line 18: Shouldn't default minimum number of servers be 1?

GigaCronos: The API is a submodule of leilfs, so if we set API_SERVERS to 1, but the module is not in the installation files the test will fail(this would happen with all the tests). I will set it to 1 after we integrate the API in leilfs repo.

antuan96314 (approved):

tests/tools/saunafs.sh
```diff
@@ -147,6 +148,18 @@ setup_local_empty_saunafs() {
...
+	# cluster is up, so each server's /readyz (which probes the master) passes.
+	# The API daemon path is POSIX-only (no Windows-client equivalent yet), so
+	# refuse it on a Windows system rather than misbehave silently.
+	if [[ $number_of_api_servers -gt 0 && ${saunafs_info_[is_windows_system]} -eq 1 ]]; then
+		test_fail "API_SERVERS is not supported on Windows systems"
+	fi
+	for ((apiid = 0; apiid < number_of_api_servers; ++apiid)); do
+		add_api_server_ $apiid
+	done
+	saunafs_info_[api_server_count]=$number_of_api_servers
```
lgsilva3087: This is only reached when every `add_api_server_` succeeded. `add_api_server_` launches the process first and then waits for `/readyz` (line 876); if that wait times out it calls `test_fail`, so the counter stays unset, `leilfs_stop_all_api_servers` sees `count=0` and the `leilfs-api` processes already started are never killed at teardown (the `pkill ... sfs` pattern in `test.sh` does not match them, as your comment there notes). The next test reuses the same ports from `FIRST_PORT_TO_USE`, so a stale server can steal a master/chunkserver port or answer a new server's `/readyz`. Only the sudo `pkill -9 -u saunafstest` in `run-test.sh` cleans this up, and that does not run for a test sourced by hand.

Suggestion: set `saunafs_info_[api_server_count]` before the loop (or bump it in `add_api_server_` right after `leilfs_api_daemon start`). `api_signal_` already returns 0 for a missing pidfile, so counting a server that never launched is harmless.

tests/tools/saunafs.sh
```diff
@@ -736,6 +834,60 @@ add_chunkserver_() {
...
+# create_leilfs_api_env_ <id> <port>: emit the LEILFS_* environment for one
+# leilfs-api instance pointed at the running master. Tests may inject extra
+# "KEY=VALUE" lines via API_EXTRA_CONFIG or API_<id>_EXTRA_CONFIG (pipe-
+# separated), matching the *_EXTRA_CONFIG convention of the other daemons.
+create_leilfs_api_env_() {
+	local api_id=$1
+	local api_port=$2
+	local this_api_cfg_variable="API_${api_id}_EXTRA_CONFIG"
+	# localhost is deliberate: the master listens on * (MATOCL_LISTEN_HOST is
+	# unset in the test config), so localhost avoids reverse-DNS and works for the
```
lgsilva3087: Small one: I don't think DNS comes into play here, `get_ip_addr` gives a numeric IP, so both choices behave the same. `localhost` is fine (the CGI server at line 915 does the same), maybe just say that and drop the `MATOCL_LISTEN_HOST` hypothetical, so the comment stays short.

tests/tools/saunafs.sh
```diff
@@ -242,6 +255,91 @@ saunafs_metalogger_daemon() {
...
+	local pidfile=${saunafs_info_[api${id}_pidfile]}
+	local logfile=${saunafs_info_[api${id}_log]}
+	case "$action" in
+		start)
+			# Export the generated env and exec the server in the background; the
+			# subshell PID becomes leilfs-api's PID via exec, so it can be signalled.
+			( set -a; source "$env_file"; set +a; exec leilfs-api ) >>"$logfile" 2>&1 &
+			local pid=$!
+			echo "$pid" >"$pidfile"
+			saunafs_info_[api${id}_pid]=$pid
```
lgsilva3087: I could not find a reader of `saunafs_info_[api${id}_pid]`, `isalive` and `api_signal_` both use the pidfile. Maybe drop these three assignments (277, 279, 280) so the pidfile stays the only source of truth? Not important if you have a use for it coming.

tests/tools/saunafs.sh
```diff
@@ -147,6 +148,18 @@ setup_local_empty_saunafs() {
 		sleep 3 # A reasonable fallback
 	fi
 
+	# Start leilfs-api servers (opt-in via API_SERVERS). Done last, once the
+	# cluster is up, so each server's /readyz (which probes the master) passes.
+	# The API daemon path is POSIX-only (no Windows-client equivalent yet), so
+	# refuse it on a Windows system rather than misbehave silently.
+	if [[ $number_of_api_servers -gt 0 && ${saunafs_info_[is_windows_system]} -eq 1 ]]; then
```
lgsilva3087: Curious about this one: in Windows-client mode the master and chunkservers still run as Linux processes in WSL, and leilfs-api only talks TCP to the master, so I would expect it to run there too. If the guard is just "not tried yet", maybe say that instead of "POSIX-only", and it could sit next to line 49-53 so it fails before the cluster comes up.

tests/tools/saunafs.sh
```diff
@@ -736,6 +834,60 @@ add_chunkserver_() {
...
+	echo "${API_EXTRA_CONFIG-}" | tr '|' '\n'
+	echo "${!this_api_cfg_variable-}" | tr '|' '\n'
+}
+
+# add_api_server_ <id>: allocate a port, write the env, start the leilfs-api
+# daemon and wait until it is ready. Mirrors add_chunkserver_/add_metadata_server_.
+add_api_server_() {
+	local api_id=$1
+	local api_port
+	local api_data_path=$vardir/api_$api_id
```
lgsilva3087: General note, not blocking: a while ago we agreed in the team to write shell expansions as `${var}` rather than `$var`, we find it more readable, especially in paths like these two lines (`${vardir}/api_${api_id}`, `${etcdir}/leilfs_api_${api_id}.env`) where the name blurs into the surrounding text. The new code mixes both (`$pid`, `$action` vs `${api_id}`, `${id}` below), so if you touch the file again it would be nice to use `${var}` throughout; no need for a dedicated round just for this.

.gitignore
```diff
@@ -109,6 +109,14 @@ tests/hdd*
 ##############
 vcpkg_installed
 
+# leilfs-api (Go) #
+##############
+# leilfs-api is a leilos-only submodule; create-softlinks.sh symlinks its Go
+# source modules + CMakeLists.txt into api/, and its test suite into
+# tests/test_suites/API_tests. None of it is tracked in leilfs.
```
lgsilva3087: Consider to reduce the comment please.

CMakeLists.txt
```diff
@@ -445,6 +445,15 @@ add_subdirectory(src/protocol)
 add_subdirectory(src/tools)
 add_subdirectory(src/mount)
 
+# Go tools (API server, leil CLI, monitoring). leilfs-api is NOT part of leilfs;
+# it lives in the private leilos superproject and is symlinked into api/ by
+# leilos/create-softlinks.sh (source modules + a CMakeLists.txt). Build it only
+# when those symlinks are present — a bare leilfs checkout has no api/ and is
+# skipped. api/CMakeLists.txt is itself self-guarding (Windows / no `go`).
```
lgsilva3087: Consider to rephrase the comment to avoid mentioning private modules, even if they are there until ready.

lgsilva3087 (commented):
Great work. Please consider my comments.

tests/tools/saunafs.sh
GigaCronos: Fixed

tests/tools/saunafs.sh
GigaCronos: Fixed

tests/tools/saunafs.sh
GigaCronos: Fixed

tests/tools/saunafs.sh
GigaCronos: Done

.gitignore
GigaCronos: Done

CMakeLists.txt
GigaCronos: Done

tests/tools/saunafs.sh
GigaCronos: Fixed

lgsilva3087 (approved):
LGTM. Great work!!!

PR #954, opened 2026-09-11
fix(cs): drop job replies to a dead connection

lgsilva3087:
A background job can finish after the master connection that asked for it is gone: the socket was killed and a new one may already be connected, or masterconn_term has released the connection and the job pools flush their last completions afterwards. The reply was then queued on the next socket, or written through a dangling pointer, which AddressSanitizer reports as a heap-use-after-free at shutdown.

Found while giving the chunkserver one connection per metadata server: connections then come and go while their jobs are still in flight, and the review of that path showed the released code has the same fault on its single connection.

Each connected socket now arms a shared flag that both job callbacks capture. Leaving CONNECTED for KILL and destroying the connection clear it, so a late completion frees its packet instead of replying on behalf of a socket that no longer exists.

Related to: LS-1075

### DIFF
--- a/src/chunkserver/master_connection.cc
+++ b/src/chunkserver/master_connection.cc
@@ -54,6 +54,7 @@
 static constexpr uint32_t kMaxBackgroundJobsThreshold = (kMaxBackgroundJobsCount * 9) / 10;
 
 MasterConn::~MasterConn() {
+	*callbackActive_ = false;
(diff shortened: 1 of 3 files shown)


rolysr (approved):
LGTM. Good catch.

ralcolea (approved):
Really nice catch!! @lgsilva3087 💪 🔥 🚀

dmga44 (approved):
LGTM.

PR #955, opened 2026-09-11
fix(mount): report why non-empty mountpoint fails

antuan96314:
fuse_mnt_check_empty() is a vendored copy of the libfuse helper with its diagnostics stripped. It returned a bare -1 both when opendir() failed and when the directory was not empty, and the caller turned that into exit(1) without printing anything. It was the only early-exit branch in main() with no message, so mounting over a non-empty directory looked like a silent failure.

Print the reason on stderr for both cases and point at the '-o nonempty' option. stderr is still connected at this point; the redirect to /dev/null happens later, in mainloop(), after fuse_session_mount() succeeds.

Close LS-789.


rolysr (approved):
LGTM.

lgsilva3087 (approved):
LGTM

PR #960, opened 2026-09-17
fix: symlink legacy saunafs paths for packaging

rolysr:
The saunafs->leil rebrand renamed usr/include/saunafs, var/lib/saunafs, and each per-package doc dir (e.g. saunafs-master) to their leil-* equivalents, but debian packaging (maintained out-of-tree) still lists the old paths, so dh_install and dh_installexamples fail with "missing files" until it's updated.

Create a symlink at each legacy path pointing at its leil replacement at install time, skipping it when a real, non-symlink directory already occupies that path so an existing deployment's actual data/config is never clobbered.


rolysr:
@gemini-code-assist review

svaroq (approved):
LGTM 👍

ralcolea (approved):
Good job @rolysr! 👍

PR #961, opened 2026-09-17
fix(mount): stop serving stale size after close

lgsilva3087:
A size cached by the mount could survive a write that had already been flushed and closed, and a reader then got a short read. SQLite reported it as disk I/O error or a malformed database.

Two defects had to combine. Ending a write record released the client side value that used to correct a cached length, and nothing invalidated the cache at that point. Separately, an invalidation that landed while a lookup or getattr was in flight was undone when the reply was cached, because entries carried no notion of when their data was fetched.

Both halves of the fix are needed, and each alone was tried against the regression test and failed. Invalidating on flush alone leaves a reply fetched before the flush free to be cached after it. The generation check alone leaves the moment the record ends without any invalidation to advance the counter, so a reply that carries the master's pre-flush size is accepted once the record is gone.

Invalidate the inode where a write record ends and in the FUSE flush: flush, the record teardown shared by release and the failed open cleanups, and the write to read transition, in the order flush, invalidate, end. Give the cache a generation counter that every invalidation advances, remember per inode at which generation it was last invalidated, and have callers capture the counter before asking the master. An insert is dropped when any inode it is keyed by was invalidated after the capture. The table is bounded; when it fills it resets to a floor that only rejects more, never less. Rename now invalidates the moved inode as link already did. Dropping a cached readdir tail past the end marker is cache maintenance and does not advance the counter, so listings still in flight are not rejected.

Checking per inode rather than globally keeps unrelated writes from evicting a traversal's answers: measured hit rate is inside the stock band, where a global check lost about 1.3 points.

Related to: LS-1120


tests/test_suites/ShortSystemTests/test_direntry_cache_size_after_close.sh
```diff
@@ -0,0 +1,175 @@
+timeout_set 3 minutes
```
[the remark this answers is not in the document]
lgsilva3087: Not applicable. `tests/data/ShortSystemTests.txt` is no longer read by CI: the Jenkinsfile runs the leil-tests orchestrator, which discovers the scripts from the directory, and `filter_tests.py` is only used by the old `run_test_concurrently.sh`. The file has not been updated since June 2025 and lacks 88 current tests. Branch build 2 ran this test in ShortSystemTests and it passed.

src/mount/sauna_client.cc
```diff
@@ -2138,6 +2156,16 @@ static finfo* fs_newfileinfo(uint8_t accmode, inode_t inode) {
...
+/*! \brief Drop any cached size for an inode whose write record was just flushed or ended.
+ *
+ * Once the data is out, the master is authoritative for the length, and ending the record also
+ * releases the client side value that used to correct a cached length. Every place that flushes
+ * or ends a write record has to call this, not just the FUSE flush path.
+ */
+static void invalidateCachedSizeAfterWrite(inode_t ino) {
+	gDirEntryCache.lockAndInvalidateInode(ino);
+}
+
```
dmga44: Why wrap the `lockAndInvalidateInode` call in here? I think the idea is clear without the dedicated funciton. Also, you are invalidating more than the cached size.

src/mount/sauna_client.cc
```diff
@@ -2381,6 +2409,18 @@ void release(inode_t ino, FileInfo *fi) {
...
+		// Only a handle that held a write record can leave a stale cached size. The order
+		// matters: get the data out so the master owns the length, drop the cached size while
+		// the write record can still correct a reply in flight, and end the record last.
+		{
+			std::lock_guard lock(fileinfo->lock);
+			if (fileinfo->mode == IO_WRITE || fileinfo->mode == IO_WRITEONLY) {
+				WriteAlgorithm::write_data_flush(fileinfo->data);
+				invalidateCachedSizeAfterWrite(ino);
+			}
+		}
```
dmga44: Please check `remove_file_info`, for the very same `fileinfo` modes you are checking here, it is calling `WriteAlgorithm::write_data_end`, which in turn is calling `write_data_flush`.

You may want to consider invalidating the cache inside the very same `remove_file_info` or making it notify the need for the invalidation.

Also take into account that `release` with its `write_data_end` do not mean that the underlying `inodedata` gets destroyed (it decreases a link counter, if 0 it gets destroyed), so the `maxfleng` is not getting wiped out in those cases.

src/mount/sauna_client.cc
```diff
@@ -2000,15 +2013,20 @@ std::vector<DirEntry> readdir(Context &ctx, uint64_t fh, inode_t ino, off_t off,
...
 		auto marker_index = entry_index;
 		if (!dir_entries.empty()) {
 			marker_index = dir_entries.back().next_index;
 		}
+		// The invalidate below is our own and moves the generation, so the marker cannot be
+		// checked against the value captured before the request.
 		gDirEntryCache.invalidate(ctx, ino, marker_index);
-		gDirEntryCache.insert(ctx, ino, 0, marker_index, marker_index, "", Attributes{{}}, data_acquire_time);
+		gDirEntryCache.insert(ctx, ino, 0, marker_index, marker_index, "", Attributes{{}},
+		                      data_acquire_time, DirEntryCache::kAnyGeneration);
```
dmga44: I think the invalidation was previously there to make sure the entry with `marker_index` index was not going to collide with any previous entry with parent `ino` with that same index.

It has some extra responsabilities now.

src/mount/sauna_client.cc
```diff
@@ -2000,15 +2013,20 @@ std::vector<DirEntry> readdir(Context &ctx, uint64_t fh, inode_t ino, off_t off,
 	gDirEntryCache.updateTime();
 
 	// dir_entries.front().index must be equal to entry_index
-	gDirEntryCache.insertSequence(ctx, ino, dir_entries, data_acquire_time);
-	if (dir_entries.size() < request_size) {
+	// One staleness decision covers this whole reply, entries and end marker alike.
+	bool inserted =
+	    gDirEntryCache.insertSequence(ctx, ino, dir_entries, data_acquire_time, cacheGeneration);
+	if (inserted && dir_entries.size() < request_size) {
```
dmga44, line 2020: Consider the case when inserting failed, thus `inserted` is false and `dir_entries` is empty. Previous behavior would anyway insert the marker entry to make use of the `gDirEntryCache` (save some requests to master). Why that wouldn't hold for the proposed implementation?

src/mount/direntry_cache.h
```diff
@@ -718,8 +755,38 @@ class DirEntryCache {
...
+		uint64_t invalidatedAt =
+		    found == invalidated_at_.end() ? invalidation_floor_ : found->second;
+		return invalidatedAt <= generation;
+	}
+
 	Timer timer_;
 	std::atomic<uint64_t> current_time_;
+	std::atomic<uint64_t> generation_;
+	std::unordered_map<inode_t, uint64_t> invalidated_at_;
+	uint64_t invalidation_floor_ = 0;
```
dmga44, line 790: After an hour or so of not understanding this, I think I got it. This is genius IMO, I tried coming up with something simpler for its target case but got reduced to this.

The case it is protecting is this one:
- you get the generation in `lookup`, `getattr` and `readdir` before the possible actual requests to the master.
- some disruptive operation (the invalidate variants) gets called due to any of the operations that touches the queried inodes after the generation is retrieved but before the entries are inserted back in the cache.
- the insertion of the entries is rejected due to its genertion being too old, meaning some invalidation happened along the way.

src/mount/direntry_cache.h
```diff
@@ -183,13 +184,29 @@ class DirEntryCache {
...
+	 */
+	static constexpr uint64_t kAnyGeneration = std::numeric_limits<uint64_t>::max();
+
+	/*! \brief Invalidation generation, bumped by every invalidate call.
+	 *
+	 * A caller reads this before asking the master and passes it back to insert, which drops
+	 * data fetched before an invalidation of an inode the entry is keyed by. That is what stops a
+	 * reply in flight from reviving an entry the invalidation erased.
+	 */
+	uint64_t generation() const { return generation_; }
```
dmga44, line 205: Consider locking the `rwlock_`to make sure getting the generation happens before or after an invalidate operation.

src/mount/direntry_cache.h
```diff
@@ -401,14 +424,20 @@ class DirEntryCache {
...
-		// Avoid inserting stale data
-		if (timestamp + timeout_ <= current_time_) {
-			return;
+	bool insertSequence(const SaunaClient::Context &ctx, inode_t parent_inode,
+	                    const Container &container, uint64_t timestamp,
+	                    uint64_t generation = kAnyGeneration) {
+		if (!currentSince(generation, parent_inode)) { return false; }
+		for (const DirectoryEntry &entry : container) {
+			if (!currentSince(generation, entry.inode)) { return false; }
 		}
```
dmga44, line 437: Please consider the case when a client is repeating some readdir operations in an interval where it is doing operations on some of their files. The retreived data from the master is not going to be cache if any of the nodes was touched while the readdir request was happening, implying extra requests to the master in the future.

This case is the main downside, though acceptable, of the proposed changes from what I see.

dmga44 (approved):
Great job.

src/mount/direntry_cache_unittest.cc
```diff
@@ -421,3 +421,149 @@ TEST(DirEntryCache, InvalidateParentItself) {
...
+	ASSERT_TRUE(cache.lookup(SaunaClient::Context(0, 0, 0, 0), 7, attr));
+	ASSERT_EQ(attr[0], 3);
+}
+
+TEST(DirEntryCache, InvalidationBumpsGeneration) {
+	DirEntryCacheIntrospect cache(5000000);
+
+	uint64_t initial = cache.generation();
+	cache.lockAndInvalidateInode(11);
+	ASSERT_NE(cache.generation(), initial);
```
ralcolea: Maybe `ASSERT_GT` is more appropriate to validate the generation bump. The same suggestion applies to `ClearBumpsGeneration` test.

src/mount/sauna_client.cc
lgsilva3087: You are right about the name, it drops the whole cached entry. Renamed to `invalidateInodeAfterWriteFlush` and its comment now says so. I kept the function because the first review round found the invalidation missing at two of the three places a write record ends; one greppable name with the rule in its comment is the guard against the next missed place.

src/mount/sauna_client.cc
lgsilva3087: Agreed, moved. The flush, invalidate, end order has to hold (invalidating after `write_data_end` leaves a window with the correction gone and nothing to reject a reply in flight), so the whole sequence now lives in `remove_file_info()`, and `release()` is back to its original shape. The two cleanup callers in `sfs_fuse.cc` run it too; no write can have happened through those handles, so it is a no-op there. On your second point: yes, `write_data_end` only drops a reference, and that is why the invalidation runs on every release of a write handle instead of trying to detect the last one. With per inode stamps the extra invalidations only ever touch that inode.

src/mount/sauna_client.cc
lgsilva3087: Good catch, and it had a cost: every completed listing stamped the parent and rejected the lookups and listings of the same directory still in flight from other processes. That invalidate is cache tail maintenance, not a directory change; whatever removed the entries it drops already stamped the parent. It no longer stamps, and the marker insert now passes the captured generation like every other insert, so the `kAnyGeneration` exception is gone from readdir.

src/mount/sauna_client.cc
lgsilva3087: Because the marker is data too. `inserted` is false only when the parent was invalidated while our request was in flight, for example a create in that directory. The master's "ends at X" then predates the new entry at X, and a cached marker would hide that file from every cached listing until the TTL expires. The old code had that race; skipping the marker costs one extra master request on the next listing instead.

src/mount/direntry_cache.h
lgsilva3087: Thanks. Your three step summary is exactly the case, and it is the one that neither half of the fix covers alone: invalidating on flush protects what is already in the cache, the generation protects against what is about to enter with data fetched before the invalidation.

src/mount/direntry_cache.h
lgsilva3087: I do not think it is needed, and it would put the lock on every lookup and getattr. The counter is atomic and every invalidation increments it before erasing, after the master already holds the mutation. A capture before the increment is rejected at insert; a capture after it means the request left after the mutation, so the reply is fresh either way.

src/mount/direntry_cache.h
lgsilva3087: Agreed, this is the main cost. It was measured with two continuous writers next to a directory traversal: hit rate inside the band of the current code, where the first global version lost about 1.3 points. Skipping only the touched entries instead of the whole batch is possible if a workload shows it; the index chain already tolerates holes from single entry invalidations.

src/mount/sauna_client.cc
```diff
@@ -2638,6 +2672,7 @@ void flush(Context &ctx, inode_t ino, FileInfo* fi) {
 	std::unique_lock lock(fileinfo->lock);
 	if (fileinfo->mode==IO_WRITE || fileinfo->mode==IO_WRITEONLY) {
 		err = WriteAlgorithm::write_data_flush(fileinfo->data);
+		invalidateInodeAfterWriteFlush(ino);
```
ralcolea, line 2677: There is another `write_data_flush()` call in the `fsync()` function. 
Is it not needed to call `invalidateInodeAfterWriteFlush()` there too?

ralcolea (approved):
Great job @lgsilva3087!! and @dmga44!! 🔥 💪 🚀

src/mount/direntry_cache_unittest.cc
lgsilva3087: Done, both tests now assert the counter grew.

src/mount/sauna_client.cc
lgsilva3087: You are right, fsync flushes without invalidating, and the comment claimed every flush path did. The claim was wider than the code, so the comment and the commit body now name the three callers and say why fsync is not one: it does not end the write record, and while the record lives lookup and getattr correct the cached size from it, so no FUSE path sees the old size after fsync. What can keep it is a readdir entry returned through the client library, until its TTL, and the same holds for the flush before a read or an mtime setattr on a written handle. Closing that needs the flush to report whether a write record existed, so a read of an unwritten file does not lose its cached attributes, plus a repeated hit-rate measurement, because fsync-heavy workloads would pay a miss per fsync. Tracked as LS-1123.

ralcolea: I liked your proposal. It makes sense as a follow-up. Great job!! :+1:

PR #962, opened 2026-09-18
feat(master): register chunkservers by identity

lgsilva3087:
## Summary

Second public step toward Phase 1 multi-MDS chunk operations, stacked on #958. It adds the
metadata-server side of the contracts that PR introduced on the chunkserver: identity-first
registration without a chunk inventory, the discovery snapshot hook, and the chunk-operations seams
that let a backend keeping its chunk table elsewhere drive maintenance, replication and reporting.
Master and Shadow take the default on every seam and keep the released registration sequence.

Draft until #958 merges. Base is `chunkserver-multi-connections`, so the diff shows only this
step.

## Why

The chunk inventory a chunkserver recites at registration is how a metadata server that holds its
chunk table in memory learns where the copies are. A server that records copies by a chunkserver's
persistent identity already knows, and asking every such server for the same inventory would
repeat work its records already carry. Registration should therefore be able to complete with the
identity alone.

## How it works

- A metadata server may request the chunkserver identity before registration completes and accept
  a registration that carries no inventory.
- The identity must arrive first, or the registration would be attributed to the wrong
  chunkserver. The sequence gains the states this needs: the request, a host reply deferred until
  the identity answers, a gate rejecting unexpected packets while it is outstanding, and a timeout
  for a peer that never answers.
- A server that does not ask keeps the released sequence, so a chunkserver predating the identity
  exchange registers exactly as it does today.
- Chunkservers learn about the other metadata servers through a snapshot the discovery hook
  supplies. No hook is installed in this build, so no snapshot is ever sent.
- The chunk operations interface gains seams for driving maintenance from outside: pacing getters
  for the maintenance loop, a maintenance step between polls, explicit version-bump control when
  replacing part locations, lost and damaged report batches, probe and registration callbacks, and
  locations for a client holding the write lock.
- A failing member of a replayed batch now fails with EIO instead of ending the process. The
  in-memory backend applies each member directly and never reaches this path.
- The retried record persist in `chunk_emergency_increase_version` emits its changelog line once.
  The in-memory backend never retries, so its behaviour is unchanged.

## Master and Shadow compatibility

Master and Shadow request no identity, install no discovery hook, and take the default on every
new seam. Registration, the maintenance loop, placement and reporting are unchanged for them.

## Scope

The four entry criteria recorded on #958 (probe reply without a to-delete marker, unacknowledged
damage reports, identity confirmation before operations, address-only discovery) are not in this
PR. None of them changes behavior while no discovery hook is installed and no probe is sent, which
is the case for every backend in this repository. Each is a separate change on top of this one.

Related to: LS-1075


src/master/chunks.cc
```diff
@@ -3184,15 +3351,116 @@ void ChunkWorker::mainLoop() {
...
+}
+
+size_t chunk_reclaim_unreferenced(size_t budget, const std::function<bool(uint64_t)> &hasRecord) {
+	// Resumes where the previous call stopped, so a small budget still covers the whole table.
+	static uint32_t bucket = 0;
+	size_t visited = 0;
+	for (uint32_t step = 0; step < static_cast<uint32_t>(kChunkHashSize) && visited < budget;
+	     ++step) {
+		for (Chunk *c : gChunksMetadata->chunkhash[bucket]) {
+			++visited;
```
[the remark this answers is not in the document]
lgsilva3087: Fixed in 08bfde26. The cursor now resumes inside a bucket as well, so one call visits at most `budget` chunks and the next tick continues where it stopped. `ReclaimStopsAtBudgetInsideOneBucket` covers three chunks in one bucket with a budget of two.

tests/test_suites/ShortSystemTests/test_chunk_version_stable_rewrites.sh
```diff
@@ -0,0 +1,48 @@
+timeout_set 2 minutes
```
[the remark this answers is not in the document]
lgsilva3087: Not applicable. CI does not read `tests/data/ShortSystemTests.txt`: the file last changed in June 2025 and 89 of the 228 tests in the suite have no entry. `filter_tests.py` is used only by `tests/tools/run_test_concurrently.sh`. Jenkins build 8 ran `test_chunk_version_stable_rewrites` and passed it.

src/master/chunks.cc
```diff
@@ -2194,6 +2323,7 @@ void chunk_operation_status(Chunk *c, ChunkPartType chunkType, uint8_t status,
 				matoclserv_chunk_status(c->chunkid, SAUNAFS_STATUS_OK);
 				c->operation = Chunk::NONE;
 				c->needVersionIncrease = 0;
+				if (gChunkOperations) { gChunkOperations->operationSettled(c->chunkid); }
```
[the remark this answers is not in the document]
lgsilva3087: Fixed in 08bfde26. `chunk_finalize_failed_operation` fires `operationSettled` after marking the chunk idle, which covers both callers. Master and Shadow take the no-op default. The path is not observable in this repository, so the backend that implements the callback covers it in its own suite.

src/master/chunk_operations_unittest.cc
```diff
@@ -0,0 +1,349 @@
+/*
+   Copyright 2026      Leil Storage
+
+   This file is part of SaunaFS.
```
dmga44: Please update the copyright to use LeilFS in the new files.

src/master/chunk_operations_unittest.cc
```diff
@@ -0,0 +1,349 @@
...
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+*/
```
rolysr: Adjust the Copyright files to the new expected branding. Which includes referring to company name as **Leil Storage OÜ**, to ensure it matches official company name. Also, remeber to update all `SaunaFS` naming to `LeilFS`, as this was one of the latest updates on the Copyright texts on the repo.

src/master/chunkserver_registration_state.h
```diff
@@ -0,0 +1,88 @@
...
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
rolysr: Check comment from: https://github.com/leil-io/leilfs/pull/962/changes/3b9aeac7152e5a7978bf8abf2d9d0725479fcf1f#r4093388969

src/master/chunkserver_registration_state_unittest.cc
```diff
@@ -0,0 +1,96 @@
...
+   the Free Software Foundation, version 3.
+
+   SaunaFS is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with SaunaFS. If not, see <http://www.gnu.org/licenses/>.
+ */
```
rolysr: Check comment from: https://github.com/leil-io/leilfs/pull/962/changes/3b9aeac7152e5a7978bf8abf2d9d0725479fcf1f#r4093388969

rolysr (commented):
First checks for rebranding related suggestions.

src/master/chunk_operations_unittest.cc
lgsilva3087: Fixed.

src/master/chunkserver_registration_state.h
lgsilva3087: Fixed.

src/master/chunkserver_registration_state_unittest.cc
lgsilva3087: Fixed.

